PR #7, opened 2019-12-12
Fixed CI pipeline on forked repositories

### LINKED ISSUE #6: GitHub actions always fails on forked repositories
The current CI workflow executes sonarqube analysis and is dependent on the presence of the secret `SONAR_TOKEN` which is (of course) not available on forks. This is ok, however the build always fails, making it hard to judge whether the code will successfully run. We should detect the presence of a fork in our GitHub actions workflow file and skip the sonarqube task.

oxisto:
Fixes #6 

- Detecting repository using `github.repository`
- Skip running SonarQube task since the `SONAR_TOKEN` secret is not available on forked repositories (which is a good thing)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,19 +12,14 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: "11"
-      - name: Testing
-        run: |
-          echo $GITHUB_REF
-          echo $GITHUB_HEAD_REF
-          echo $GITHUB_BASE_REF
+      # this should either be triggered on the forked repository OR a pull_request
       - name: Build Only
         run: |
           ./gradlew --parallel build
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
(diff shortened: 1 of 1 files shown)


oxisto:
Ok looks like secrets are not available for the `pull_requests` trigger, so we cannot run any quality checks on PRs on forked PRs but they are run on the `Fraunhofer-AISEC/cpg` repository.

oxisto:
Not the best solution but ok for now. It executes the following

- `Build Only` in a commit in a forked repository using `push` trigger
- `Build Only` in a PR using `pull_request` trigger
- `Build and Quality Check` in a commit in this repository

This will mean that for a PR coming from a forked repository it will
- Execute `Build Only` in the forked repository (if they enabled workflows, disabled by default)
- Execute `Build Only` in here. We cannot execute quality checks, since secrets are not available in PR runs

For a PR coming from our repository this probably will mean that
- A `Build and Quality Check` is executed for the commit of the branch that the PR is coming from
- A `Build Only` is executed for the PR itself. This is unnecessary but since I do not see any way to deduce where the PR is coming from, it seems to be the best solution.

PR #12, opened 2019-12-13
Added jshell to README

oxisto:
Added a paragraph about using the cpg library using jshell on the CLI to the README.


JulianSchuette:
Looks good. Thank you!

PR #14, opened 2019-12-17
Introduce proper handling for static calls

### LINKED ISSUE #5: Wrong Base-Reference for Static calls
for 
```java
private static byte[] setMasterPass(String mastPass, byte[] salt) throws IOException, NoSuchAlgorithmException {
        //Create new hash with SHA512
        MessageDigest md = MessageDigest.getInstance("SHA-512");
        ByteArrayOutputStream bout = new ByteArrayOutputStream();
        bout.write(salt);
        bout.write(mastPass.getBytes());
        return md.digest(bout.toByteArray());
    }
```
(MessageDigest.getInstance is a static Call)

the `base` of `getInstance` should not be `this`!

Masrepus:
Fixes #5

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.java
@@ -586,8 +586,10 @@ public class ExpressionHandler
       // done, but we can try to resolve the Expression, and if the Javaparser does not know about
       // it, we assume that this is a static call
       boolean isresolvable = false;
+      String scopeName = null;
       try {
         if (scope instanceof NameExpr) {
+          scopeName = ((NameExpr) scope).getNameAsString();
           ((NameExpr) scope).resolve();
           isresolvable = true;
         }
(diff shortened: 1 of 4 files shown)


oxisto:
Hmm can we somehow address this with the existing `MemberCallExpression` after all, it is still a call to a member of a class, is it not? Not quite sure what the `base` in this case would be. Hmm

Masrepus:
I think the current definition of our `MemberCallExpression` treating it as a call to a function which is a "member of an object" should not be changed to "member of a class". Otherwise one could argue that basically all calls are somehow member calls, as each method is the member of a class. In the end I would still handle static calls as normal `CallExpressions`, as it is done here, with the added benefit that we can distinguish static calls from non-static ones in the call resolver

oxisto:
I agree that the definition should not be changed. However, for example if you look at how Java reflect deals with static calls, it is still referring to the method/member of the class but then invoking this call with `null` instead of a reference to a particular object. 

I am not sure if we ever build the functionality on the CPG to retrieve the matching `MemberExpression` for a `MemberCallExpression`, it was my original intent to do so. But by looking at the current `MemberCallExpression` class, we might be able to get rid of it all together and just have a `CallExpression`

Masrepus:
Merging them all into `CallExpression` sounds like a good idea on the longer term

PR #33, opened 2020-01-10
Adds a bugfix for #31 and adds some comments at other places

### LINKED ISSUE #31: EOG edges missing for Short-circuit evaluation of && and ||
In both Java and C++ the evaluation of the first Operand in && and || determines whether or not the second operator is evaluated or skipped: true || x = true and false && x = false. 
Therefore the EOG must have two edges originating from the first operand, one going to the second operand and the other returning to the operator root node.

konradweiss:
Fixed a mistake in the EOG where the short-circuit evaluation of && and || was not considered.

Fixes #31

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
@@ -128,6 +128,14 @@ public class EvaluationOrderGraphPass implements Pass {
 
   private LanguageFrontend lang;
 
+  /**
+   * Searches backwards in the EOG Graph on whether or not there is a path from a function
+   * declaration to the given node. After the construction phase some unreachable nodes may have EOG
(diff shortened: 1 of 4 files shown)


oxisto:
Could you please rebase it with the current master? There was a problem with the SonarQube checks and I had to modify the build gradle.

src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
```diff
@@ -372,6 +380,7 @@ private void createEOG(@Nullable Statement statement) {
 
       // push statement itself
       pushToEOG(statement);
+
```
oxisto: Unnecessary new line :)

konradweiss: Tried to use newlines to segment code blocks a bit, as they were all squshed up ;)

PR #42, opened 2020-01-21
Fixed code and region

### LINKED ISSUE #41: CXXLanguageFrontend#getRegionFromRawNode parses region incorrect if includes are used
To reproduce, parse any cpp file with the `loadIncludes` option on and supply the path to the include files.

It seems that it tries to parse the region by accessing the `source` field and storing it in the `parentRawSig` variable. However, this contains the source for the parent cpp source file and not the header files that are parsed as a result of it. This leads to errors and in many cases crashes since the graph nodes of the include file are looking for code/region in the source of the cpp file.

To crash, simply make the header file substantially longer than the cpp file, this way, the nodes of the header file will definitely refer to code positions that do not exist in the cpp file.

Is the alternative with using the supplied `getRawSignature` that much worse performance-wise? This function seems to be very hacky.

oxisto:
Fixes #41


oxisto:
Nice! Thanks. I added a simple assert to the unit test and will merge once pipeline goes through.

PR #45, opened 2020-01-24
Improved parsing of method declarations

### LINKED ISSUE #38: constructors and methods in cpp file are not parsed properly / not linked to declaration of header
Consider the following code (parse it with `loadIncludes(true)`

CPP file
```cpp
#include "SimpleClass.h"

SimpleClass::SimpleClass()
{
}

int SimpleClass::SomeFunction()
{
	return 0;
}
```

Header file
```cpp
#pragma once
class SimpleClass
{
	SimpleClass();

	int SomeFunction();
};
```

This will create a graph with 3 declarations:

- A `RecordDeclaration` for `SimpleClass` containing the declaration of the constructor and the function
and
- A `FunctionDeclaration` for `SimpleClass::SimpleClass`, which is the constructor, incorrectly parsed and
- A `FunctionDeclaration` for `SimpleClass::SomeFunction`

In reality, of course those two function declarations should at least be method declaration, the first one should be a constructor declaration and they should somehow be linked to the record declaration.

oxisto:
Fixes #38


oxisto:
Will merge this once a quick check in codyze succeeds.

PR #50, opened 2020-01-30
Allow scanning of all files in a folder

JulianSchuette:
CPG was only scanning the files that were explicitly added to its configuration. This PR makes it accept folder or file names. In case of folders, all source files in folder and sub-folders will be parsed.

This PR is required for `feature/cli` branch of analysisserver.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
@@ -81,14 +81,14 @@ public class TranslationConfiguration {
   public final Map<String, String> symbols;
 
   /** Source code files to parse. */
-  private List<File> sourceFiles;
+  private List<File> sourceLocations;
 
   private File topLevel;
   private List<Pass> passes;
 
   private TranslationConfiguration(
       Map<String, String> symbols,
-      List<File> sourceFiles,
+      List<File> sourceLocations,
       File topLevel,
       boolean debugParser,
       boolean failOnError,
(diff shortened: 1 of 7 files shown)


oxisto:
This had some issues on sonarqube but it does not show the branch anymore. I am trying to re-trigger the PR checks when I get home again.

src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -116,7 +116,7 @@ public static Builder builder() {
     return this.symbols;
   }
 
-  public List<File> getSourceFiles() {
+  public List<File> getSourceLocations() {
     return this.sourceFiles;
```
oxisto: Inconsistent naming

src/main/java/de/fraunhofer/aisec/cpg/TranslationManager.java
```diff
@@ -133,22 +141,37 @@ public boolean isCancelled() {
   private HashSet<LanguageFrontend> runFrontends(
       TranslationResult result, TranslationConfiguration config) throws TranslationException {
 
-    List<File> sourceFiles = this.config.getSourceFiles();
+    List<File> sourceLocations = new ArrayList<>(this.config.getSourceLocations());
     HashSet<LanguageFrontend> usedFrontends = new HashSet<>();
-    for (File sourceFile : sourceFiles) {
-      log.info("Parsing {}", sourceFile.getAbsolutePath());
+    Iterator<File> it = sourceLocations.iterator();
```
oxisto: Not used

oxisto:
Also this implementation confuses me. It took me a while to see that it is adding to the array while looping over it :) Maybe first gather all the source files into an array and then run `parse` over this array?

PR #53, opened 2020-02-02
Fix a few NPEs

JulianSchuette:
EvaluationOrderGraphPass was assuming that frontend is not null, which it actually may be.


oxisto:
There are still weird conflicts with the `TranslationManager.java`, probably the imports. It is really annoying to get them consistent. For now, must files have explicit imports, so no wildcards. It would be good to have spotless enforce this like code-style but seems not possible :(

PR #56, opened 2020-02-03
ScopeManager is now created only once per analysis and for each unit test, not implicitly for every language frontend.

konradweiss:
Inside the Tests when creating the Frontend.

### DIFF
--- a/src/test/java/de/fraunhofer/aisec/cpg/ScopeManagerTest.java
+++ b/src/test/java/de/fraunhofer/aisec/cpg/ScopeManagerTest.java
@@ -1,7 +1,5 @@
 package de.fraunhofer.aisec.cpg;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-
 import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend;
 import de.fraunhofer.aisec.cpg.frontends.TranslationException;
 import de.fraunhofer.aisec.cpg.frontends.java.JavaLanguageFrontend;
@@ -22,9 +20,9 @@ public class ScopeManagerTest {
   void testSetScope() throws TranslationException {
     LanguageFrontend frontend = new JavaLanguageFrontend(config, new ScopeManager());
 
-    assert(frontend == frontend.getScopeManager().getLang());
+    assert (frontend == frontend.getScopeManager().getLang());
 
     frontend.setScopeManager(new ScopeManager());
-    assert(frontend == frontend.getScopeManager().getLang());
+    assert (frontend == frontend.getScopeManager().getLang());
   }
 }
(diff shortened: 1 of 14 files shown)


oxisto:
Not sure if I understand the new behaviour correctly, but not sure if it is a good idea to have the user in charge of creating the scope manager, if you look at this line:
```
TranslationConfiguration.builder().loadIncludes(true).defaultPasses().build(), new ScopeManager())
```

It seems to be very inconsistent in style (builder pattern vs. explicit creation of a document). What happens if i do not create a new scope manager? pass null or an existing reference?

oxisto:
Updated: Nevermind, it is only relevant for the unit tests where we create the language frontend explicitly. Be aware however, that in that case, the passes will not be executed! This might be misleading (and is probably to be discussed anyway).

To give an example, this will NOT execute passes:
```java
TranslationUnitDeclaration tu =
        new CXXLanguageFrontend(
                TranslationConfiguration.builder().defaultPasses().build(), new ScopeManager())
            .parse(new File("src/test/resources/symbols.cpp"));
```

But this will:

```java
TranslationManager analyzer =
        TranslationManager.builder()
            .config(
                    .sourceLocations(new File("src/test/resources/symbols.cpp"))
                    .defaultPasses()
                    .build())
            .build();
    TranslationResult res = analyzer.analyze().get()
```

It is a bit misleading that the `LanguageFrontend` retrieves a configuration, i.e. for passes, even though it is not in charge of executing that passes because the `TranslationManager` is

konradweiss:
Yes, this was only intended as a refactoring, and the Tests were only adapted to run with the new refactored structure.
I did not intend to make passes execute there.

oxisto:
Just a suggestion: Would it make sense to have an (abstract) test class somewhere and define a utility function like `newLanguageFrontend` for unit tests that ensure that all the tests (at least the ones without the pass invocations) have a consistent creation-pattern for the language frontend?, i.e.

- The desired language frontend class
- A "dummy" translation manager configuration
- Path to the correct file (as argument)
- a new `ScopeManger`

This function could then be used to document that this will NOT run the passes. Otherwise we have 20+ tests that need to be adjusted and this way it is only 1 function.

src/test/java/de/fraunhofer/aisec/cpg/ScopeManagerTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+  private TranslationConfiguration config;
+
+  @BeforeEach
+  void setUp() {
+    config = TranslationConfiguration.builder().defaultPasses().build();
+  }
+
+  @Test
+  void testLiteral() throws TranslationException {
```
oxisto: `testEquals`?

konradweiss: It does not matter, they don't implement their own equals method. But I can change it.

oxisto: No i meant the wording of the test :)

PR #61, opened 2020-02-18
Support for function pointers

### LINKED ISSUE #32: Crash related to function pointer call
Nondeterministic crash in CallResolver for this code:

```cpp
#include <iostream>

class A {
  public:
    void fun() {
      std::cout << "Hello" << std::endl;
    }

    int mainA() {
      A a;

      void (A::* f_ptr) () = &A::fun;
      a.fun();
      (a.*f_ptr)();
    }
};

int main() {
  A a;

  void (A::* f_ptr) () = &A::fun;
  a.fun();
  (a.*f_ptr)();
}
```
Additionally, some quick checks showed some further problems in this snippet:
- [x] `(a.*f_ptr)()` has name ".", which makes totally no sense
- [x] `a.fun()` is currently not resolved as the member call on class A that it actually is
- [x] Function pointer handling needs to be added to the call resolver
- [x] Tests for function pointer call resolving are needed

Masrepus:
This closes #32


Masrepus:
Adding more test cases revealed some problems with the type system that correlate with function pointer resolving. These have to be fixed before the PR is mergeable

konradweiss:
Ok this duplicate lines of code are not really a problem and removing the duplications would mean we have to overengineer some inheritance structures or rename stuff which would be just a workaround so I will accept the pull request.

oxisto:
Sorry to interject but what is the point of quality gates if we do not stick with them then? :) Please consider adjusting the quality gate then because at the moment it fails for the master branch: https://sonarcloud.io/dashboard?id=Fraunhofer-AISEC_cpg

PR #62, opened 2020-02-18
Add a way of completely disconnecting a node from the rest of the graph

### LINKED ISSUE #60: Unused DeclaredReferenceExpression still connected via DFG
When parsing the following code, we get a `DeclaredReferenceExpression` "i" that is only connected to the graph via a `DFG` edge:

```c++
class Inner {
public:
    int value;

    int init(int a) {
      value = a;
      return a;
    }
};

class Outer {
public:
    Inner inner;

    int init(Inner i) {
        inner = i;
        return i.value;    // <-- HERE
    }
};



int main() {
  Outer o;
  Inner i;

  i.init(17);

  return o.init(i);
}
```
According to its region, the unused node is related to `return i.value` in line 17. I suspect that this is the `DeclaredReferenceExpression` that was originally a child of the `MemberExpression` for `i.value`. It has since been replaced by a direct connection between the `MemberExpression` and the corresponding `FieldDeclaration`, so it is no longer present in the official `AST`. But as there is still a remaining `DFG` edge, it is  not completely purged from the graph.

Masrepus:
Fixes #60

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/graph/Node.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/graph/Node.java
@@ -239,6 +239,9 @@ public class Node {
    * issue #60). It will most probably be referenced somewhere via DFG or EOG edges. Thus, if it
    * needs to be disconnected completely, we will have to take care of correctly disconnecting these
    * implicit edges.
+   *
+   * <p>ATTENTION! Please note that this might kill an entire subgraph, if the node to disconnect
+   * has further children that have no alternative connection paths to the rest of the graph.
    */
   public void disconnectFromGraph() {
     for (Node n : nextDFG) {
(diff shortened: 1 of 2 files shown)


konradweiss:
I thought about the effect in a graph when a Node is removed that is not a root or leaf node, then the graph would disconnect at that point and it would be necessary to somehow connect the nodes in the prev-Set and the next-Set. It probably is not a problem for the Issue that is fixed. So my proposal would be to implement this connection in the subgraph or my current tendency to document the function to only use is for those specific cases and keep the "Graph stitching" in minde if we ever want to remove an arbitrary node from the graph.

Masrepus:
Ah that's true, I didn't think about the effects of removing intermediate nodes. But I think for now removing nodes is only used for the children of `MemberExpressions`. I think we will most likely never see the need to remove arbitrary intermediate nodes, but if we should, you are right that we need to consider a more complete way of graph stitching

PR #77, opened 2020-03-07
Added SARIF compatible location to nodes

### LINKED ISSUE #69: Better error logging
Just a few examples where things can be improved:

All problems should be logged as errors or at least a warning, i.e. in 
https://github.com/Fraunhofer-AISEC/cpg/blob/5898c909f97f3af01c74653d7eaaba5083f72104/src/main/java/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.java#L219

I like the idea of logging the line when an error occurred, this should be done by all error logs (maybe we should have a special function to enforce it), but then it should also log the file, otherwise it is useless in a multi-file/header scenario.

Other candidates include

https://github.com/Fraunhofer-AISEC/cpg/blob/d6d9b469899ba6f516669ba0156cec22019a2248/src/main/java/de/fraunhofer/aisec/cpg/frontends/Handler.java#L110

https://github.com/Fraunhofer-AISEC/cpg/blob/02dd217321c8d9ecad7eb776edf42b807c342169/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java#L159-L163

Pretty much everytime we log an error but have either a raw ast node or a cpg node available.

oxisto:
Implemented SARIF-compatible physical locations to retrieve the file of a node (see https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127869)

Added utility functions for error and warn logging to show the file location. Fixes #69


oxisto:
This *is* API breaking, so it might make sense to bump the version number up to 1.4.0 if we merge this @konradweiss

konradweiss:
There are still two bugs? Or at least sonar cloud is reporting something?

oxisto:
> There are still two bugs? Or at least sonar cloud is reporting something?

Yes, two possible NPE is a piece of code that looks extremely weird anyway. I will try to resolve it.

src/main/java/de/fraunhofer/aisec/cpg/frontends/java/StatementAnalyzer.java
```diff
@@ -410,6 +444,11 @@ public CompoundStatement handleBlockStatement(Statement stmt) {
 
     if (caseExpression == null) {
       if (optionalTokenRange.isPresent()) {
+        /*
+        TODO: not sure if this is really necessary, it seems to be the same location as
+         parentLocation, except that column starts 1 character later and I am not sure if
+         this is correct anyway
+        */
         // Compute region and code for self generated default statement to match the c++ versions
```
oxisto: I am not really sure if this piece of code is really necessary. I just made it work like before but I do not really see the point in this. I have included a comment about it, so we might decide to deal with it later

PR #93, opened 2020-04-27
TypeSystem Rework

### LINKED ISSUE #35: Type-System needs rework
The current way we parse and store types will likely not work in all cases, especially for C-Types.

Code like `static const char * const somearray[]` can currently not be parsed, and I am not sure if it can even be stored correctly in the adjustment and type-modifier

vfsrfs:
Types are represented as nodes in the graph. This also closes #35


Masrepus:
Looks good to me, everything seems to work as expected, so I think this can now finally be merged @konradweiss :+1:

Issue #92, opened 2020-04-24
Endless parsing for some files


dmitrylavrenov:
Sometimes I try to get CPG for one file. But it looks like endless process. I haven't got any errors but the process takes more than 1 hour and I stop it. 

I can provide with such several files.

JulianSchuette:
Yes, could you please provide an example file?

dmitrylavrenov:
Maybe, the following makes sense.
In first case, I used jshell to parse a file/project. Endless parsing was present.

After that I analysed cpg-vis-neo4j and in the same way tried to parse project and create simple queries in Kotlin. There were no endless parsing.

JulianSchuette:
Could you provide the source file that results in endless parsing?

The following works fine for me, e.g. using the code of `cat` as an example:

```
import de.fraunhofer.aisec.cpg.TranslationConfiguration;
import de.fraunhofer.aisec.cpg.TranslationManager;
import de.fraunhofer.aisec.cpg.graph.FunctionDeclaration;
 
var path = Paths.get("cat.c");
var config = TranslationConfiguration.builder().sourceLocations(path.toFile()).defaultPasses().debugParser(true).build();
var analyzer = TranslationManager.builder().config(config).build();
var result = analyzer.analyze().get();
var tu = result.getTranslationUnits().get(0);
```

dmitrylavrenov:
Sure

`import java.io.File;
import java.nio.file.Path;
import java.nio.file.Paths;
import java.util.*;
import java.util.concurrent.atomic.AtomicInteger;
import java.util.function.Predicate;
import java.util.function.UnaryOperator;
import java.util.stream.Collectors;
import java.util.stream.IntStream;

import org.bytedeco.javacpp.*;
import org.bytedeco.llvm.clang.*;
import org.bytedeco.llvm.global.clang;

import com.google.gson.Gson;

public class ClangCodePathBuilder {
    public static final long FILE_LOC_ID_MULTIPLIER = 10000000L;


//    private HashMap<Cursor, ArrayList<Cursor>> cursorTree;

    public static void main(String[] args) throws Exception {
        System.out.println(clang.clang_getClangVersion().getString());

        for(int i=0; i < 1; i++) {
            System.out.println(i);
            test();
        }
    }

    private static void test() throws Exception{
        long tm = System.nanoTime();


        // Location of the *.cc file
[...]

PR #94, opened 2020-05-06
Reduce graph traversals in toString() methods to avoid unreasonably verbose logging

JulianSchuette:
Possible fix for #92 

Some classes are logging an unreasonable amount of information, such as the full body of a MethodDeclaration. By traversing the graph using `appendSuper`, the result of `toString()` may get out of control and result in several kBs of output per call.

This PR removes the most irrelevant fields from the output of `toString()`.


Masrepus:
We should probably also think about whether logging every single node that is created is really necessary...

src/main/java/de/fraunhofer/aisec/cpg/graph/MethodDeclaration.java
```diff
@@ -17,7 +17,7 @@
  *                   $$  __$$\ $$  __$$\ $$  __$$\
  *                   $$ /  \__|$$ |  $$ |$$ /  \__|
  *                   $$ |      $$$$$$$  |$$ |$$$$\
- *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |      $$  ____/ $$ |\_$$ |g
```
Masrepus, line 20: unintended letter

JulianSchuette: Who says it's unintended? :-)

PR #97, opened 2020-05-11
Fix for null parent of field expression, leading to missing locations.

### LINKED ISSUE #96: Nodes referring to Java fields are missing parent and location
The following snippet instantiates a field `cipher` in an incorrect/unsafe way. However, the node referring to `cipher` does not have a `parent`, nor a `range` and consequently no `location`.

```
public class SealObject {
    private final Cipher cipher;

    public SealObject(String Password) throws Exception  {
        cipher = Cipher.getInstance("SomeUnsafeCipher");
    }
}
```

JulianSchuette:
In [1] and [2], `ExpressionHandler` simplifies the AST by replacing some nodes by a `FieldAccessExpr`. However, the newly created `FieldAccessExpr` lacks some properties - especially its `parent`, `range`, and `tokenRange` are null. This results later in an `IllegalArgumentException` when a rule violation is found at that node and we try to retrieve its position in the code.

This PR fixes #96. But the original author might want to review and possibly simplify the method in general.

[1] https://github.com/Fraunhofer-AISEC/cpg/blob/487558afba9539db62484b02eb23112ecc85478a/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.java#L498

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.java
@@ -492,7 +492,11 @@ public class ExpressionHandler
         if (!field.isStatic()) {
           // convert to FieldAccessExpr
           FieldAccessExpr fieldAccessExpr = new FieldAccessExpr(new ThisExpr(), field.getName());
+          expr.getRange().ifPresent(fieldAccessExpr::setRange);
+          expr.getTokenRange().ifPresent(fieldAccessExpr::setTokenRange);
+          expr.getParentNode().ifPresent(fieldAccessExpr::setParentNode);
           expr.replace(fieldAccessExpr);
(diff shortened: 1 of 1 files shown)


oxisto:
Good catch. Actually it does not simplify the AST but rather create additional AST nodes (the `this` expression), since this is a reversal of syntactic sugar, i.e. using `field` instead of `this.field`. So I think the overall idea how to do it has some value because there might be some additional logic depending on that `this` expression that otherwise will not be triggered - or needs to be simulated in the name handler.

At the time, the original author - which is me ;), wrote this we did not have any additional fancy things like passes and so so, so I was not aware of any other problems.

So there are two issues here:

- Obviously I was to lazy to correctly set needed values, this is now fixed, thanks to @JulianSchuette 
- Does this `replace` operation cause any other problems? For that I am not sure and I would suspect that it would be ok.

PR #104, opened 2020-05-25
Fix Incorrect Bindings

### LINKED ISSUE #103: Incorrect Bindings when setting RefersTo Edge
Bindings may be incorrect if they are not available or not being updated in cachedDeclaration (Binding -> Declaration).

1. When an Expression uses a Variable that is parsed afterwards, there is no Declaration stored in the cachedDeclaration and therefore the refersTo cannot be set in expressionRefersToDeclaration (CXXLanguageFrontend).
Example:
```
class MyClass {
  void someFun() {
    x = 5;
  }
  int x;
};
```

2. When a VariableDeclaration is replaced by a FieldDeclaration (handleCompositeTypeSpecifier) the cached Declarations are not updated. If a Expression wants to set now the refersTo-Edge it will get the old variable declaration instead of the new FieldDeclaration.
Example:
```
class MyClass {
  int a = 0;
  void someFun() {
    int c = a + 1;
  }
};
```

vfsrfs:
Closes #103


oxisto:
Something very weird is going on. This also broke the build on master now.

PR #106, opened 2020-06-05
Add support for Java super invocations

Masrepus:
Closes #99 and finishes a task in #37 (SuperExpr support)

PR #107, opened 2020-06-05
Add missing DFG edges for self-modifying operators

### LINKED ISSUE #105: Missing DFG Edge in compound/pre-/postfix operators
Some operators implicitely modify the variable they are applied on. Therefore the DFG Edge needs to also point backwards from the operator to the DeclaredReferenceExpresion. Additionally the DeclaredReferenceExpression needs to support readwrite access, which means that there are DFG edges from the DeclaredReferenceExpression to the VariableDelaration and vice versa. The following cases show the operators which have missing DFG edges:

1. Pre- and Postfix operators such as `++` or `--`.
```
int i = 0;
i++;
```
pre-postfix

Missing DFG Edges:
- From ++ to i (DeclaredReferenceExpression)
- From i (DeclaredReferenceExpression) to i (VariableDeclaration) as ++ is readwrite access

2. Compound Operators such as +=.
```
int i = 0;
i+= 2;
```
compound
Missing DFG Edges:
- From += to i (DeclaredReferenceExpression)
- From i (DeclaredReferenceExpression) to i (VariableDeclaration)

vfsrfs:
Closes #105


src/main/java/de/fraunhofer/aisec/cpg/graph/BinaryOperator.java
```diff
@@ -51,6 +50,11 @@
   /** The operator code. */
   private String operatorCode;
 
+  /** Required for compound BinaryOperators. This should not be stored in the graph */
+  @Transient
+  private final List<String> compoundOperators =
+      new ArrayList<>(Arrays.asList("*=", "/=", "%=", "+=", "-=", "<<=", ">>=", "&=", "^=", "|="));
```
JulianSchuette, line 56: This creates a partly immutable list and then a (mutable) ArrayList from it. Why not directly instantiate the type that is needed?

oxisto:
Nice debugging :)

vfsrfs:
I only implemented the change @JulianSchuette suggested, but a non-related test started failing on the Github Workflow. Locally it was passing every time so I had to do some remote debugging with Github as debugger 😂. Apparently the order in which the tests got executed mattered.

oxisto:
Yes, we were experiencing similar issues on #106. Let's see if this really fixed now or if this will come back. Good catch.

Masrepus:
No that's really fixed now :+1:

src/main/java/de/fraunhofer/aisec/cpg/graph/BinaryOperator.java
```diff
@@ -29,10 +29,9 @@
 import de.fraunhofer.aisec.cpg.graph.HasType.TypeListener;
 import de.fraunhofer.aisec.cpg.graph.type.Type;
 import de.fraunhofer.aisec.cpg.graph.type.TypeParser;
-import java.util.HashSet;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
```
oxisto, line 32:
```suggestion
import java.util.HashSet;
import java.util.Objects;		
import java.util.Set;
```

src/main/java/de/fraunhofer/aisec/cpg/graph/ValueAccess.java
```diff
@@ -0,0 +1,12 @@
+package de.fraunhofer.aisec.cpg.graph;
+
+/** Expressions can have multiple types of accesses. Determines dataflow (DFG) edges */
+public interface ValueAccess {
+  enum accessValues {
```
oxisto, line 5: Is there a specific reason for a non-public enum here? Not sure if I have ever seen this. Makes it a bit annoying to always write `ValueAccess.accesValues.READ`

konradweiss:
@vfsrfs There are some new merge conflicts and UnaryOperator still uses the old Enum, at least in the version that I am pulling, which does not exist anymore.

vfsrfs:
should be fixed now 😀

PR #112, opened 2020-06-18
tag-based Releases

oxisto:
This creates a system similar to the one we use in codyze, where we trigger a "release", i.e. a publish to Maven central as well as a fancy GitHub release based on a tag that is either created on the GitHub page or by pushing a tag that begins with the prefix `v`. 

This is my opinion less error prone then using the gradle publish plugin. It does have the small drawback that we will not able to release SNAPSHOT builds to maven snapshots - however this is my opinion not needed anyway, since we have a fairly small release cycle. It could also allows us to have some kind of automation, i.e. create new releases automagically 1 per week if something changed on master or something like that.


Masrepus:
I wouldn't mind having that, what does everyone else think about this?

JulianSchuette:
Installing SNAPSHOTs in local Maven repo are of course required to keep up with Codyze but I understand this would not be a problem?

oxisto:
It is not even necessary to use SNAPSHOTS for that. You can launch gradle using `./gradlew -Pversion=2.0.2` to install it as a specific version.

oxisto:
Any thoughts, @konradweiss ?

konradweiss:
> Any thoughts, @konradweiss ?

Sounds good to me.

PR #117, opened 2020-06-19
Typeparser hotfix

Masrepus:
The fix to the type parser turned out to not be robust enough, in cases automatic cleanup after the analysis phase has to be skipped. Thus I fixed it in a better way now, and introduced measures to reset all statically held state before each and every unit test, to avoid undefined results in the future

### DIFF
--- a/src/test/java/de/fraunhofer/aisec/cpg/frontends/LanguageFrontendTest.java
+++ b/src/test/java/de/fraunhofer/aisec/cpg/frontends/LanguageFrontendTest.java
@@ -28,27 +28,15 @@ package de.fraunhofer.aisec.cpg.frontends;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import de.fraunhofer.aisec.cpg.BaseTest;
 import de.fraunhofer.aisec.cpg.TranslationConfiguration;
 import de.fraunhofer.aisec.cpg.TranslationManager;
 import de.fraunhofer.aisec.cpg.TranslationResult;
-import de.fraunhofer.aisec.cpg.graph.TypeManager;
-import de.fraunhofer.aisec.cpg.graph.type.TypeParser;
 import java.io.File;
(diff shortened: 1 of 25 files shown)


oxisto:
Minor suggestion: Instead of having the `resetPersistentState` function copy/pasted into all the tests that need it can we introduce an abstract test class and implement the function there? `@BeforeEach` should then propagate to the derived class as well. This will make it easier if something like this needs to be done in the future for other subsystems as well.

src/test/java/de/fraunhofer/aisec/cpg/frontends/LanguageFrontendTest.java
```diff
@@ -31,12 +31,25 @@
...
 import org.junit.jupiter.api.Test;
 
 class LanguageFrontendTest {
 
+  /**
+   * {@link TypeParser} and {@link TypeManager} hold static state. This needs to be cleared before
+   * all tests in order to avoid strange errors
+   */
+  @BeforeEach
+  void resetPersistentState() {
```
oxisto, line 48: Minor suggestion: Instead of having the resetPersistentState function copy/pasted into all the tests that need it can we introduce an abstract test class and implement the function there? @BeforeEach should then propagate to the derived class as well. This will make it easier if something like this needs to be done in the future for other subsystems as well.

Masrepus: Ah does that work? Didn't know whether this propagates

oxisto: I am 99% sure it does :)

Masrepus: Just tested it, it does :+1:

Masrepus:
Alright, should be done now

Issue #119, opened 2020-06-20
CPG Relationship Types:DFG,EOG


niuzhi:
Hello,
DFG,EOG detailed meaning，I don't understand these releationship type,can you help me to explain?
thank you very much.

JulianSchuette:
Yes, documentation is still a bit scarce - we are just getting started with it ([here](https://www.codyze.io/docs/)), but details on the CPG are still missing.

`DFG` means Data Flow Graph and represents edges from _definitions_ of a variable to _usages_ of that variable.

`EOG` means Evaluation Order Graph and represents edges between (executable) AST nodes in the order they would be executed when running the code.

PR #120, opened 2020-06-21
Automatic CPG Documentation

JulianSchuette:
This PR publishes JavaDocs under https://fraunhofer-aisec.github.io/cpg/

Resolves https://github.com/Fraunhofer-AISEC/cpg/issues/119

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: JavaDoc
         run: |
-          ./gradlew gitPublishPush
+          ./gradlew -Dorg.ajoberstar.grgit.auth.username=$GITHUB_TOKEN gitPublishPush
         env:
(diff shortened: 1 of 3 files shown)


oxisto:
This will be executed for all branches, possibly overwriting the doc with stuff from feature branches and pull requests. I would either restrict it to the master branch or better yet, wait for the #112 PR and only apply it on the version tags.

.github/workflows/build.yml
```diff
@@ -30,6 +30,12 @@ jobs:
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: JavaDoc
```
oxisto, line 33: This will be executed for all branches, possibly overwriting the doc with stuff from feature branches and pull requests. I would either restrict it to the master branch or better yet, wait for the #112 PR and only apply it on the version tags.

oxisto, line 33:
```suggestion
      - name: JavaDoc
         if: startsWith(github.ref, 'refs/tags/v')
```

PR #121, opened 2020-06-25
Refactor EOGPass to have a more readable and modular structure

konradweiss:
Closing #3 The class is restructured and the else-if construct is replaced by a handler structure similar to the Handler in the different Frontends. Although this does not decrease the size of the file, it makes it more readable and more searchable. Extracting more of the code into other classes was currently not done and can be done after identifying functional groups.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
@@ -37,7 +37,6 @@ import java.util.*;
 import java.util.stream.Collectors;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
-import org.eclipse.cdt.internal.core.model.Binary;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


oxisto:
You seem to suffer from the heisenbug test, please consider rebasing to the current master.

konradweiss:
> You seem to suffer from the heisenbug test (fixed in #117), please consider rebasing to the current master.

Thank you, for helping me with my suffering from quantum coding paradoxe.

src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
```diff
@@ -67,12 +69,58 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(EvaluationOrderGraphPass.class);
 
+  protected final HashMap<Class<? extends Node>, CallableInterface<? extends Node>> map =
```
oxisto, line 72: Please use abstract types for fields, i.e. `Map`.

oxisto, line 72:
```suggestion
  protected final Map<Class<? extends Node>, CallableInterface<? extends Node>> map =
```

oxisto:
Documentation could still be approved a little bit but I think it is already a lot better than it way. LGTM.

Masrepus:
+1 from my side, I like this style that is similar to our frontend handlers better than the old one

Issue #116, opened 2020-06-19
More thorough testing


Masrepus:
The test suite will be analyzed for feature coverage and updated/extended to cover the newest capabilities our graph has. Some tests are still only rather superficial concerning the results that are expected from the graph, e.g. they were made at a time where we did not have some call and variable resolving capabilities. Other parts of the CPG are still more or less untested, so additional tests will be created to make sure that we have a consistent state when introducing new features

konradweiss:
We will close this issue for now as testing by now has reached an acceptable state and this issue does not reference any concrete missing tests that one could tackle to close the issue

PR #122, opened 2020-06-26
First batch of test improvements

### LINKED ISSUE #57: Utility helper function to create consistent language frontends for unit tests
Just a suggestion: @konradweiss: Would it make sense to have an (abstract) test class somewhere and define a utility function like `newLanguageFrontend` for unit tests that ensure that all the tests (at least the ones without the pass invocations) have a consistent creation-pattern for the language frontend?, i.e.

- The desired language frontend class
- A "dummy" translation manager configuration
- Path to the correct file (as argument)
- a new `ScopeManger`

This function could then be used to document that this will NOT run the passes. Otherwise we have 20+ tests that need to be adjusted and this way it is only 1 function.

_Originally posted by @oxisto in https://github.com/Fraunhofer-AISEC/cpg/pull/56#issuecomment-581361387_

Masrepus:
Belongs to the ongoing process of #116 and closes #57. This PR contains the following things:
*  Standard ways of parsing files to a graph for more unified tests
*  Some tests didn't handle the situation where the graph contains an "unknown declarations" TU
*  Test cases for the common supertype finder, including fixes for multi-inheritance scenarios as possible in C++

Plus some other more local fixes for problems that came up in the process of doing those things


oxisto:
What was the trick to force SonarCloud in analyzing previous branches that are now PRs? Another commit?

Masrepus:
yup

src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -79,6 +79,9 @@
 
   @NonNull private List<Pass> passes;
 
+  /** For test purposes, the cleanup phase after the analysis may want to be skipped */
+  private boolean cleanupOnCompletion;
```
oxisto, line 83: I am not a big fan of including functionality that is only needed during testing, especially in a very public facing class such as the configuration. Is there any other way to do this?

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java
```diff
@@ -219,6 +221,22 @@ private RecordDeclaration handleCompositeTypeSpecifier(CPPASTCompositeTypeSpecif
...
+              recordDeclaration.getName(), recordDeclaration.getName(), recordDeclaration);
+      recordDeclaration.getConstructors().add(constructorDeclaration);
+      lang.getScopeManager().addValueDeclaration(constructorDeclaration);
+    }
+
+    lang.getScopeManager().leaveScope(recordDeclaration);
+    return recordDeclaration;
+  }
+
+  private void handleMembers(
```
oxisto, line 238: This `handle` function brakes the pattern of returning the handled type, maybe rename this to avoid confusion.

src/test/java/de/fraunhofer/aisec/cpg/TestUtils.java
```diff
@@ -59,30 +59,74 @@
...
+   * @return A list of {@link TranslationUnitDeclaration} nodes, representing the CPG roots
+   * @throws Exception Any exception thrown during the parsing process
+   */
+  public static List<TranslationUnitDeclaration> analyze(
+      List<File> files, Path topLevel, boolean usePasses, boolean cleanupOnCompletion)
+      throws Exception {
+    TranslationConfiguration.Builder builder =
         TranslationConfiguration.builder()
-            .sourceLocations(files)
+            .sourceLocations(files.toArray(File[]::new))
```
oxisto, line 103: Can we convert `sourceLocations` to a list as well? Why is it an array

src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
Masrepus: In order to correctly test the common type finder, I need the TypeManager to hold all state that it collected when the graph has been built. Thus cleaning everything up is counterproductive here. Anyways I don't think that skipping cleanup has any bad consequences either.

src/test/java/de/fraunhofer/aisec/cpg/TestUtils.java
Masrepus: It has always been a File vararg. No idea whether this should be changed. I wouldn't mind

Masrepus: I think I'll just add an overloading function. I just saw that it's converted back to a list anyway :sweat_smile:

oxisto: Ah, I didn't see that it's a varags, I think that's probably there for convenience. I avoid changing too much public API in this PR, so adding an overloaded function is ok as long as the old API still works.

src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
Masrepus: I'll do that differently and mock the cleanup method instead of creating a global config for it

oxisto: Yeah that sounds like a better solution.

src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -146,6 +146,7 @@ public File getTopLevel() {
     private List<String> includePaths = new ArrayList<>();
     private List<Pass> passes = new ArrayList<>();
     private boolean codeInNodes = true;
+    private boolean cleanupOnCompletion = true;
```
oxisto, line 149:
```suggestion

```

src/test/java/de/fraunhofer/aisec/cpg/TestUtils.java
```diff
@@ -59,30 +62,68 @@
...
+  /**
+   * Like {@link #analyze(List, Path, boolean)}, but for all files in a directory tree having a
+   * specific file extension
+   *
+   * @param fileExtension All files found in the directory must end on this String. An empty string
+   *     matches all files
+   * @param topLevel The directory to traverse while looking for files to parse
+   * @param usePasses Whether the analysis should run passes after the initial phase
+   * @param cleanupOnCompletion Whether {@link de.fraunhofer.aisec.cpg.graph.TypeManager} etc.
+   *     should be cleaned up after the analysis has ended
```
oxisto, line 74:
```suggestion
```

src/test/java/de/fraunhofer/aisec/cpg/TestUtils.java
```diff
@@ -59,30 +62,68 @@
...
+   * @return A list of {@link TranslationUnitDeclaration} nodes, representing the CPG roots
+   * @throws Exception Any exception thrown during the parsing process
+   */
+  public static List<TranslationUnitDeclaration> analyze(
+      String fileExtension, Path topLevel, boolean usePasses, boolean cleanupOnCompletion)
```
oxisto, line 79:
```suggestion
      String fileExtension, Path topLevel, boolean usePasses)
```

Masrepus: Huh seems like I forgot these

oxisto: :) Almost there. Nice job with mockito.

oxisto: Ah.. you need to get rid of the additional parameters of the `analyze` call as well ;)

Masrepus: :sweat_smile:

PR #146, opened 2020-07-10
Fix: Field "fqn" of CallExpression/MemberCallExpression must not contain modifiers "*" or "[]" for PointerTypes

JulianSchuette:
Codyze relies on the fact that type names in the "fqn" field of `CallExpression`/`MemberCallExpression` do not have any pointer modifiers.

The new type system introduced such modifiers and was simply copying them into the `fqn` field. This PR leaves `Type.getTypeName()` untouched but restores the original representation in `MemberCallExpression.fqn`.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.java
@@ -392,14 +392,15 @@ class ExpressionHandler extends Handler<Expression, IASTInitializerClause, CXXLa
 
     CallExpression callExpression;
     if (reference instanceof MemberExpression) {
-      String baseTypename =
-          ((Expression) ((MemberExpression) reference).getBase()).getType().getTypeName();
-      // FIXME this is only true if we are in a namespace! If we are in a class, this is wrong!
-      //  happens again in l398
-      // String fullNamePrefix = lang.getScopeManager().getFullNamePrefix();
-      // if (!fullNamePrefix.isEmpty()) {
-      //  baseTypename = fullNamePrefix + "." + baseTypename;
-      // }
+      String baseTypename;
+      // Pointer types contain * or []. We do not want that here.
+      Type baseType = ((Expression) ((MemberExpression) reference).getBase()).getType();
+      if (baseType instanceof PointerType) {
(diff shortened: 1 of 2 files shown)


Masrepus:
You should consider `Type.getRoot().getTypeName()` as you might encounter nested pointers

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.java
```diff
@@ -394,12 +391,9 @@ private CallExpression handleFunctionCallExpression(CPPASTFunctionCallExpression
...
       String baseTypename;
       // Pointer types contain * or []. We do not want that here.
-      Type baseType = ((Expression) ((MemberExpression) reference).getBase()).getType();
-      if (baseType instanceof PointerType) {
-        baseTypename = (((PointerType) baseType).getElementType().getTypeName());
-      } else {
-        baseTypename = baseType.getTypeName();
-      }
+      Type baseType = ((Expression) ((MemberExpression) reference).getBase()).getType().getRoot();
+      assert !(baseType instanceof SecondOrderType);
```
oxisto, line 395: Do we have a policy about the usage of `assert`? Maybe through our custom `TranslationException` instead?

JulianSchuette:
Like this? Terminology is a bit confusing here: what is the difference between a "first order type" and a "root type"? Should `getRoot()` not rather be named `getFirstOrderType()` or `asFirstOrderType()`?

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.java
JulianSchuette: AFAIK there is no poliy. But for me, it was not obvious that `getRoot()` is supposed to guarantee that the result is _not_ a `PointerType` and _not_ a `ReferenceType`. This is not clearly documented and it does not become obvious from the method's name so I made it explicit with an `assert` for my future self to remember.
`asserts` should not add overhead to runtime, so it's a matter of taste.

Masrepus:
Yes that's how I meant it. I think the root of a type is always a first order type, but that's something @vfsrfs should clarify.

PR #148, opened 2020-07-10
Made adjustments to support usage of CPG in a Graal native image

### LINKED ISSUE #147: non-reflective code for CXXLanguageFrontend::getLocationFromRawNode is wrong
If reflective access fails, we currently provide another way to retrieve the "parent signature", i.e. the signature of the file where the AST node is located in.

https://github.com/Fraunhofer-AISEC/cpg/blob/b8801eb7f38b04e4254f0876e7e439cb56824c41/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontend.java#L261-L266

However this horribly fails if it tries to retrieve the code for an AST element that is inside a header that is included in a cpp file. In this case, the loop will traverse all the way to the CPP file while the code region refers to the header file, because (for some reason) CDT considers the CPP the "parent" of the included header.

There are not many scenarios where this non-reflective code is actually needed. Currently one is, if the cpg is executed in Graal/Substrate VM - however there is a workaround to allow reflection for certain fields there as well. So the easy thing would be to just remove this code, because I am not quite sure if we can really fix it.

oxisto:
This PR introduces some necessary adjustments, mostly additional resource files to make CPG compatible with https://www.graalvm.org/, i.e. needed for https://github.com/Fraunhofer-AISEC/cpg-vis-neo4j/issues/1

Graal offers a possibility to compile a Java program with an ahead of time (AOT) compiler not a native image. However, there are some constraints on reflection, that need to be specified beforehand.

Speed improvements are pretty crazy though... simple sample programs run almost 10x faster.

Closes #147


oxisto:
Ok.. interesting, a) why is this function returning `null` but annotated as `NonNull` and b) why is this only failing now?

oxisto:
Looks like this was a bug, the parent method signature already had `@Nullable`.

oxisto:
Ok after some more testing, also `getTranslationUnit` will refer back to the CPP file, not the header file. Not sure if we can fix the non-reflective code at all at this point.

PR #154, opened 2020-07-11
Correctly setting the type of a CPP simple declaration

### LINKED ISSUE #153: TypeParser causes StringIndexOutOfBoundsException
The typeparser seems to use some fancy-magic to parse the type string looking for brackets, such as (, [, however, it seems to be missing code that handles if for some reason it cannot find the end-bracket or if multiple brackets are stacked.

Consider the following snippet:

```cpp
#define SOME_MACRO(t) t

void SomeFunction(const char* text=SOME_MACRO(""));
```

The first issue, is that the type parser seems to try to parse the complete function declaration signature to retrieve a type (probably the return type) of it. Not sure what the intent here is.

https://github.com/Fraunhofer-AISEC/cpg/blob/1024f02164d547f961c0f2f3a6d0642410ef6711/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationListHandler.java#L58-L59

Second issue is, that this will crash within the type parser, since he seems to be confused by the double `)` within the function signature.

```
Caused by: java.lang.StringIndexOutOfBoundsException: String index out of range: 16
	at java.base/java.lang.StringLatin1.charAt(StringLatin1.java:48)
	at java.base/java.lang.String.charAt(String.java:709)
	at de.fraunhofer.aisec.cpg.graph.type.TypeParser.findMatching(TypeParser.java:187)
	at de.fraunhofer.aisec.cpg.graph.type.TypeParser.separate(TypeParser.java:321)
	at de.fraunhofer.aisec.cpg.graph.type.TypeParser.createFrom(TypeParser.java:648)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationListHandler.handleSimpleDeclaration(DeclarationListHandler.java:58)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationListHandler.lambda$new$0(DeclarationListHandler.java:48)
	at de.fraunhofer.aisec.cpg.frontends.Handler.handle(Handler.java:112)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationHandler.handleSingleDeclarator(DeclarationHandler.java:199)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationHandler.handleSimpleDeclaration(DeclarationHandler.java:171)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationHandler.lambda$new$0(DeclarationHandler.java:53)
	at de.fraunhofer.aisec.cpg.frontends.Handler.handle(Handler.java:112)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationHandler.handleTranslationUnit(DeclarationHandler.java:261)
	at de.fraunhofer.aisec.cpg.frontends.cpp.CXXLanguageFrontend.parse(CXXLanguageFrontend.java:218)
	at de.fraunhofer.aisec.cpg.TranslationManager.runFrontends(TranslationManager.java:196)
	at de.fraunhofer.aisec.cpg.TranslationManager.lambda$analyze$0(TranslationManager.java:90)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1771)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1763)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1016)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1665)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1598)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
```

oxisto:
This sets the type of a CPP simple declaration according to its declspecifier instead of the raw signature.

This partially fixes #153. It will not occur that often now probably but the core issue is still there.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
@@ -339,7 +339,7 @@ public class DeclarationHandler extends Handler<Declaration, IASTDeclaration, CX
   static String getTypeStringFromDeclarator(
       IASTDeclarator declarator, IASTDeclSpecifier declSpecifier) {
     // use the declaration specifier as basis
-    StringBuilder typeString = new StringBuilder(declSpecifier.getRawSignature());
+    StringBuilder typeString = new StringBuilder(declSpecifier.toString());
 
     // append names, pointer operators and array modifiers and such
     for (IASTNode node : declarator.getChildren()) {
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
@@ -27,6 +27,7 @@
 package de.fraunhofer.aisec.cpg.frontends.cpp;
 
 import static de.fraunhofer.aisec.cpg.helpers.Util.errorWithFileLocation;
+import static de.fraunhofer.aisec.cpg.helpers.Util.getRawFunctionReturnType;
 
 import de.fraunhofer.aisec.cpg.frontends.Handler;
 import de.fraunhofer.aisec.cpg.graph.*;
@@ -109,9 +110,10 @@ public class DeclarationHandler extends Handler<Declaration, IASTDeclaration, CX
 
     lang.getScopeManager().enterScope(functionDeclaration);
 
-    functionDeclaration.setType(
-        TypeParser.createFrom(
-            ctx.getRawSignature().split(functionDeclaration.getName())[0].trim(), true));
+    // use our own version, since the decl specifier might be imprecise, i.e. not having pointers
+    typeString = getRawFunctionReturnType(functionDeclaration, ctx.getRawSignature());
+
+    functionDeclaration.setType(TypeParser.createFrom(typeString, true));
 
     if (ctx.getBody() != null) {
       Statement bodyStatement = this.lang.getStatementHandler().handle(ctx.getBody());
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
@@ -27,7 +27,6 @@
 package de.fraunhofer.aisec.cpg.frontends.cpp;
 
 import static de.fraunhofer.aisec.cpg.helpers.Util.errorWithFileLocation;
-import static de.fraunhofer.aisec.cpg.helpers.Util.getRawFunctionReturnType;
 
 import de.fraunhofer.aisec.cpg.frontends.Handler;
 import de.fraunhofer.aisec.cpg.graph.*;
@@ -39,8 +38,12 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import org.eclipse.cdt.core.dom.ast.IASTArrayModifier;
+import org.eclipse.cdt.core.dom.ast.IASTDeclSpecifier;
 import org.eclipse.cdt.core.dom.ast.IASTDeclaration;
+import org.eclipse.cdt.core.dom.ast.IASTDeclarator;
 import org.eclipse.cdt.core.dom.ast.IASTNode;
+import org.eclipse.cdt.core.dom.ast.IASTPointerOperator;
 import org.eclipse.cdt.core.dom.ast.IASTTranslationUnit;
 import org.eclipse.cdt.internal.core.dom.parser.cpp.*;
 
@@ -101,7 +104,7 @@ public class DeclarationHandler extends Handler<Declaration, IASTDeclaration, CX
     FunctionDeclaration functionDeclaration =
         (FunctionDeclaration) this.lang.getDeclaratorHandler().handle(ctx.getDeclarator());
 
-    String typeString = ctx.getDeclSpecifier().toString();
+    String typeString = getTypeStringFromDeclarator(ctx.getDeclarator(), ctx.getDeclSpecifier());
 
     // It is a constructor
     if (functionDeclaration instanceof MethodDeclaration && typeString.isEmpty()) {
@@ -109,9 +112,6 @@ public class DeclarationHandler extends Handler<Declaration, IASTDeclaration, CX
     }
 
     lang.getScopeManager().enterScope(functionDeclaration);
-
-    // use our own version, since the decl specifier might be imprecise, i.e. not having pointers
-    typeString = getRawFunctionReturnType(functionDeclaration, ctx.getRawSignature());
 
     functionDeclaration.setType(TypeParser.createFrom(typeString, true));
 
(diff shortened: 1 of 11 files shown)


oxisto:
*sigh*, looks like the internal `IASTDeclSpecifier` class gets rid of the pointer in the type for some reason, so this is not good :(

oxisto:
I have added additional logic, that is also applied in other cases it seems to "clean" the type string before handing it to the type parser. This addresses this issue for now but it could still potentially creep up.

oxisto:
Back to using the decl-specifier, "enriched" with the information given in the declarator, i.e. pointer and array types. Way less hacky than before.

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
```diff
@@ -325,4 +327,27 @@ TranslationUnitDeclaration handleTranslationUnit(CPPASTTranslationUnit translati
...
+
+    // append names, pointer operators and array modifiers and such
+    for (IASTNode node : declarator.getChildren()) {
+      if (node instanceof IASTPointerOperator || node instanceof IASTArrayModifier) {
+        typeString.append(node.getRawSignature());
+      }
+    }
+
+    return typeString.toString();
+  }
```
vfsrfs, line 352: This seems to be less-error-prone than using `ctx.getRawSignature` for most cases. Nonetheless, we have a problem with function pointers, as by using this method to determine the raw type string instead of using `ctx.getRawSignature`, the fptr `void ((*single_param)(int))` is transformed into `void`. This is not the intended behaviour, since fptrs are addressed by their own type.

oxisto: That is why this is not used for function pointers (yet)

Basically function pointers are handled (for some reason) here:

https://github.com/Fraunhofer-AISEC/cpg/blob/8ff74381b5a2a753727da2eb4db890f79024c84a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationListHandler.java#L67-L68

I can add a comment to make it more clear

src/main/java/de/fraunhofer/aisec/cpg/graph/type/TypeParser.java
```diff
@@ -180,6 +204,11 @@ private static int findMatching(char openBracket, char closeBracket, String stri
     int i = 0;
 
     while (counter != 0) {
+      if (i >= string.length()) {
+        // dirty hack for now
+        return string.length();
+      }
+
```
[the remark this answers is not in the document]
oxisto: Sure, feel free to adjust this PR

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
vfsrfs: Yes I have seen that if and that should be correct. However I tested the following snippet and I 
reached this code here: [https://github.com/Fraunhofer-AISEC/cpg/blob/8ff74381b5a2a753727da2eb4db890f79024c84a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java#L107](https://github.com/Fraunhofer-AISEC/cpg/blob/8ff74381b5a2a753727da2eb4db890f79024c84a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java#L107)
```
int main(){
void ((*single_param)(int));
}
```

vfsrfs: In that case apparently the code is treated as a VariableDeclaration and therefore the if statement is true.

oxisto: Can you add this snippet somewhere to the existing function pointer unit tests. You can deliberately make the unit test fail and then I can have a look.

konradweiss: > Can you add this snippet somewhere to the existing function pointer unit tests including an assert on how the type should be parsed. You can deliberately make the unit test fail and then I can have a look.

Ok I looked through the PR, and will await until you took that look.

vfsrfs: Yes I will add the snippet to a test

vfsrfs:
I have added the snippet and a test, which is failing due to the usage of `getTypeStringFromDeclarator`

oxisto:
You made it fail! ;) Thanks, I will look into it.

oxisto:
Ok looks like there are some more issues that creep up here. I have adjusted the test file to contain two different function pointers each on global and local scope:
```
void ((*global_single_param)(int));
int (*global_two_param)(int, unsigned long);

int main() {
  void ((*local_single_param)(int));
  int (*local_two_param)(int, unsigned long);
}
```

Now the "original" problem was, that your function pointer `local_single_param` was nested in "()" and we do not handle those very well. In fact, `DeclarationHandler::handleFunctionPointer` was never called and it was just treated as a regular variable and things went wrong. So I am not sure if we want to actually fix this in this PR, but we do not handle nested declarators very well at the moment. This is not specific to just function pointers. The `local_two_param` gets parsed correctly (including its type).

The second issue is, that while the `global_two_param` function pointer is parsed correctly (including its type), it is - for some reason - being parsed as a `FieldDeclaration`, so it seems we have another issue where global function pointers are parsed as field declarations.

Lastly, is it intended, that the function pointer type has an empty name, @vfsrfs?  Wouldn't it make sense to set the signature as name? This is how clang does it for example:

```
|-VarDecl 0x7fe6e5070d28 <types/fptr_type.cpp:1:1, col:27> col:9 single_param 'void ((*)(int))'
`-VarDecl 0x7fe6e5070ff0 <line:2:1, col:35> col:7 intFptr1 'int (*)(int, unsigned long)'
```

oxisto:
I fixed the aforementioned issues with function pointer types, all tests should run now.

I think it would still be better to parse the function pointer type in the `handleSimpleDeclaration` by going downwards into the decl-specifiers and declarator and see whether we can re-construct them in the correct order for the function type, rather than going upwards from the declarator and use the raw string, but that is the way it is for now.

I left this comment in the code as well for future discussions.

src/main/java/de/fraunhofer/aisec/cpg/graph/type/TypeParser.java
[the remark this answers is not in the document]
konradweiss: @vfsrfs @oxisto can one of you resolve this if it is still relevant? Then we can merge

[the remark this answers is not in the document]
oxisto: I would suggest a merge now and then @vfsrfs can see afterwards if any more issues creep up. The PR already fixed way more bugs than it intended to ;)

Issue #152, opened 2020-07-11
Several issues in analysis of CPP and associated header


oxisto:
Not sure how to exactly phrase this and maybe this is part of a bigger issue but consider the following two very basic files:

Test.cpp
```cpp
#include "Test.h"

void SomeClass::SomeMethod() {
    someField->anotherField = true;
}
```

Test.h
```h
class AnotherClass {
public:
    bool anotherField;
};

class SomeClass {
public:
	void SomeMethod();

private:
	AnotherClass* someField;
};
```

This produces the following graph

![Screenshot 2020-07-11 at 12 11 59](https://user-images.githubusercontent.com/12459061/87221873-bb564200-c36f-11ea-903a-16206c475aa4.png)

There are several issues that we do not tackle very good currently (and this might actually be split up into multiple issues):

1) The cpp file contains the *definition* of `SomeClass::SomeMethod` ("right" side of the graph connected to TU 253), while the header file includes its *declaration* ("left" side of the graph, connected to TU 266). However, there is no clear link between those two nodes of type `MethodDeclaration` (except the reference to the type `void`). The definition also does not connect to its class, i.e. by `METHOD` and `RECORD_DECLARATION` nodes.

2) This dis-connection also leads to the issue, that the `DeclaredReferenceExpression` for `someField` is not referring to the `FieldDeclaration` for `someField` and it's type is thus unknown. This might solve itself, if 1) is fixed. #155 will fix this.

3) There are constructor declarations for `SomeClass` and `AnotherClass` (yellow nodes), which are not there in the code. While this might make sense to have them (because they are implicitly there), but neither the field `dummy` nor `implicit` has been set to to true. 

4) They also refer to the *same* unknown type, while in theory this makes sense from a programming point of view, for a practical analysis this could lead to the confusion that both constructors refer to the *same* type (which they do not).

5) Furthermore, the constructors' type should actually be known, i.e. the type of the class itself.

6) Lastly, is `AnotherClass` and `AnotherClass*` seem to be two different types, which might technically be correct. However, should there not be any sort of connection between those to types, since the are referring to the same class?

oxisto:
I have updated the example using the current CPG master + a merged in PR #179 

![Screenshot 2020-07-24 at 15 07 23](https://user-images.githubusercontent.com/12459061/88394246-61b63480-cdbf-11ea-87dc-f944e5425d98.png)

It's looking far better than before. Let's go through the issues

[...]

PR #155, opened 2020-07-12
Correctly connecting field references in C++ method definitions

oxisto:
Fixes parts of the issues described in #152 

Still a draft, because I am not happy with the way that the scope walker currently works. He currently operates on its own AST walk and does not leverage the information of the ScopeManager. This leads to incorrect behavior in C++ method *definitions*, because they are not surrounded by a record AST node but the function body should be interpreted to be within the scope of its record. I currently built a workaround but a better way would be to actually use the `ScopeManager`'s stored scope.

Furthermore, currently all `CPPASTIdExpression` are treated as `DeclaredReferenceExpressions`, while they could also be member / field expressions, because `someField` can be syntactic sugar for `this->someField`. Only the latter is treated by CDT as a `CPPASTFieldReference`, which is then parsed as a `MemberExpression`.

There are further issues in parsing `this->someField` because the `this` is not handled correctly, but I am not sure if this will be part of this PR or not.


oxisto:
Looks like this is the best we can do for now without a major scope manager overhaul. Ready to review.

oxisto:
There is still a big issue we need to discuss. How should we handle method declarations and definitions. Before this PR, if we encounter a method *definition* and we deduce the correct associated `RecordDeclaration`, we set the `record` on the `MethodDeclaration` (which can represent a definition and declaration), but we did not add the method to the record declaration. The record declaration only listed the method *declaration*. So if we start walking from record declarations, there was no way to reach the method *definition* node at some point. However, looping all methods will work and you will end up at the correct record.

This PR changes this and just stupidly adds the `MethodDeclaration` node, which represents the method *definition* to the `RecordDeclaration` as well. Now this might not actually be what we want, since the `methods` field is tagged as `@AST` and the *definition* was NOT part of the record's AST. 

We have two possible ways in my opinion:
* The way clang handles this afaik, is that the method *declaration* node has a link to its *definition* (if presents), and only keeps the original members that were part of the record AST in the `members` field of the record
* Do not interpret the `members` field as `@AST` but rather something more like the refersTo fields. However I am not sure if this will break existing walkers that assume that you can just traverse from a class to its fields.

konradweiss:
@oxisto I am working on making that larger Scope Manager rework, that allows to handle these external Definitions of internally Declared methods, fields etc. However the problem you described with the difference between Definition and Declaration was one that I could not solve until now. Two proposals of mine are:

- simply adding a `definitions` reference similar to your proposal above only that we decouple the definition and Declaration such that the current `members` and `fields`? And we have `FunctionDefinition extends Function Declaration` with a `defines`-field  if recognized as external?
- Have a general AST edge that is labeled as such and overwrite the member and field with the found definition.

src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/ScopeManager.java
```diff
@@ -227,6 +228,8 @@ public void enterScope(Node nodeToScope) {
         newScope = new TryScope(nodeToScope);
       } else if (nodeToScope instanceof NamespaceDeclaration) {
         newScope = new NameScope(nodeToScope, getCurrentNamePrefix(), lang.getNamespaceDelimiter());
+      } else if (nodeToScope instanceof TranslationUnitDeclaration) {
```
konradweiss, line 231: I had a similar Idea but now we have a global scope, with no node that is created in the scope manager as root node and several global scopes as his children that have a translation-unit (always a file) as its children.

oxisto: Hm yes you are correct and I was actually under the impression that global scopes are limited to a particular TU in C++, but they are not. Top-level declarations of multiple files that are linked together need to be unique, because the global scope is common across all bundled TUs. Meh :(

I will instead just reset the scope to the global scope after each TU entry for now. Reasoning behind this was, that a scope error in one TU A spilled over into the second TU B, making the cpg think, that all declarations of B were in the scope of a class of A.

oxisto:
> @oxisto I am working on making that larger Scope Manager rework, that allows to handle these external Definitions of internally Declared methods, fields etc. However the problem you described with the difference between Definition and Declaration was one that I could not solve until now. Two proposals of mine are:
> 
> * simply adding a `definitions` reference similar to your proposal above only that we decouple the definition and Declaration such that the current `members` and `fields`? And we have `FunctionDefinition extends Function Declaration` with a `defines`-field  if recognized as external?
> * Have a general AST edge that is labeled as such and overwrite the member and field with the found definition.

I would rather not have another node type because that creates a mess because at least two different classes derive from `FunctionDeclaration`: `MethodDeclaration` and `ConstructorDeclaration`. So I would rather have a boolean field `isDefinition` inside the existing node type to denote it is a definition and then have the `definitions` (can there be multiple?!) field you suggested. This is imho very close to what clang does.

konradweiss:
You wrote that this is a draft, do we want to accept this as a temporary fix until we have discussed on whether we introduce the concept of a definition alongside a declaration or replace the past declaration with the external definition?
What I see here is that we lose the semantical difference between declaration and definition (having only declarations) or the semantic of a member in members (field in fields) of being inside the class(Part of its AST).

oxisto:
> You wrote that this is a draft, do we want to accept this as a temporary fix until we have discussed on whether we introduce the concept of a definition alongside a declaration or replace the past declaration with the external definition?
> What I see here is that we lose the semantical difference between declaration and definition (having only declarations) or the semantic of a member in members (field in fields) of being inside the class(Part of its AST).

My suggestions would be that I quickly try to implement my proposed clang-like solution and see if it works for my use case. This will at least allow us to differentiate between a declaration and definition and offer more than we have now. If this is then the preferred long-run solution is up to you.

konradweiss:
> d be that I quickly try to implement my proposed clang-like solution and see if it works for my use case. This will at least allow us to differentiate between a declaration and definition and offer more than we have now. If this is then the preferred long-run solution is up to you.

Sounds good

oxisto:
Created this very basic implementation now. It has several constraints, such as it will only work on methods and only if the declaration was parsed before the definition. Plus, the code is not very good, but will do for now. In the long run we probably want a pass, that takes care of this.

@konradweiss It is your decision to accept this as-is or if you want to do it cleanly via an extra pass.

konradweiss:
> Created this very basic implementation now. It has several constraints, such as it will only work on methods and only if the declaration was parsed before the definition. Plus, the code is not very good, but will do for now. In the long run we probably want a pass, that takes care of this.
> 
> @konradweiss It is your decision to accept this as-is or if you want to do it cleanly via an extra pass.

This is fine for me, we can later extract It into an extra pass but it working for now is better then never getting the perfect solution.

PR #157, opened 2020-07-13
Support for black and whitelisting include files

oxisto:
Still a draft for now. Currently, only absolute paths are supported.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
@@ -62,13 +62,15 @@ public class TranslationConfiguration {
 
   /**
    * This acts as a white list for include files, if the array is not empty. Only the specified
-   * includes files will be parsed and processed in the CPG.
+   * includes files will be parsed and processed in the CPG, unless it is a port of the blacklist,
+   * in which it will be ignored.
    */
   public final List<String> includeWhitelist;
 
   /**
    * This acts as a black list for include files, if the array is not empty. The specified includes
-   * files will excluded from being parsed and processed in the CPG.
+   * files will excluded from being parsed and processed in the CPG. The blacklist entries always
+   * take priority over those in the whitelist.
    */
   public final List<String> includeBlacklist;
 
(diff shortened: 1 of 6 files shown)


oxisto:
Supports relative and absolute paths now.

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontend.java
```diff
@@ -86,19 +88,85 @@
...
+            LOGGER.debug("Include file {} not on the whitelist. Ignoring.", path);
+            return null;
+          }
+
           LOGGER.debug("Loading include file {}", path);
           FileContent content = FileContent.createForExternalFileLocation(path);
           return (InternalFileContent) content;
         }
 
+        private boolean hasWhitelist() {
```
JulianSchuette, line 120: Propose to name it `hasIncludeWhitelist` as otherwise it will be unclear what "whitelist" refers to.

src/test/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXIncludeTest.java
```diff
@@ -88,4 +90,114 @@ void testCodeAndRegionInInclude() throws Exception {
 
     assertEquals(new Region(16, 3, 16, 15), location.getRegion());
   }
+
```
JulianSchuette, line 93: Propose to add a test case that tests a file contained in white- *and* blacklist as this is what will most likely create confusion/error cases.

src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -60,6 +60,18 @@
    */
   public final String[] includePaths;
 
+  /**
+   * This acts as a white list for include files, if the array is not empty. Only the specified
+   * includes files will be parsed and processed in the CPG.
+   */
+  public final List<String> includeWhitelist;
+
+  /**
```
JulianSchuette, line 69: Add remark to clarify that blacklist has priority over whitelist

src/test/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXIncludeTest.java
oxisto: I added a test case (only for relative paths though).

PR #160, opened 2020-07-13
Fixed potential copy/paste error in namespace declaration handling

### LINKED ISSUE #159: C++ namespace definitions lead to incorrect type names
Consider the following file:

```cpp
namespace FirstNamespace {
  class SomeClass {

  };
};

class AnotherClass {

};
```

This results into the following declarations:

- A namespace declaration for `FirstNamespace`, which also  contains a record declaration within that namespace for `FirstNamespace::SomeClass` - ok
- A record declaration for `FirstNamespace::AnotherClass` which is incorrect, since the namespace scope is already closed and `AnotherClass` should be part of the global namespace

oxisto:
Closes #159


JulianSchuette:
This refers to NamespaceDeclarations, but it leaves `Type` untouched. Would it be possible to turn `Type.getTypeName()` into FQNs where possible, i.e. where the namespace is known?

oxisto:
I would rather not do more just to fix this obvious error in this PR, tbh. Seems like @konradweiss is working on more namespace stuff anyway.

PR #163, opened 2020-07-14
Type hierarchy (for Java)

JulianSchuette:
This PR (once finished) brings in a `JavaExternalTypeHierarchy` pass. This pass populates an additional `SUB_TYPES` (aka `getSubTypes()`) edge between `Type` nodes, reflecting class _and_ interface inheritance.

Codyze needs type hierarchy information in almost all rules, as users will typically want to refer to generic super types such as `java.security.Key`, rather than copying the same rule for all possible sub types (_DHPrivateKey, DHPublicKey, DSAPrivateKey, DSAPublicKey, ECPrivateKey, ECPublicKey, PBEKey, PrivateKey, PublicKey, RSAMultiPrimePrivateCrtKey, RSAPrivateCrtKey, RSAPrivateKey, RSAPublicKey, SecretKey_, etc..)

Note that this is orthogonal to `TypeHierarchyResolver` which creates hierarchies over `RecordDeclarations`, i.e. between classes which are available as source code and in scope of the analysis (i.e. it would not be able to capture the aforementioned `java.security.Key` hierarchy). This PR creates hierarchies over _types_ rather than _record declarations_ - these are related but different concepts.


oxisto:
Can we somehow integrate this with the `TypeHierarchyResolver` pass, which somehow tries to do the same but on the record nodes instead of the types?

JulianSchuette:
> Can we somehow integrate this with the `TypeHierarchyResolver` pass, which somehow tries to do the same but on the record nodes instead of the types?

See my description above.

JulianSchuette:
Seriously? This thing is failing because of JavaDoc? *sigh*...

oxisto:
> > Can we somehow integrate this with the `TypeHierarchyResolver` pass, which somehow tries to do the same but on the record nodes instead of the types?
> 
> See my description above.

Ah the description was not there before :D

src/main/java/de/fraunhofer/aisec/cpg/passes/JavaExternalTypeHierarchyResolver.java
```diff
@@ -0,0 +1,56 @@
...
+import org.slf4j.LoggerFactory;
+
+public class JavaExternalTypeHierarchyResolver extends Pass {
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(JavaExternalTypeHierarchyResolver.class);
+
+  @Override
+  public void accept(TranslationResult translationResult) {
+    // Run only for Java.
+    if (JavaLanguageFrontend.class.equals(this.lang.getClass())) {
```
oxisto, line 25: Could we add function to the `Pass` interface that returns the list of supported frontends? We probably need more passes that are language specific, i.e. the upcoming C++-only pass that connects definitions and declarations. We could already prepare this here.

src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -300,6 +300,7 @@ public Builder registerPass(@NonNull Pass pass) {
     public Builder defaultPasses() {
       registerPass(new FilenameMapper());
       registerPass(new TypeHierarchyResolver());
```
oxisto, line 302: Not necessarily related to this PR, but can we rename the `TypeHierarchyResolver` since it does not really resolve types at all?

Masrepus: How do you mean? It resolves the hierarchy of object types in the graph by adding the necessary relationships to their respective RecordDeclarations

konradweiss: @Masrepus I think he refers to the last Note in the initial description

Masrepus: Ah I see, so renaming to something like RecordHierarchyResolver would probably be the way to go

PR #164, opened 2020-07-14
Support for C++ attributes as annotations

oxisto:
This PR parses C++ attributes (https://en.cppreference.com/w/cpp/language/attributes) as annotation nodes. It is not yet a full annotation support, this will be a separate PR.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/graph/Annotation.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/graph/Annotation.java
@@ -27,6 +27,7 @@
 package de.fraunhofer.aisec.cpg.graph;
 
 import java.util.List;
+import java.util.Objects;
 
 public class Annotation extends Node {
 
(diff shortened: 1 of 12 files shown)


src/main/java/de/fraunhofer/aisec/cpg/graph/Annotation.java
```diff
@@ -0,0 +1,62 @@
...
+      return false;
+    }
+
+    Annotation that = (Annotation) o;
+    return super.equals(that) && Objects.equals(values, that.values);
+  }
+
+  @Override
+  public int hashCode() {
+    return super.hashCode();
```
Masrepus, line 60: Why override the hashcode function here if you just call super? Should we include the values list in the hash? Otherwise I would just remove the overriding function

oxisto: I was basically copying the way all other nodes do it (which in my opinion does not make sense at all). I would suggest keeping it consistent with the others and see if we want to change it globally.

Masrepus: Ah alright, that makes sense. I'll open another issue for that 👍

PR #165, opened 2020-07-15
Adding Tests for variable resolution that test diverse cases and use …

konradweiss:
…same nameed variables and fields to showcase problems with variable resolution. The Tests are currently disable and the plan is to iteratively enable the tests when a fixpr got through. Planed improvements are: Using the scopeManager, introducing Namespaces as structure holding nodes, allowing external declarations of fields, structures etc. that is allowed in c++. Improving resolution through ScopeManager.


konradweiss:
I added flexibility to the tests, the variables  `ENFORCE_REFERENCES` and `ENFORCE_MEMBER_EXPRESSION` were added and set to false to flexibly enforce theire exitstence in tests and not blame failing tests on this nuance on how `referseTo` should be structured

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverCppTest.java
```diff
@@ -0,0 +1,433 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+// Todo VariableResolverPass 23 Failed, 5 Passed
+
+@Disabled(
+    "Until changing variable resolution to ScopeManager. Then in detail disable the tests that need specific fixes")
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class VariableResolverCppTest {
```
oxisto, line 44: Shouldn't all test derive from `BaseTest`?

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverCppTest.java
```diff
@@ -0,0 +1,433 @@
...
+    main = Util.getOfTypeWithName(nodes, FunctionDeclaration.class, "main");
+
+    outer_function1 =
+        outerClass.getMethods().stream()
+            .filter(method -> method.getName().equals("function1"))
+            .collect(Collectors.toList())
+            .get(0);
+    forStatements = Util.filterCast(SubgraphWalker.flattenAST(outer_function1), ForStatement.class);
+
+    // Functions i nthe outer and inner object
```
oxisto, line 138: spelling

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverCppTest.java
```diff
@@ -0,0 +1,433 @@
...
+
+  private RecordDeclaration innerClass;
+  private FieldDeclaration innerVarName;
+  private FieldDeclaration innerStaticVarName;
+  private FieldDeclaration innerImpThis;
+  private FieldDeclaration innerImpOuter;
+
+  private FunctionDeclaration main;
+
+  private MethodDeclaration outer_function1;
```
oxisto, line 63: no underscores in Java variable names please

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverCppTest.java
```diff
@@ -0,0 +1,433 @@
...
+  private MethodDeclaration outer_function4;
+  private MethodDeclaration outer_function5;
+
+  private MethodDeclaration inner_function1;
+  private MethodDeclaration inner_function2;
+
+  private Map<String, Expression> callParamMap = new HashMap<>();
+
+  @BeforeAll
+  public void initTests() throws ExecutionException, InterruptedException {
```
oxisto, line 76: override the preperation function of BaseTest and call super

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverCppTest.java
```diff
@@ -0,0 +1,433 @@
...
+    final String topLevelPath = "src/test/resources/variablesExtended/";
+    List<String> fileNames =
+        Arrays.asList("scope_variables.cpp", "external_class.cpp", "external_class.h");
```
oxisto, line 79: Why is the header file included as a translation unit? It should not be and rather `loadIncludes` should be set to true

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverCppTest.java
```diff
@@ -0,0 +1,433 @@
...
+    return null;
+  }
+
+  public MemberExpression getCallWithMemberExpression(String literal) {
+    Expression exp = callParamMap.get(literal);
+    if (exp instanceof MemberExpression) return (MemberExpression) exp;
+    return null;
+  }
+
+  @Test
```
oxisto, line 191: I am not sure if I like the test-design of this test class. Having just one assert in a test is unnecessary overhead of launching a test in my opinion

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverJavaTest.java
```diff
@@ -0,0 +1,305 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+// Todo VariableResolver 7 Failed, 7 Passed
+
+@Disabled(
+    "Until changing variable resolution to ScopeManager. Then in detail disable the tests that need specific fixes")
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class VariableResolverJavaTest {
```
oxisto, line 39: Same here, extend from BaseTest

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverJavaTest.java
```diff
@@ -0,0 +1,305 @@
...
+
+  private MethodDeclaration outer_function1;
+  private List<ForStatement> forStatements;
+  private MethodDeclaration outer_function2;
+  private MethodDeclaration outer_function3;
+  private MethodDeclaration outer_function4;
+
+  private MethodDeclaration inner_function1;
+  private MethodDeclaration inner_function2;
+  private MethodDeclaration inner_function3;
```
oxisto, line 68: Lot of copy/paste from the other test?!

src/test/resources/variablesExtended/external_class.cpp
```diff
@@ -0,0 +1,9 @@
+#include<string>
```
oxisto, line 1:
```suggestion
#include <string>
```

oxisto:
44 code smells? That is too much :)

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverCppTest.java
konradweiss: It makes it more modular to se which cases exactly fail for the test-first approach, after all tests run properly we can consider packing them into one class.

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverCppTest.java
konradweiss: true

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverJavaTest.java
konradweiss: The class structure was intentionally kept similar, so it made sense to do so and then add only the variables structures that were not part of the other test class.

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variableResolution/VariableResolverCppTest.java
oxisto: No, you have to load all cpp files, but the header files will be loaded through `loadIncludes`. This simulates the usual behaviour if you would compile the program, you would only specify the cpp files as translation units.

src/main/java/de/fraunhofer/aisec/cpg/helpers/Util.java
```diff
@@ -79,6 +79,36 @@
         .collect(Collectors.toList());
   }
 
+  /**
+   * Returns the first element of the specified Class-type {@param specifiedClass} that has the name
+   * {@param name} in the list {@param listOfNodes}.
+   *
+   * @param <S> Some class that extends {@link Node}.
+   */
+  public static <S extends Node> S getOfTypeWithName(
```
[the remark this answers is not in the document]
oxisto: Btw it seems that a similar function already exists in `Util.subnodesOfType`

konradweiss:
> 44 code smells? That is too much :) You can - and should - remove the `public` identifier from the test functions.

I will work on that one as well

src/main/java/de/fraunhofer/aisec/cpg/helpers/Util.java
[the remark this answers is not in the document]
konradweiss: @oxisto Is that relevant here? This function specifically should work not only on TranslationUnits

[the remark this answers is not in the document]
oxisto: I am referring to those functions
https://github.com/Fraunhofer-AISEC/cpg/blob/1024f02164d547f961c0f2f3a6d0642410ef6711/src/main/java/de/fraunhofer/aisec/cpg/helpers/Util.java#L91-L104

There is an exponential increase of utility functions and we should be a bit careful here.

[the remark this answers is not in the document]
Masrepus: I would recommend such a workflow instead, using TestUtils functions created specifically for this purpose: https://github.com/Fraunhofer-AISEC/cpg/blob/64ca4c77f23cc25d41eada7524d7db29f2ab5739/src/test/java/de/fraunhofer/aisec/cpg/enhancements/VariableResolverTest.java#L50-L51

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverJavaTest.java
```diff
@@ -0,0 +1,306 @@
...
+        Util.filterCast(nodes, CallExpression.class).stream()
+            .filter(call -> call.getName().equals("printLog"))
+            .collect(Collectors.toList());
+    calls.sort(new NodeComparator());
+
+    List<RecordDeclaration> records = Util.filterCast(nodes, RecordDeclaration.class);
+
+    // Extract all Variable declarations and field declarations for matching
+    externalClass =
+        Util.getOfTypeWithName(nodes, RecordDeclaration.class, "variables_extended.ExternalClass");
```
Masrepus, line 107: Using the workflow mentioned above would also make constructs like this much less verbose

src/main/java/de/fraunhofer/aisec/cpg/helpers/Util.java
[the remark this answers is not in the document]
oxisto: Good point, @Masrepus. Can we document this somehow, i.e. in a `CONTRIBUTING.md` or something? At "some" point we probably want to cleanup the unit tests, but at least it would be good to have a a reference for future tests.

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverJavaTest.java
Masrepus: This would then be e.g.
```java
List<RecordDeclaration> records = Util.subnodesOfType(tu, RecordDeclaration.class);
externalClass = TestUtils.findByUniqueName(records, "variables_extended.ExternalClass");
```

src/main/java/de/fraunhofer/aisec/cpg/helpers/Util.java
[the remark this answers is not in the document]
Masrepus: Yeah, cleaning up the tests and unifying them is something I'm currently doing (or better will continue to be doing after the exams)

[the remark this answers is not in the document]
konradweiss: > I would recommend such a workflow instead, using TestUtils functions created specifically for this purpose:
> 
> https://github.com/Fraunhofer-AISEC/cpg/blob/64ca4c77f23cc25d41eada7524d7db29f2ab5739/src/test/java/de/fraunhofer/aisec/cpg/enhancements/VariableResolverTest.java#L50-L51

I don't like that idea because it works on the elements twice, which is not a large issue when we talk about testing, but specifically because we then have to add an additional line for that call in 28 + 17 places in those two TestClasses only, at that point I prefer single utility function more, to reduce code reuse. I will however see which functions in Utils are only used in tests and extract them to the TestUtils class.

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverJavaTest.java
konradweiss: > verbose

Do you mean verbose or readable, are there logs written when using these?

src/main/java/de/fraunhofer/aisec/cpg/helpers/Util.java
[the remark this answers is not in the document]
konradweiss: > @oxisto Is that relevant here? This function specifically should work not only on TranslationUnits

Ah the duplicate is one message that I did not see when replying

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverJavaTest.java
Masrepus: I mean verbose in the sense of a lot of code

src/main/java/de/fraunhofer/aisec/cpg/helpers/Util.java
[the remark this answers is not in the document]
Masrepus: I mean you can of course create a function that wraps the subnode and finding calls into one method, but inside this function you should probably use calls to the existing subnode and finding methods, as your method basically recreates everything the other two are doing already, only in a single method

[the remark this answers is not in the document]
Masrepus: Additionally the assumption that exactly one node matching the criteria needs to be present in some cases is enforced in `findByUniqueName` instead of just returning null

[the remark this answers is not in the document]
konradweiss: Partially true, thats why I mentioned that it will increase the runtime of every place where these functions are used, however, they are only used in tests where this is ok, I will see if I can either use a customizable predicate(checking for name and/or class and therefore reducing runtime) there or adopt your proposal.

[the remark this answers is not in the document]
Masrepus: There is also a finding function that takes a custom predicate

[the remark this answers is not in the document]
konradweiss: Although I ment one that does not need another call but merges to of my functions to use only predicates and performs only one run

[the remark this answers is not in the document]
konradweiss: Also these functions perform not only finding by predicate but specifically for the find by type name, also cast the returning list to the search type, not the type of the given collection. I will leave it by moving the functions. If someone finds a way to reduce the number of functions without increasing the code in the test files by a to large portion he is welcome to do so.

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverJavaTest.java
konradweiss: I found two places where there is a `filterCast` and then a filter by name, there I decided to adopt the proposal because it is only two places and an additional function would not have been worth the reduction in code duplication.  nested `filterCast` and `findByName`

oxisto:
I am still not quite happy with the test-design but I think it's ok for now. @Masrepus is anyway cleaning up tests so we should merge this and he can rebase his branch on this.

Issue #95, opened 2020-05-08
DFG is not control-flow, context, or field sensitive


konradweiss:
The samples show that the DFG is not control flow sensitive. 
`
int a = 1;
if(args.length > 3){
      a = 2;
}else{
      System.out.println(a);
}
`
The usage of `a` in the print statement has a dfg edge to the declaration. The branch where `a` is used and where `a` gets the value `2` are mutually exclusive. There should be a DFG path from `1` to `a` but not from `2` to `a`.
In other words, the DFG edges currently are drawn intra-transactionally and are not transitive: paths of edges can not be considered to be true positive data-flows.
We could change the DFG-construction to follow the EOG and keep the last definitions of a variable, e.g. right-hand-side of an assignment, and draw edges between those.
So between the right-hand-side of a definition and the references in an expression where the variable is used. 
We have ongoing work where context and field-sensitivity will be solved as well, so this issue could be solved temporarily by the method described above, or by waiting until the more sophisticated approach with push-down systems or slicing is incorporated. 
![dfg_no_context_sensitivity](https://user-images.githubusercontent.com/12885000/81386499-5be07980-9115-11ea-8a25-cd526e322116.png)
![query_no_cfg_sensitivity](https://user-images.githubusercontent.com/12885000/81386506-600c9700-9115-11ea-86ef-abe27af412bc.png)
[DFGSample.zip](https://github.com/Fraunhofer-AISEC/cpg/files/4597807/DFGSample.zip)

johspaeth:
I just read this issue and was curious, when I read context and field-sensitivity. I have been working on this form of data-flow for a while.
 
You may want to looked at https://github.com/CROSSINGTUD/WPDS, it might be adoptable to your problem as well. Feel free to contact me.

konradweiss:
Thanks for your message :) we already took inspiration from the paper that you wrote, and a student is looking to adapt the approach to the CPG. We may come back to you, until then congratulations on the work you did on WPDS and greetings to the IEM.

PR #167, opened 2020-07-16
Add partial support for Control Flow Sensitive DFG

vfsrfs:
Addresses the missing Control Flow Sensitivity of the DFG mentioned in #95, by modifying ingoing and outgoing DFG edges in VariableDeclarations. This does not include Control Flow Sensitivity for FieldDeclarations, as their value might be modified at other locations, not necessarily in the analyzed method. SwitchStatements are supported, but they are trated as if they were always mutually exclusive. Mutual exclusion is not guaranteed, as fall throughs may occur.  In order to incorporate this behavior it is first necessary to be able to obtain the order to the CaseStatemtns, to know the target of a fall through.

Before (#95):
dfg1

After:
dfg2

Code:
```java
public class Main {
  void func() {
    int a = 1;
    if (args.length > 3) {
      a = 2;
    } else {
      System.out.println(a);
    }
  }
}

```


oxisto:
Why is there a DFG node from "1" which is part of the `a=1` to the reference `a` that is part of the if - and no references to the variable declaration?

Masrepus:
That's the modification that was desired to get controlflow sensitive behavior: We don't want value flow to show up at sinks that would never actually be able to see it, e.g. if it happens in mutually exclusive if else branches. So we either need to split up the variable declarations in kind of a single static assignment form, or we modify the DFG edges directly and don't go over the variable declaration. The last approach was the one that was chosen as it stays more true to the actual code graph

oxisto:
@vfsrfs Could you update the picture above with how it will look like after we would merge this? Probably easier to see visually if we are "happy" with the DFG nodes.

vfsrfs:
This PR addresses partial Control Flow Sensitive DFG by modifying ingoing and outgoing DFG edges in VariableDeclarations. If it is not possible to create a more refined DFG, the current DFG edges are not modified. Control Flow Sensitive DFG supports IfStatements and Switchstatements (including Fallthrough).

### Example

#### Code
```java
public class Main {
  void func() {
    int a = 1;
    if (args.length > 3) {
      a = 2;
    } else {
      System.out.println(a);
    }
  }
}
```

#### New Graph
newdfg



## Graph-Changes
### Added

- If mutually exclusive DFG flows are identified, DFG edges from the original node where the DFG flow begins (e.g. LiteralNode) to the DeclaredReferenceExpression where it is used are added.

### Removed

- If we can identify mutually exclusive DFG flows, outgoing DFG edges from the VariableDeclaration to the DeclaredReferenceExpression, where the variable is used, are removed.

## Interface-Changes
None

Issue #158, opened 2020-07-13
Improve documentation of IncompleteType


oxisto:
The semantics for `IncompleteType` is misleading, it seems to be the void type:

https://github.com/Fraunhofer-AISEC/cpg/blob/1024f02164d547f961c0f2f3a6d0642410ef6711/src/main/java/de/fraunhofer/aisec/cpg/graph/type/IncompleteType.java#L8

But why is it incomplete? Should it not rather be just `VoidType` or do I miss something here?

Intended for @vfsrfs.

vfsrfs:
When implenting the TypeParser @titze and I decided to base the naming scheme of the TypeSystem on the C Standard (see [http://www.dii.uchile.cl/~daespino/files/Iso_C_1999_definition.pdf](http://www.dii.uchile.cl/~daespino/files/Iso_C_1999_definition.pdf) on page 33). The C Standard defines the semantics of incomplete type as objects with unknown size, therefore including not only void, but for example also arrays of unknown length. Right now we are using the IncompleteType only for void types and for the remaining use cases we are using the UnknownType. For this reason, it is perfectly conceivable to name VoidType, if this is what you prefer.

oxisto:
Ah, understood. Well another possibility would be to use something like a PrimitiveType for all primitives, such as int and also void. Then void would not have a type just alone for itself. Or something like a BuiltinType.

While I agree that basing it on a standard makes perfect sense, I am not sure whether we want to constrain ourselves with the C99 standard, since a) we do support C++ up to C++17 and there are more built-in types, such as `nullptr` and b) we aim to support a myriad of "newer" languages, such as Java, and possibly Python in the future.

vfsrfs:
I don’t that it is feasible to find a naming scheme that is general enough so that it adheres perfectly to every possible language scheme. At the time of defining the naming scheme for the typesystem we already took into consideration the C++17 Documentation and the Java Documentation. As all of the languages do define types very differently and even within the versions of a languages there are substantial contradictions, e.g. the use of the keyword auto in C++17 vs. C++11. For this reason, we had to make a compromise and we chose the C Standard because of the fact that its definition regarding the types was the most abstract and was most compatible between the three languages (C, C++, Java).

[...]

PR #169, opened 2020-07-16
Fix leaking types

JulianSchuette:
There is one major problem and a few minor ones with the TypeManager:
The major problem is that the singleton instance of TypeManager keeps lists of Types. These Types may refer to RecordDeclarations (including the whole AST of a source file) and were never removed from memory. This is not only a memory leak but also messes up later analysis runs.

The minor problems are related to unclear nullability of fields and the fact that unit tests call TranslationManager.analyze(), but expect that TypeManager is still populated afterwards (although it should in fact be cleaned up). Actually, tests are making false assumptions here.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
@@ -266,6 +266,11 @@ public class TranslationConfiguration {
       return this;
     }
 
+    public Builder disableCleanup() {
+      this.disableCleanup = true;
+      return this;
+    }
+
     /**
      * Adds the specified file to the include blacklist. Relative and absolute paths are supported.
      *
--- a/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
@@ -74,6 +74,13 @@ public class TranslationConfiguration {
    */
   public final List<String> includeBlacklist;
 
+  /**
+   * Switch off cleaning up TypeManager memory after analysis.
+   *
+   * <p>Set this to {@code true} only for testing.
+   */
+  public boolean disableCleanup = false;
+
   /** should the code of a node be shown as parameter in the node * */
   public final boolean codeInNodes;
 
@@ -104,7 +111,8 @@ public class TranslationConfiguration {
       List<String> includeWhitelist,
       List<String> includeBlacklist,
       List<Pass> passes,
-      boolean codeInNodes) {
+      boolean codeInNodes,
+      boolean disableCleanup) {
     this.symbols = symbols;
     this.sourceLocations = sourceLocations;
     this.topLevel = topLevel;
@@ -117,6 +125,7 @@ public class TranslationConfiguration {
     this.passes = passes != null ? passes : new ArrayList<>();
     // Make sure to init this AFTER sourceLocations has been set
     this.codeInNodes = codeInNodes;
+    this.disableCleanup = disableCleanup;
   }
 
   public static Builder builder() {
@@ -166,6 +175,7 @@ public class TranslationConfiguration {
     private List<String> includeBlacklist = new ArrayList<>();
     private List<Pass> passes = new ArrayList<>();
     private boolean codeInNodes = true;
+    private boolean disableCleanup = false;
 
     public Builder symbols(Map<String, String> symbols) {
       this.symbols = symbols;
@@ -326,7 +336,8 @@ public class TranslationConfiguration {
           includeWhitelist,
           includeBlacklist,
           passes,
-          codeInNodes);
+          codeInNodes,
+          disableCleanup);
     }
   }
 }
--- a/src/main/java/de/fraunhofer/aisec/cpg/TranslationManager.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/TranslationManager.java
@@ -106,15 +106,17 @@ public class TranslationManager {
             throw new CompletionException(ex);
           } finally {
             outerBench.stop();
-            log.debug("Cleaning up {} Passes", passesNeedCleanup.size());
-            passesNeedCleanup.forEach(Pass::cleanup);
+            if (!this.config.disableCleanup) {
+              log.debug("Cleaning up {} Passes", passesNeedCleanup.size());
+              passesNeedCleanup.forEach(Pass::cleanup);
 
-            if (frontendsNeedCleanup != null) {
-              log.debug("Cleaning up {} Frontends", frontendsNeedCleanup.size());
-              frontendsNeedCleanup.forEach(LanguageFrontend::cleanup);
+              if (frontendsNeedCleanup != null) {
+                log.debug("Cleaning up {} Frontends", frontendsNeedCleanup.size());
+                frontendsNeedCleanup.forEach(LanguageFrontend::cleanup);
+              }
+
+              TypeManager.getInstance().cleanup();
             }
-
-            TypeManager.getInstance().cleanup();
           }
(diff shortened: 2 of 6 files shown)


oxisto:
Looks like a merge conflict here? We got rid of the `disableCleanup` option and this is now done through mocking in unit tests.

Masrepus:
Mocking away this reset behaviour works even though it is a static method, as the mock replaces the type manager instance. The corresponding trigger is `TestUtils.disableTypeManagerCleanup()`

JulianSchuette:
> Mocking away this reset behaviour works even though it is a static method, as the mock replaces the type manager instance. The corresponding trigger is `TestUtils.disableTypeManagerCleanup()`

I tried that but `reset()` was not called. And this somehow makes sense because Mockito currently replaces the _instance_ , not the static implementation. A quick Google search told me that Mockito alone cannot do that properly but rather PowerMock had to be used.

So I went for the `disableCleanup` solution.

The underlying problem is that tests are expecting content in TypeManager to be available which actually should _not_ be available after the analysis. I guess the tests should be re-written in the longer term.

Masrepus:
Ah, I just saw that there is a difference between reset and cleanup in the TypeManager. I think you should call cleanup from the type resolver, which is non-static and actually clears the type to record mapping as you want it to

Masrepus:
In my opinion we should remove the static reset function and use the cleanup method instead

JulianSchuette:
> Ah, I just saw that there is a difference between reset and cleanup in the TypeManager. I think you should call cleanup from the type resolver, which is non-static and actually clears the type to record mapping as you want it to

It was unclear to me why there is a `cleanup()` and a `reset()` method. But what we actually need is to clean these two lists:

```
  private Set<Type> firstOrderTypes = new HashSet<>();
  private Set<Type> secondOrderTypes = new HashSet<>();
```

`cleanup()` does not do that.

Masrepus:
I see. It seems that when implementing the cleanup method I forgot to clear those as well

oxisto:
Can we change the mocked method to clean those up as well then? I would prefer the mocked-solution to introducing a config parameter that is only used in tests.

Masrepus:
Now I remember why there is a separate reset method. When a test mocks the TypeManager instance, the next test would also get the mock as the new instance. This is why before each test the TypeManager instance is reset to a fresh one. A better solution would probably be to remove this reset method, make the cleanup one really clean up all state, and then in the beforeEach test method to set the TypeManager instance to null, so that it is re-created as it should, upon first use

JulianSchuette:
> Can we change the mocked method to clean those up as well then? I would prefer the mocked-solution to introducing a config parameter that is only used in tests.

TBH, I see this as a quick fix that I need to get Codyze up to date. In the longer term, there are some architectural smells that could be changed:

- `TypeManager` should not need to be a singleton. Singletons are prone to memory leaks, make CPG almost impossible to use in multi-threaded applications, and add complexity (see the cleanup/reset problem).
- Most CPG nodes access the TypeManager singleton in their `typeChanged` method. This tightly couples the CPG nodes to the specific TypeManager. This type of logic should reside outside of the actual CPG, maybe in some visitor.
- Tests should test for the expected behavior and not use reflection magic to inject slightly manipulated objects into TypeResolver at runtime. Either we should have tests that call `TranslationManager.analyze()` - they should expect that everything is cleaned up afterwards. When testing the `TypeManager` itself, tests should create an instance of it and feed it (dummy) data to check if it behaves correctly.

Masrepus:
That's true, although I'm currently unsure how the correct type behavior can be realized without a global singleton "supervisor", e.g. setting the type of a node to the lowest common supertype of all possible subtypes. Maybe inside a very early pass, e.g. the TypeHierarchyResolver?

Concerning the tests, you're right that they should test the type manager on its own, with dummy data

JulianSchuette:
Ah, seems like there are some more broken parts now, possibly caused by the commits in this PR:

```
java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 1
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:458)
	at de.fraunhofer.aisec.cpg.graph.type.TypeParser.createFrom(TypeParser.java:715)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationListHandler.handleSimpleDeclaration(DeclarationListHandler.java:71)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationListHandler.lambda$new$0(DeclarationListHandler.java:52)
	at de.fraunhofer.aisec.cpg.frontends.Handler.handle(Handler.java:112)
	at de.fraunhofer.aisec.cpg.frontends.cpp.StatementHandler.handleDeclarationStatement(StatementHandler.java:267)
	at de.fraunhofer.aisec.cpg.frontends.cpp.StatementHandler.lambda$new$2(StatementHandler.java:53)
	at de.fraunhofer.aisec.cpg.frontends.Handler.handle(Handler.java:112)
	at de.fraunhofer.aisec.cpg.frontends.cpp.StatementHandler.handleCompoundStatement(StatementHandler.java:289)
	at de.fraunhofer.aisec.cpg.frontends.cpp.StatementHandler.lambda$new$0(StatementHandler.java:49)
	at de.fraunhofer.aisec.cpg.frontends.Handler.handle(Handler.java:112)
	at de.fraunhofer.aisec.cpg.frontends.cpp.StatementHandler.handleCatchHandler(StatementHandler.java:105)
	...
```

@Masrepus This is thrown when running `RealBotanTest.testArsenic()`.

Masrepus:
That looks like a problem in the type parsing algorithm, @vfsrfs will know more about this I guess.
What was the String parsed there?

JulianSchuette:
> That looks like a problem in the type parsing algorithm, @vfsrfs will know more about this I guess.
> What was the String parsed there?

Empty string.

Masrepus:
Empty string or does it contain spaces? If it's the latter, I think the problem lies here, it is only checked whether the String has length 0 but not whether it only consists of whitespaces: https://github.com/Fraunhofer-AISEC/cpg/blob/ff3dd6d87437c2a2995f7a9a9c566be03041395d/src/main/java/de/fraunhofer/aisec/cpg/graph/type/TypeParser.java#L666-L669

JulianSchuette:
> Empty string or does it contain spaces? If it's the latter, I think the problem lies here, it is only checked whether the String has length 0 but not whether it only consists of whitespaces:

Yes, that was one problem. It's fixed now.

Another problem is the type string `const` which comes from the source line `const auto error = QString{e.what()};`.

JulianSchuette:
"Fixed" also that problem. What is the expected behavior of "auto" types? Seems like currently they are not supported. This PR now bails with an `UnknownType`. Is that intended?

Possibly related to issue #158

Masrepus:
I would say unknown is correct, as the behavior is the same as a `var` in Java -> The type comes from the right hand side of the assignment and is thus correctly propagated if the left hand side is of type unknown

JulianSchuette:
Yes, right. I just found this piece:

```
    if (finalType.getTypeName().equals("auto") || (type.contains("auto") && !primitiveType)) {
      // In C++17 if auto keyword is used the compiler infers the type automatically, hence we
      // are not able to find out, which type this should be, it will be resolved due to
      // dataflow
      return UnknownType.getUnknownType();
    }
```

JulianSchuette:
Good, for the moment this PR should do the job. @konradweiss  would you merge it?

PR #170, opened 2020-07-16
Simple try to make analysis run in parallel

oxisto:
Just a draft for now. This is a very basic implementation of a parallel *frontend* run. Passes will be executed after all frontends have run and then passes will run after each other.

Problem: ScopeManager loses its context after each frontend run now, since I just created a new ScopeManager for each frontend. This is probably not a problem currently, since all passes do not use the scope manager information anyway but will be a problem, once @konradweiss re-organises the scope manager. So it might be worth having in mind that the scope manager should be 

The type manager and parser are probably not thread-safe as well.


oxisto:
Creating a new `GlobalScope` in `analysis` now. This should pretty much simulate the behaviour of the cpg that we have so far. The common global scope should hold everything that needs to be accessed from multiple translation units. Then scope managers exist for each TU. In the future, also the type manager should probably register its "global" to this `GlobalScope`.

This scope could then also be propagated to the passes, so that they can access the gathered information of the scope manager, since (theoretically), the scopes should still be associated to the global scope as children.

konradweiss:
It should be possible to merge the scope information that every pass builds in parallel, the ScopeManager keeps multiple trees of scopes and there's trees could start at every translation unit.

PR #173, opened 2020-07-16
Fix parsing error when elaborated types are used as generics

### LINKED ISSUE #171: C++ generics with elaborated type specifiers fail to parse
The following: `TypeParser.createFrom("Array<struct Node>", true)` will create a generic type with a type parameter of type `"structNode"`. This seems to happen here, because of some reason this function here will remove all the whitespace within the generic block. Therefore, the type parser cannot correctly identity the `struct` as an elaborated type specifier (which is normally ignored). However, simply removing the function also does not work since then it is parsed as seperate typeblocks and things really start to go wrong :(

https://github.com/Fraunhofer-AISEC/cpg/blob/3f3542015dd6c989b8f8274b89478dfe67922126/src/main/java/de/fraunhofer/aisec/cpg/graph/type/TypeParser.java#L261-L300

vfsrfs:
fixes #171

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/graph/type/TypeParser.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/graph/type/TypeParser.java
@@ -273,6 +273,15 @@ public class TypeParser {
     if (type.contains("<") && type.contains(">") && getLanguage() == TypeManager.Language.CXX) {
       String generics = type.substring(type.indexOf('<') + 1, type.lastIndexOf('>'));
 
+      /* Explanation from @vfsrfs:
+       * We fist extract the generic string (the substring between < and >). Then, the elaborate
+       * string can either start directly with the elaborate type specifier e.g. struct Node or it
+       * must be preceded by <, \\h (horizontal whitespace), or ,. If any other character precedes
+       * the elaborate type specifier then it is not considered to be a type specifier e.g.
+       * mystruct. Then there can be an arbitrary amount of horizontal whitespaces. This is followed
+       * by the elaborate type specifier and at least one more horizontal whitespace, which marks
+       * that it is indeed an elaborate type and not something like structMy.
+       */
       for (String elaborate : elaboratedTypes) {
         generics = generics.replaceAll("(^|(?<=[\\h,<]))\\h*(?<main>" + elaborate + "\\h+)", "");
       }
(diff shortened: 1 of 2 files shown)


src/main/java/de/fraunhofer/aisec/cpg/graph/type/TypeParser.java
```diff
@@ -263,6 +270,20 @@ private static boolean isUnknownType(String typeName) {
    */
   @NonNull
   private static String fixGenerics(@NonNull String type) {
+    if (type.contains("<") && type.contains(">")) {
+      String generics = type.substring(type.indexOf('<') + 1, type.lastIndexOf('>'));
+      for (String elaborated : elaboratedTypes) {
+        if (getLanguage() == TypeManager.Language.CXX && generics.contains(elaborated + " ")) {
+          generics = generics.replace(elaborated + " ", "");
```
oxisto, line 277: Not quite sure about this solution. Wouldn't the following fail?
`Array<myclass >`

oxisto: Can't we just pipe between `<>` through  `TypeParser.createFrom` BEFORE splitting the outer type? Then you would not need to remove white-spaces at all?

oxisto:
I'll let @konradweiss decide this ;-) In my opinion it would be better overhaul the `createFrom` function to fix a few issues with type parsing, i.e. if open and closed brackets are not matching and so on, rather than have yet another fix based on a replace with an obscure regex.

vfsrfs:
I agree. Nonetheless, this was intended as a preliminary fix so that it works for now. A rework for the `TypeParser.createFrom` will be necessary anyways regarding the brackets matching and at that point we can address this as well.

oxisto:
Could you at least provide some kind of comment what the regex does. To be honest, I have no clue :D

vfsrfs:
Sure. We fist extract the generic string (the substring between `<` and `>`). Then, the elaborate string can either start directly with the elaborate type specifier e.g. `struct Node` or it must be preceded by `<`, `\\h` (horizontal whitespace), or `,`. If any other character precedes the elaborate type specifier then it is not considered to be a type specifier e.g. `mystruct`. Then there can be an arbitrary amount of horizontal whitespaces. This is followed by the elaborate type specifier and at least one more horizontal whitespace, which marks that it is indeed an elaborate type and not something like `structMy`.

I will also document this in the code.

PR #182, opened 2020-07-28
Restored documented behaviour of loadIncludes

### LINKED ISSUE #181: Clarify behaviour of loadIncludes
https://github.com/Fraunhofer-AISEC/cpg/blob/6af6499ae50cfc1ea8b51631a0d17bcf146b415a/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java#L58-L59

According to the documentation of loadIncludes, includes such as header files will be parsed but not included to the CPG. However, at first glance it seems that that either the  documentation or the behaviour is incorrect 

If loadInclude is set to false, the include provider for CDT will be set to an empty files provider, effectively hindering CDT to parse the header file at all. Or do I miss another step where they are considered?

So this looks ok-ish (although I am not sure if `getContainingFilename` will point to the header file if it is included through a chain:

https://github.com/Fraunhofer-AISEC/cpg/blob/9231e141b9b5b043fd59371ce21b80a52c7124ab/src/main/java/de/fraunhofer/aisec/cpg/frontends/Handler.java#L82-L90

But this does not look ok then:

https://github.com/Fraunhofer-AISEC/cpg/blob/ab802001b88dd1def0914f0c46bee6e86a60f415/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontend.java#L275

_Originally posted by @oxisto in https://github.com/Fraunhofer-AISEC/codyze/pull/35#issuecomment-664987729_

oxisto:
Closes #181

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/Handler.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/Handler.java
@@ -77,15 +77,14 @@ public abstract class Handler<S, T, L extends LanguageFrontend> {
       return null;
     }
 
-    // If we do not want to load includes into the CPG and the current fileLocation is an include
-    // file ->
+    // If we do not want to load includes into the CPG and the current fileLocation was included
     if (!this.lang.config.loadIncludes && ctx instanceof ASTNode) {
       ASTNode astNode = (ASTNode) ctx;
-      for (String inclPath : this.lang.config.includePaths) {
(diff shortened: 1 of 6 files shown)


oxisto:
Note: This ignores *all* included files, not just includes that are specified on the `includePath`. Which kind of behaviour do we want? Ignore all includes or only ignore "extra" includes, i.e. specified in `includePath` but NOT ignore "local" includes, i.e. that are within the same top level directory.

PR #193, opened 2020-08-05
Fix for null annotations

### LINKED ISSUE #192: Field `annotations` must not be null, breaks OGM
Commit 7d56eda487e8ec408b13036581d08b5e9800ff3e breaks the object graph mapper (OGM) of Codyze (and possibly Neo4J) and makes CPG structure ambiguous.

Collection fields of `Node`s must not be null.

Fields of classes in the CPG structure will result in either _edges_ or _properties_  of a node. The clearest way is to make that explicit by using the `@Relationship` annotation to enforce an edge. If that is not present, the OGM will investigate the value of the respective field and see if it is a `Node` type or Collection of `Node`s. This requires the field to be non-null, otherwise the OGM will consider it a property on the one hand, but not be able to provide type information on the other hand.

In its current form, Codyze will run into many `IllegalStateException`s.

JulianSchuette:
Fixes #192

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/graph/Node.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/graph/Node.java
@@ -29,11 +29,7 @@ package de.fraunhofer.aisec.cpg.graph;
 import de.fraunhofer.aisec.cpg.helpers.LocationConverter;
 import de.fraunhofer.aisec.cpg.processing.IVisitable;
 import de.fraunhofer.aisec.cpg.sarif.PhysicalLocation;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 import org.checkerframework.checker.nullness.qual.NonNull;
@@ -96,8 +92,7 @@ public class Node extends IVisitable<Node> {
   @Relationship(value = "DFG")
   protected Set<Node> nextDFG = new HashSet<>();
 
-  @NonNull
-  protected Set<TypedefDeclaration> typedefs = new HashSet<>();
+  @NonNull protected Set<TypedefDeclaration> typedefs = new HashSet<>();
 
   /**
    * If a node is marked as being a dummy, it means that it was created artificially and does not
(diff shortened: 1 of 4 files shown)


oxisto:
My original though was to set it to `null` to avoid creation of many list objects in the in-memory graph if the annotation processing feature is turned off. Can you give an example of the `IllegalStateException`s that codyze gives? I was under the impression that I "fixed" this in https://github.com/Fraunhofer-AISEC/codyze/pull/41

src/main/java/de/fraunhofer/aisec/cpg/graph/Node.java
```diff
@@ -116,7 +120,7 @@
 
   /** List of annotations associated with that node. */
   @SubGraph("AST")
-  protected List<Annotation> annotations;
+  public List<Annotation> annotations = new ArrayList<>();
```
oxisto, line 123: Why is this member public?

src/main/java/de/fraunhofer/aisec/cpg/graph/Node.java
```diff
@@ -259,6 +267,15 @@ public boolean isImplicit() {
     return this.implicit;
   }
 
+  @NonNull
+  public List<Annotation> getAnnotations() {
+    return annotations;
+  }
+
+  public void setAnnotations(@NonNull List<Annotation> annotations) {
```
oxisto, line 275: The only issue that I have with this, is that we create two list for each node, if `processAnnotations` is enabled. Once for each node in the init and another one in the `CXXLanguageFrontend::handleAttributes` function which is then used in this setter here. We should probably follow a similar pattern like the DFG and EOG properties that *add* to it (or clear it) rather than creating a new list.

JulianSchuette: Agree, re-creating the collections is not expensive, but we should keep it consistent and possibly even make collection fields `final`  one day.

src/main/java/de/fraunhofer/aisec/cpg/graph/Node.java
JulianSchuette: All members are public to allow reflection access from outside of the module without having to use `--add-opens` workarounds in Java >= 9.

oxisto: All the other members are protected?

JulianSchuette: Err... Anyway, it's now consistent with the other fields :-)

PR #198, opened 2020-08-14
Carry over updated ScopeManager and NamespaceDeclaration from outdate…

konradweiss:
…d Branch

Updates to the ScopeManager and allows namespace declarations to contain static fields and functions. This is an extract of the old branch that was supposed to contain various fixes to variables resolution that then got a bit out of hand.

Variable resolution should work properly if the frontend calls addDeclaration in the right places. I will be able to test this once the PR is merged and I can work on integrating the ScopeManager into the VariableResolution.


src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
```diff
@@ -109,7 +109,8 @@ private Declaration handleNamespace(CPPASTNamespaceDefinition ctx) {
     lang.getScopeManager().enterScope(declaration);
     for (IASTNode child : ctx.getChildren()) {
       if (child instanceof IASTDeclaration) {
-        declaration.add(this.lang.getDeclarationHandler().handle((IASTDeclaration) child));
+        lang.getScopeManager()
```
oxisto, line 112: I think this is the only occurrence where adding a declaration is done by the scope manager and not by the translation manager. We should keep things consistent

src/main/java/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontend.java
```diff
@@ -119,25 +119,36 @@ public TranslationUnitDeclaration parse(File file) throws TranslationException {
...
         declaration = namespaceDeclaration;
       }
 
       for (TypeDeclaration<?> type : context.getTypes()) {
-        declaration.add(getDeclarationHandler().handle(type));
+        Declaration typeD = getDeclarationHandler().handle(type);
+        if (declaration instanceof TranslationUnitDeclaration) {
+          ((TranslationUnitDeclaration) declaration).add(typeD);
+        } else {
+          scopeManager.addDeclaration(typeD);
```
oxisto, line 141: Not sure I understand this change (and the ones below and above). In which cases is it not a Translation Unit ?! Line 119 specifically creates a translation unit decl.

oxisto:
BTW this branch also fails on SonarCloud: https://sonarcloud.io/dashboard?branch=konrad%2Fupdate-scope-manager-namespaces&id=Fraunhofer-AISEC_cpg 

Seems to be the old bug again, where SonarQube will only post its information in the PR after you commit something in the PR.

src/main/java/de/fraunhofer/aisec/cpg/graph/FunctionDeclaration.java
```diff
@@ -50,6 +50,12 @@
   @SubGraph("AST")
   protected Statement body;
 
+  /**
+   * Classes and Structs can be declared inside a function and are only valid within the function.
+   */
+  @SubGraph("AST")
+  protected List<RecordDeclaration> records = new ArrayList<>();
```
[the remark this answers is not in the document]
konradweiss: Good point especially because they are already contained in the AST of the declaration. Hopefully.

src/main/java/de/fraunhofer/aisec/cpg/graph/NamespaceDeclaration.java
```diff
@@ -26,4 +26,104 @@
...
+ * appearance a namespace declaration can contain {@link FunctionDeclaration}, {@link
+ * FieldDeclaration} and {@link RecordDeclaration} similar to a {@link RecordDeclaration} and the
+ * semantical difference between NamespaceDeclaration and {@link RecordDeclaration} lies in the
+ * non-instantiabillity of a namespace.
+ */
+public class NamespaceDeclaration extends Declaration {
+
+  /** Edges to a {@link FieldDeclaration} defined in a namespace, these fields are static. */
+  @SubGraph("AST")
+  private List<FieldDeclaration> fields = new ArrayList<>();
```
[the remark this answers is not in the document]
konradweiss: Why does this argument not hold for records? You could argue that a Record should only have a declarations field. I don't like the utility function because those are not available to graph queries later. Having edges that encode the specific type of declaration is not a problem IMO. However, I see the argument that ProblemDeclarations, which will be present wherever we have declarations meaning we have to switch to default declarations for every Node that contains declarations including RecordDeclarations.

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
konradweiss: Yes, and consistently add them over the scope Manager. When a Declaration is processed, context knowledge is needed to know to which node a declaration has to be added. This is the exact purpose of a Manager whose job it is to keep track of scopes.
This is the only occurrence in `DeclarationHandler`, however, there are many more in the `DeclaratorHandler`.

src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/ScopeManager.java
```diff
@@ -364,80 +380,191 @@ public LabelStatement getLabelStatement(String labelString) {
...
+      name =
+          name.substring(
+              name.lastIndexOf(
+                  lang.getNamespaceDelimiter() + lang.getNamespaceDelimiter().length()));
+      declaration.setName(name);
+
+      addDeclarationInCurrentScope(declaration);
+      currentScope = tmpScope;
+    }
+    */
```
Masrepus, line 546: I would rather just delete all of this commented out code

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
Masrepus: If we consistently add them via the scope manager that's fine for me :+1:

src/main/java/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontend.java
konradweiss: We would not have to make a difference here if there was a scope for `TranslationUnitDeclaration`, which may actually be something that we can introduce in the future. Previously a NamespaceDeclaration extended from a TranslationUnit, with this PR NamespaceDeclarations are no longer TranslationUnitDeclarations, which IMO they shouldn't be. In java, for a long time now we had a `TranslationUnitDeclaration` containing a `NamespaceDeclaration` which then contains the rest of the file. This served the purpose of unifying how packages and namespaces work.

So basically this is for the case that the condition that there is a `packageDecl` and the statement in line 133 is executed `declaration = namespaceDeclaration;`.

src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/ScopeManager.java
konradweiss: Yes that was something I was not sure about because of me potentially needing it for future PRs but I guess I will keep my local branch where I still have these implementations and remove it for the sake of code quality.

src/main/java/de/fraunhofer/aisec/cpg/graph/NamespaceDeclaration.java
```diff
@@ -26,4 +26,104 @@
...
+
+/**
+ * Declares the scope of a namespace and appends its own name to the current namespace-ürefix to
+ * form a new namespace prefix. While RecordDeclarations in C++ and Java have their own namespace,
+ * namespace declarations can be declared multiple times. At the beginning of a Java-file, a
+ * namespace declaration is used to represent the package name as namespace. In its explicit
+ * appearance a namespace declaration can contain {@link FunctionDeclaration}, {@link
+ * FieldDeclaration} and {@link RecordDeclaration} similar to a {@link RecordDeclaration} and the
+ * semantical difference between NamespaceDeclaration and {@link RecordDeclaration} lies in the
+ * non-instantiabillity of a namespace.
```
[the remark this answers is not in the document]
konradweiss: For me, a Translation unit is clearly only for a file, or a piece of code, so I do see a Namespace close to a `RecordDeclaration` especially because of RecordDeclarations adding "qualification" to a type name. As mentioned, the difference lies in the non-instantiabillity, which means that a Namespace can only have the static variants of fields and functions.

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
oxisto: Then at least `addDeclaration` should contain a javadoc containing information that this not only adds the declaration to the scope but also to the ast node containing that scope. I am still not a big fan of this because this way you hide the behaviour deep down in the scope manager but at least it should be documented and consistent.

oxisto: Update: Ok it looks like we are already doing it this way in the other handlers, looks like this one snuck by me :) Then let's do it this way - although I am not a fan, but please add documentation to the call.

src/main/java/de/fraunhofer/aisec/cpg/graph/NamespaceDeclaration.java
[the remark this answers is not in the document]
Masrepus: This is probably something that we can improve once we have edge properties. Then we can annotate the declarations edges with the correct type (at least I guess that this will be possible :man_shrugging:)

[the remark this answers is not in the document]
oxisto: I am not sure if this will be at all possible in Codyze because of OverflowDB. Afaik edges only exist "virtually" in overflowdb and I am not sure if they support properties on edges.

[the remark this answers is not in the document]
Masrepus: Edges do have properties there, if you look at the following code:

https://github.com/Fraunhofer-AISEC/codyze/blob/51ce074e262009234cf82a5e9aa72acc9b909f7f/src/main/java/de/fraunhofer/aisec/crymlin/connectors/db/OverflowDatabase.java#L968-L1000

At the moment this is done via an `@EdgeProperty` annotation that is present in the CPG, but I guess we could use different mechanisms with more fine-grained control, e.g. some kind of wrapper like `List<AnnotatedNode>` where an `AnnotatedNode` consists of a target node and a list of edge properties, which is then unpacked in the codyze OGM. At least OverflowDB edges do have the possibility to assign properties to each one

PR #210, opened 2020-08-25
Handling void function parameters

### LINKED ISSUE #208: How to deal with void being indicator for no arguments?
`void` can be used  to indicate that a function does not take any arguments, e.g. the following functions both have the same signature (no arguments):

```
void foo(void);
void foo2();
```

However, the graph shows that the function `foo` contains a parameter of type `void`. This may be also the case when dealing with function pointers.

argument_graph

Is this the desired behavior in order to stay as close as possible to the source code or would it instead be better to clearly indicate that the function `foo` does not take any argument by removing the void parameter in the graph?

oxisto:
Fixes #208 

How should we deal with error cases? For example, specifying `void` more than once is not allowed and also specifying `void` in a named parameter is not allowed. CDT does not enforce this as it just produces a parse tree.

For now, I just emitted a warning but continued parsing the erroneous cases.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java
@@ -26,6 +26,7 @@
 
 package de.fraunhofer.aisec.cpg.frontends.cpp;
 
+import static de.fraunhofer.aisec.cpg.helpers.Util.warnWithFileLocation;
 import static java.util.Collections.emptyList;
 
 import de.fraunhofer.aisec.cpg.frontends.Handler;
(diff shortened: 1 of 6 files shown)


Masrepus:
A warning is probably good enough for us. If somebody really gives us multiple void parameters, this is a very weird situation anyway.

What should be added is another fix inside the type parser: Function pointer signatures can also have void in their parameter list. I think having the fix for this in the same pull request would be nice and tidy. Otherwise we can of course create a separate issue for that aspect of the problem.

Masrepus (approved):

oxisto:
Makes sense. I will check if I can provide a fix for the type parser

konradweiss:
> e a fix for the type parser

@oxisto Are you going to add this is this PR or create a separate issue? Otherwise, I can merge the PR.

oxisto:
> > e a fix for the type parser
> 
> @oxisto Are you going to add this is this PR or create a separate issue? Otherwise, I can merge the PR.

I'll quickly see if I can provide this today in this PR otherwise we can merge it tomorrow.

PR #213, opened 2020-08-27
Make MemberExpression subclass of DeclaredReferenceExpression

### LINKED ISSUE #151: Fields of member expressions have an empty location
Example code:
```
void SomeClass::SomeMethod()
{
	fieldObject->anotherField = true;
}
```

The `fieldObject->anotherField`  is correctly interpreted as a `MemberExpression`, but the member `anotherField` has an empty location

### LINKED ISSUE #190: MemberExpression base is not correctly resolved
Currently the resolver expects a base to be a field. This is incorrect, as we can also have normal variables etc here

Masrepus:
Closes #151 and closes #190 

## Graph Changes
- `MemberExpressions` are now `DeclaredReferenceExpressions` that additionally include a base expression. Thus the old member expression has been removed and replaced by standard `refersTo` edges.
- `MemberExpression` base is now a `DeclaredReferenceExpression` that is subsequently connected to the field or variable it refers to, as opposed to the node-switching magic that had happened before
- `StaticReferenceExpression` is gone. Its replacement is the `staticAccess` flag inside `DeclaredReferenceExpression`. This is cleaner than before, as `StaticReferenceExpressions` had no further use other than the meaning of this new flag.

## API Changes
- The `refersTo` collection of a `DeclaredReferenceExpression` is no longer limited to `ValueDeclaration` but also accepts `Declarations`. This is due to the fact that bases of `MemberExpressions` can refer to `RecordDeclarations`.
- `DeclaredReferenceExpression.setRefersTo(Declaration)` can now also handle null values for convenience. They are simply discarded if null
- Converted `IVisitable` to an interface


src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
```diff
@@ -319,15 +319,7 @@ private void handleCallExpression(@NonNull Node node) {
...
-    // analyze the base
-    if (memberExpression.getBase() instanceof Statement) {
-      createEOG(memberExpression.getBase());
-    }
-
-    // analyze the member
-    if (memberExpression.getMember() instanceof Statement) {
-      createEOG(memberExpression.getMember());
-    }
+    // TODO how to handle base?
```
oxisto, line 322: call `createEOG` to the base?

oxisto (commented):
I like it. The only thing that confuses me a little bit... Do you have an example where member expressions refer to record declarations?

I thought they could only refer to declarations that contains a value, hence the set of value declarations.

src/main/java/de/fraunhofer/aisec/cpg/graph/DeclaredReferenceExpression.java
```diff
@@ -43,63 +44,82 @@
 public class DeclaredReferenceExpression extends Expression implements TypeListener {
 
   /** The {@link ValueDeclaration}s this expression might refer to. */
-  private Set<ValueDeclaration> refersTo = new HashSet<>();
+  private Set<Declaration> refersTo = new HashSet<>();
```
oxisto, line 47: See my other comment. I thought `ValueDeclaration` is fine for the reference.

src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
oxisto, line 322: Do we explicitly create the EOG to the refers or or is already done by handleDeclaredReferenceExpression? Otherwise we need to create EOG to refers and to base.

src/main/java/de/fraunhofer/aisec/cpg/graph/DeclaredReferenceExpression.java
```diff
@@ -43,63 +44,82 @@
...
-            this.removeNextDFG(r);
-            this.removePrevDFG(r);
-          }
-          r.unregisterTypeListener(this);
-          if (r instanceof TypeListener) {
-            this.unregisterTypeListener((TypeListener) r);
-          }
-        });
+  public void setRefersTo(@NonNull Set<Declaration> refersTo) {
+    this.refersTo.stream()
```
oxisto, line 83: So we are filtering only for ValueDeclarations anyway? So I do not see the need for settings the set to declarations.

Masrepus:
Things like `SomeClass.staticField` is also a member expression, with a RecordDeclaration as the base. Hence the generalization.

The case where we filter out all non-ValueDeclarations is only when we register type listeners, which only works with ValueDeclarations.

oxisto:
> Things like `SomeClass.staticField` is also a member expression, with a RecordDeclaration as the base. Hence the generalization.

Yes, but that won‘t affect the `refersTo` field, right?
> 
> The case where we filter out all non-ValueDeclarations is only when we register type listeners, which only works with ValueDeclarations.

Masrepus:
Ah I see what you mean. It affects the refersTo field indirectly, as this comes from DeclaredReferenceExpression, and the base being a DeclaredReferenceExpression as well requires those to support all Declarations. For a MemberExpression's refersTo this would not strictly be needed, but I guess it's better if all DeclaredReferenceExpressions can link to any Declaration instead of inventing a new node type for the base

oxisto:
Ah, I see, you are correct. In this case it of course makes sense.

src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
Masrepus: If I understand the mechanics correctly, the handleDeclaredReferenceExpression won't be called if there is a handler for MemberExpression. So we'll have to handle all parts of the EOG here. @konradweiss what would be the optimal EOG pattern for MemberExpressions?

konradweiss: @Masrepus As the member in theory can contain an Expression, although it was pointed out by @oxisto that not all expression are allowed, the member and the base need to be evaluated and therefore an edge is needed to both, the order can be found in the old code as far as I known. handleMemberExpression should be called first by the handler, as it is more specific than handleDeclaredReferenceExpression. Does this answer the question?

oxisto:
Might I suggest that we converted `IVisitable` to an interface? This would reduce the hierarchy

Masrepus:
That's a good idea

PR #218, opened 2020-08-28
Converting refersTo back to a single edge field for convenience

### LINKED ISSUE #217: Investigate conversion of DeclaredReferenceExpression.refersTo back to a single edge
Currently, the `refersTo` field of `DeclaredReferenceExpressions` is a set, but originally it was a single node. The change was introduced to support function pointer resolution at a time where we didn't have full function pointer signatures in their type, so the `refersTo` had to be set to all functions of a specific name. Picking the correct target was only possible once the actual call was made. This is no longer the case, so unless we have any other cases where we would need multiple `refersTo` edges, we should migrate back to it being a single edge

oxisto:
Closes #217 

# Graph Changes

The proposed change *should* actually not have a change in the graph structure, because the relationship of the field stays the same, even though the Java representation has changed.

# API Changes

The `refersTo` field of a `DeclaredReferenceExpression` is no longer a collection, but rather a single reference to a `ValueDeclaration`.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/graph/DeclaredReferenceExpression.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/graph/DeclaredReferenceExpression.java
@@ -33,6 +33,7 @@ import java.util.Objects;
 import java.util.Set;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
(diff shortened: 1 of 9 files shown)


oxisto:
First shot looks good, I have yet to convert the getter function, since a lot of tests use it and I just quickly wanted to check if it works, so the getter returns a temporary set of the declaration.

Also probably this will create some merge conflicts with #213, so I guess it would make sense to merge #213 first.

oxisto:
In discussion with @Masrepus, we first merge this one, then #213

PR #220, opened 2020-08-28
Executing filename as latest pass and mark file field as Nullable

### LINKED ISSUE #219: Wrong order of passes leads to empty file field in Node
The `file` field is neither annotated as `@NonNull`, nor `@Nullable`, however, codyze *assumes* that it is non-null. The `file` property is set by the `FilenameMapper`, a pass which is executed as first. 

By creating dummy AST nodes in later passes, these nodes do not get their `file` property set. So two things are needed:

- Decide, whether we want to ensure `@NonNull` on the `file` property
- Move the `FilenameMapper` to a later stage in the passes, probably at the latest.

oxisto:
Fixes #219


oxisto:
It seems that we (maybe always?) have a circular AST reference problem, which should not be there. It seems to be created by the `VariableUsageResolver`. For now, this was not an issue since the `FilenameMapper`, which completely traverses the AST was executed before.

PR #221, opened 2020-08-29
Codyze PR check

oxisto:
This PR adds a check in parallel to a regular build which tries to build the PR branch against codyze.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,23 @@ on:
   - pull_request
 
 jobs:
+  check_pr:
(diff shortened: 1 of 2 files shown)


oxisto:
This will (intentionally) fail now, until #213, #218 and #220 are merged, a new version is released and used in Codyze.

Possibly, we should also wait for https://github.com/Fraunhofer-AISEC/codyze/pull/64 for faster Codyze tests.

oxisto:
Another idea might be to just trigger the codyze PR check if a certain label has been assigned to a PR.

Issue #212, opened 2020-08-26
Parsing elaborated type specifiers is not supported (yet)


oxisto:
For now we just mostly skip over these. I think we correctly parse them in the type parser but we do not turn them into declarations if needed (i.e. the first time they are used).

In my usage scenario I have a lot of those, so I will have a look at that

PR #223, opened 2020-08-30
Bugfix/variable resolution

konradweiss:
This solves various problems in the current Variable resolution for java and C++.
#### Added
* A definition edge for fields as not only methods can have definitions and declarations(Still missing for RecordDeclarations etc.)


konradweiss:
I still need to solve problems for some other tests that fail now. However, I expect that we only need to adapt the tests

oxisto:
I was working on something similar, mainly to address #212. The idea that I had was to have a `Redeclarable` class or interface that defines whether something can be redeclared and had a definition. Seems I have never pushed the branch, I can do so later and maybe you can recycle certain aspects of it.

Not sure about field declarations though, is there a scenario where we need to distinguish between def and decl here?

Masrepus:
This branch is now up to date and just needs a few fixes regarding the 3 failing resolver tests. Then we're good to go to merge this

konradweiss:
> I was working on something similar, mainly to address #212. The idea that I had was to have a `Redeclarable` class or interface that defines whether something can be redeclared and had a definition. Seems I have never pushed the branch, I can do so later and maybe you can recycle certain aspects of it.
> 
> Not sure about field declarations though, is there a scenario where we need to distinguish between def and decl here?

Is there a difference between functions and fields? I would have said that we just do it consistently.

Masrepus:
Ready for review @konradweiss @oxisto

gradlew.bat
```diff
@@ -29,7 +29,7 @@ if "%DIRNAME%" == "" set DIRNAME=.
 set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
-@rem Resolve any "." and ".." in APP_HOME to make it shorter.
+@rem Resolve any  and ".." in APP_HOME to make it shorter.
```
oxisto, line 32: Maybe edited per mistake? Should not be part of this commit

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java
```diff
@@ -89,14 +90,36 @@ private Declaration handleDeclarator(CPPASTDeclarator ctx) {
...
+      RecordDeclaration recordDeclaration = this.lang.getRecordForName(recordName).orElse(null);
+
+      declaration = FieldDeclaration.from((VariableDeclaration) declaration);
+      if (recordDeclaration != null) {
+        Optional<FieldDeclaration> existingDeclaration =
+            recordDeclaration.getFields().stream()
+                .filter(f -> f.getName().equals(fieldName))
+                .findFirst();
+        if (existingDeclaration.isPresent()) {
+          existingDeclaration.get().setDefinition((FieldDeclaration) declaration);
```
oxisto, line 120: As mentioned in the comments, can you provide an example for this? I am still not sure if field declarations can have a definition.

src/main/java/de/fraunhofer/aisec/cpg/graph/FieldDeclaration.java
```diff
@@ -43,6 +44,13 @@
   @Nullable
   private Expression initializer;
 
+  /** Specifies, whether this field declaration is also a definition, i.e. has an initializer. */
+  private boolean isDefinition = false;
+
+  /** If this is only a declaration, this provides a link to the definition of the field. */
+  @Relationship(value = "DEFINES")
+  private FieldDeclaration definition = this;
```
oxisto, line 52: see comment above, I am not convinced :D

src/main/java/de/fraunhofer/aisec/cpg/graph/FieldDeclaration.java
```diff
@@ -88,6 +96,7 @@ public Expression getInitializer() {
 
   public void setInitializer(Expression initializer) {
     if (this.initializer != null) {
+      setIsDefinition(true);
```
oxisto, line 99: Ah, I see your point, but in my opinion also variables that are not explicitly initialized are actually "defined" (or to put it the other way around there is no difference between declaring and defining a variable).

If it has no initialiser it might have a default value, such as `0` or `null`.

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverCppTest.java
```diff
@@ -341,7 +368,7 @@ void testAccessStaticMemberThroughInstanceFirst() {
     VRUtil.assertUsageOfMemberAndBase(
         callParamMap.get("func5_staticVarName_throughInstance_first"),
         declaration,
-        outerStaticVarName); // external_static_field
+        outerStaticVarName.getDefinition()); // external_static_field
```
oxisto, line 371: Ah! external static fields might be a use case for declaration vs. definition of fields, but not every un-initialized field is undefined

oxisto (commented):
I think we may need a quick teams discussion on what nodes can be 'defined' vs. 'declared'.

src/main/java/de/fraunhofer/aisec/cpg/graph/FieldDeclaration.java
Masrepus: Or even worse, undefined values that can indeed be used later on. This is in contrast to functions that can't be called when there is no definition. So I would prefer regarding fields as always being auto-defined, removing the distinction between declaration and definition.

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java
```diff
@@ -381,15 +397,16 @@ private void processMembers(
...
+            .addDeclaration(FieldDeclaration.from((VariableDeclaration) declaration));
+        lang.getScopeManager().removeDeclaration(declaration);
       } else if (declaration instanceof FieldDeclaration) {
-        recordDeclaration.getFields().add((FieldDeclaration) declaration);
+        // recordDeclaration.getFields().add((FieldDeclaration) declaration);
+        lang.getScopeManager().addDeclaration(declaration);
       } else if (declaration instanceof RecordDeclaration) {
-        recordDeclaration.getRecords().add((RecordDeclaration) declaration);
+        // record is not stored as reference in the scope
+        lang.getScopeManager().addDeclaration(declaration);
```
Masrepus, line 409: Same question as in previous PRs: Do you need the commented-out lines or can we remove them?

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverCppTest.java
```diff
@@ -34,14 +34,24 @@
...
+// Todo VariableResolverPass 11 Failed, 17 Passed after adding member declarations with the scope
+// Todo VariableResolverPass 6 Failed, 22 Passed after adding declaration definition pattern of
+// Todo VariableResolverPass 4 Failed, 24 Passed after searching fields by their simple name and
+// extracting the base class from the prefix
+// methods to fields
+// manager
+
+// @Disabled(
+//    "Until changing variable resolution to ScopeManager. Then in detail disable the tests that
+// need specific fixes")
```
Masrepus, line 54: This can be removed now that everything is done

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverJavaTest.java
```diff
@@ -29,14 +29,17 @@
...
 // Todo VariableResolver 7 Failed, 7 Passed
+// Todo VariableResolver 6 Failed, 8 Passed after correcting a bug in the tests themselves
+// Todo VariableResolver 5 Failed, 9 Passed after defaulting unknown types to be in the same package
+// Todo Solved all after fixing order on stack in ScopeWalker
 
-@Disabled(
-    "Until changing variable resolution to ScopeManager. Then in detail disable the tests that need specific fixes")
+// @Disabled(
+//    "Until changing variable resolution to ScopeManager. Then in detail disable the tests that
+// need specific fixes")
```
Masrepus, line 42: Can be removed

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverJavaTest.java
```diff
@@ -218,6 +231,7 @@ void testVarNameInSecondLoop() {
 
   @Test
   void testImplicitThisVarNameAfterLoops() {
+    // Todo here we get a heisenbug, this test randomly fails sometimes
     // Todo refers to the second loop local
     // Todo This is correct now because we properly pop the loop context
```
Masrepus, line 236: Still relevant?

src/test/java/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontendTest.java
```diff
@@ -359,19 +359,26 @@ void testCast() throws Exception {
     assertNotNull(stmt);
 
     VariableDeclaration e = stmt.getSingleDeclarationAs(VariableDeclaration.class);
-    assertEquals(TypeParser.createFrom("ExtendedClass", true), e.getType());
+    // This test can be simplified once we solved the Issue with
```
Masrepus, line 362: What issue do you mean?

Masrepus (commented):
I like the new use of a linked hash set for child order. Overall, the PR looks good to me, only the thing about distinguishing field declaration and definition needs some more discussion

gradlew.bat
konradweiss: No definitely no intentional change

src/test/java/de/fraunhofer/aisec/cpg/enhancements/variable_resolution/VariableResolverCppTest.java
konradweiss: Yes, with regards to the above comments, there can be a definition of a field and a declaration outside of the class, while in most cases I guess that declarations without initializer can have a default value, there are these static field declarations where there is an actual declaration that we need to know about

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java
konradweiss: Remove them, these were unintentionally not removed.

src/test/java/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontendTest.java
konradweiss: We still need to have a discussion on how to represent names, do we always use FQNs or can we store the SimpleName only? Currently, it is inconsistent. For example, this case of type apparently Used the simple name previously.

oxisto:
Do you remember when you branched this PR originally? It seems like that the `initializer` of those two static members are not being parsed. Not sure why or which change introduced this.

oxisto:
Might I suggest that we merge #232 first?

konradweiss:
I merged master back in, they fail because of some changes introduced there. Does #232 solve the issue?

oxisto:
> I merged master back in, they fail because of some changes introduced there. Does #232 solve the issue?

Not necessarily, but it will make a lot of things, such as function declarations, variable declarations and adding them to a TU more consistent. So it might be easier to spot the problem that we currently have on master.

src/test/java/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontendTest.java
oxisto: I would actually propose to introduce a `Name` class, similar to the `Type` class, which can be used to actually compare types and internally has a representation of what that name is, i.e. its FQN and so on.

oxisto:
> > I merged master back in, they fail because of some changes introduced there. Does #232 solve the issue?
> 
> Not necessarily, but it will make a lot of things, such as function declarations, variable declarations and adding them to a TU more consistent. So it might be easier to spot the problem that we currently have on master.

After testing a merge-in of #232 into this branch, the tests above are fixed but FunctionPointerTest#testC fails now ^^

oxisto:
BTW this is not valid *C* (you are re-declaring a function with the same name). But not sure if this is a problem for us. In C++ this is possible, so our code should be able to handle it.

https://github.com/Fraunhofer-AISEC/cpg/blob/0e21d17cb9c42286b0889b4756ae9b4420a06992/src/test/resources/functionPointers/func_ptr.c#L1-L2

oxisto:
The problem seems to be, that somehow instead of using the dedicated `resolveFunctionPtr` function in the `VariableUsageResolver` the regular `resolve` of the `ScopeManager` is used, and this one only checks for the name of the variable (which goes horribly wrong if you refer to functions instead of variables): 

https://github.com/Fraunhofer-AISEC/cpg/blob/0e21d17cb9c42286b0889b4756ae9b4420a06992/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/ScopeManager.java#L537

src/main/java/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.java
```diff
@@ -208,10 +203,8 @@ private void resolveLocalVarUsage(RecordDeclaration currentClass, Node parent, N
         return;
       }
 
-      var refersTo =
-          walker.getDeclarationForScope(
-              parent,
-              v -> !(v instanceof FunctionDeclaration) && v.getName().equals(ref.getName()));
+      Optional<? extends ValueDeclaration> refersTo =
+          Optional.ofNullable(lang.getScopeManager().resolve(ref));
```
oxisto, line 207: This is a problem, since this will invoke the regular resolve of the ScopeManager instead of the resolveFunctionPtr a few lines below. This is a problem for functions that have the same name but different parameters.

oxisto (changes requested):

src/main/java/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.java
oxisto: A possible solution would be to integrate the `resolveFunctionPtr` functionality into the `resolve` method of the ScopeManager.

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java
```diff
@@ -112,22 +121,60 @@ private FieldDeclaration handleFieldDeclarator(CPPASTDeclarator ctx) {
...
+              this.lang.getLocationFromRawNode(ctx),
+              initializer,
+              true);
+
+      var recordDeclaration =
+          this.lang
+              .getScopeManager()
+              .getRecordForName(this.lang.getScopeManager().getCurrentScope(), recordName);
+
+      // prepared for PR #223 - to set the definition here
```
oxisto, line 150: @konradweiss This would be the place to add the definition magic for external static fields, if it is still intended.

konradweiss: Currently the associated tests don't fail so I wonder why. I will take a look into it.

konradweiss: @oxisto I finally got around to integrate the function pointer resolution of fptrs that do not reference a class over the scope manager by integrating a part of the function pointer resolution. If this is the change you accept we can finally merge.

gradlew.bat
oxisto: Can you revert this change in this file? Then let’s merge it

oxisto (approved):

PR #228, opened 2020-09-01
Restructure node package structure

### LINKED ISSUE #84: Refactor package structure for nodes
We shouldn't have one big `graph` package containing both nodes and any related helper classes. I would propose having at least a `node` subpackage, probably even `node.declaration`, `node.expression` and `node.statement`, so that the broader inheritance structure is reflected. This would make things a lot more organized, while still providing close to zero refactoring effort. Of course this is a breaking change for library users, but nothing semantic, so it can just be resolved by fixing any imports.

Surely this is not of top priority, but I still would like to have this inside an issue for further discussion.

Masrepus:
Closes #84

# API Changes

Moving Declarations to `.graph.declarations`, statements to `.graph.statements` and expressions to `.graph.statement.expressions` for easier overview of the whole CPG structure


oxisto:
Just a minor suggestion: Can we use the plural terms? This would be more fitting with the rest of the package structures as in `passes`, `helpers`, or `frontends`.

Masrepus:
Sure thing. While at it, I also changed `.graph.type` to plural as well

oxisto:
Looks good to me then. Probably needs a `2.4.0` or `3.0.0` release then :)

PR #231, opened 2020-09-02
Declaration Chains and Redeclarations

oxisto:
This is a first jab at implementing re-declarations. It is not quite finished yet, but rather provides the basic framework for it. It later needs to be tied in with the `ScopeManager` to deal with the actual (re)-definitions.


oxisto:
I will mark this as blocked until we reach a decision about #280

PR #232, opened 2020-09-02
Adding declarations to translation units through ScopeManager

oxisto:
Declarations in the translation units where the only ones not added by the scope manager. This PR fixes that, leaving the scope manager the singular place to add new declarations. This is necessary to implement things like re-declarables (see #231).

It also contains some cleaner handling of fields, functions, methods and constructors, i.e. checking based on the current scope if it is a function or method. This should remove the need to create function declarations, which are then later on "upgraded" to methods or constructors, effectively removing the need for `ScopeManager::replaceNode` (and #226)

# API Changes

I have added a new interface `DeclarationHolder`, which CPG nodes can implement, which denotes that this node holds further declarations. It basically consists of one method, `addDeclaration`, which the holder needs to implement and this controls how and to which field (i.e. methods, constructors, etc. in the case of a record decl) the declaration is added to.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/graph/Statement.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/graph/Statement.java
@@ -29,15 +29,17 @@ package de.fraunhofer.aisec.cpg.graph;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import org.checkerframework.checker.nullness.qual.NonNull;
 
 /** A statement. */
-public class Statement extends Node {
+public class Statement extends Node implements DeclarationHolder {
 
   /**
    * A list of local variables associated to this statement, defined by their {@link
    * VariableDeclaration} extracted from Block because for, while, if, switch can declare locals in
    * their condition or initializers
    */
+  // TODO: This is actually an AST node just for a subset of nodes, i.e. initializers in for-loops
   @SubGraph("AST")
   protected List<VariableDeclaration> locals = new ArrayList<>();
 
@@ -64,5 +66,12 @@ public class Statement extends Node {
   @Override
   public int hashCode() {
     return super.hashCode();
+  }
+
+  @Override
+  public void addDeclaration(@NonNull Declaration declaration) {
+    if (declaration instanceof VariableDeclaration) {
+      this.locals.add((VariableDeclaration) declaration);
+    }
   }
 }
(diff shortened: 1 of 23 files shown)


oxisto:
Ready for review

oxisto:
Furthermore, I have added a new interface `DeclarationHolder`, which CPG nodes can implement, which denotes that this node holds further declarations. It basically consists of one method, `addDeclaration`, which the holder needs to implement and this controls how and to which field (i.e. methods, constructors, etc. in the case of a record decl) the declaration is added to.

An alternative approach would be that the declaration holder would only return the appropriate collection and the scope manager would do the actual *adding* (see https://github.com/Fraunhofer-AISEC/cpg/pull/232/commits/677cf626a7281db700160560f0cc28930e47a31b). Not quite sure, which one is the better approach.

src/main/java/de/fraunhofer/aisec/cpg/graph/Statement.java
```diff
@@ -38,7 +38,8 @@
    * VariableDeclaration} extracted from Block because for, while, if, switch can declare locals in
    * their condition or initializers
    */
-  private @SubGraph("AST") List<VariableDeclaration> locals = new ArrayList<>();
+  @SubGraph("AST")
```
Masrepus, line 41: That's what I removed in my new PR, maybe that's something that can already be removed in this one

src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/RecordScope.java
```diff
@@ -30,6 +30,8 @@
 
 public class RecordScope extends NameScope {
 
+  enum Visiblity {}
```
Masrepus, line 33: What exactly does an enum without entries do :sweat_smile:

Masrepus (dismissed):
The concept of a `DeclarationHolder` is quite elegant in my opinion, I like it. The only thing whose purpose I don't understand is the empty `Visibility` enum :smile:

src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/RecordScope.java
oxisto: Leftover... I wanted to start implementing visibility of class members but decided against it. I can remove it.

src/main/java/de/fraunhofer/aisec/cpg/graph/Statement.java
oxisto: can do

oxisto: See discussion on Mattermost, I have decided against removing it, since we are not quite clear if it is actually an AST node if just a `Statement` is used as part of the condition or initialiser of a `for` or `if`.

oxisto:
Any objections to merging this, @konradweiss ?

Masrepus (approved):

PR #236, opened 2020-09-08
Consistent equals/hashCode strategy

### LINKED ISSUE #177: Refactor equals/hashCode strategy
As mentioned in #164 we should think about a sensible way of creating hash codes for nodes. Does it really make sense to only implement this in the base class or do we actually want to create specific implementations for each node type?

Masrepus:
Closes #177 
Equals checks now consistently only use AST children of nodes, to avoid any kind of circular checking. Hashing of a node is still done solely in `Node`. The reason for this is that many fields of a node may change throughout its lifecycle. And if these values are included in the hash calculation, hash sets etc may not recognize the node they contain anymore, due to its changed hash. As such, only a node's name, code and java class name are used to calculate the hash, as these are things that don't change, but are more or less enough to quickly distinguish nodes. Upon a hash match the full `equals` method is called anyways


Masrepus:
Now I know why we had this weird pattern of overriding hashCode in every node but only calling super.hashCode. It's because sonar does not like this. What should we do about that? I'm not a fan of overriding a method just to call the actual super implementation, but as I explained above, overriding hashCode in every node does not make sense for us...

Masrepus:
Merging is blocked until #228 is merged

oxisto:
Any idea how to satisfy SonarQube? :)

Masrepus:
Well, originally we satisfied it by using the following code in all nodes:
```java
@Override
public int hashCode() {
    return super.hashCode();
}
```
Which in my opinion is even worse :sweat_smile:

oxisto:
Closing this for now, we need a fresh approach for this it seems

PR #242, opened 2020-09-14
Fix issue when parsing functions whose name is in extra parentheses

Masrepus:
As found by @peckto when trying to parse the eLua library

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java
@@ -166,15 +166,19 @@ class DeclaratorHandler extends Handler<Declaration, IASTNameOwner, CXXLanguageF
   }
 
   private ValueDeclaration handleFunctionDeclarator(CPPASTFunctionDeclarator ctx) {
-    // Attention! This might actually be a function pointer
-    if (ctx.getNestedDeclarator() != null
-        && ctx.getNestedDeclarator().getPointerOperators().length > 0) {
-      return handleFunctionPointer(ctx);
+    // Programmers can wrap the function name in as many levels of parentheses as they like. CDT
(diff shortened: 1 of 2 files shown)


oxisto:
Hm I was under the impression that I am handling nested declarators somewhere else already. Could you maybe check where that is, I only have access to the web Ui currently. It should be handled consistently.

oxisto:
See https://github.com/Fraunhofer-AISEC/cpg/blob/0b6dda87191cbe8066d0b60b41ed957387cc1d22/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.java#L89

I would prefer something similar to this versus looping in a handle function

Masrepus:
The problem here is that we only need the nested declarator for the name, the rest of the handling needs to be done with the top level declarator. And as far as I can see it, the loop will always terminate after 1-2 iterations, as deeper nestings would mean that we are parsing really ugly code, which won't be the normal case

oxisto:
Ah I see, then I think it is ok for now. We might need a handler to actually properly parse names (also including scopes, etc.) to something like a `Name` node.

oxisto:
It seems, that further modifications are necessary, I changed the `TypeTest::testFunctionPointerTypes` to include such a scenario, i.e. changing the `fptr_type.cpp` to `int (*(global_two_param))(int, unsigned long);`, this construct seems to be parsed as a function declaration instead of a function pointer (var decl).

oxisto:
Ah the issue seems to be detecting both  
`int ((*global_two_param))(int, unsigned long);` and
`int (*(global_two_param))(int, unsigned long);`

Although I am not 100 % sure if the first one is actually valid

Masrepus:
Hm we could check if there is only a single pointer op in all loop iterations. I guess this would be the best to do

oxisto:
I'll merge this for now, there is still a slight issue in the type parser that does not handle these parentheses very well.

PR #249, opened 2020-09-29
Java annotations

oxisto:
Basic implementation of Java annotations (not the declaration itself but the usage)

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontend.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontend.java
@@ -419,6 +419,7 @@ public class CXXLanguageFrontend extends LanguageFrontend {
     List<AnnotationMember> list = new ArrayList<>();
 
     for (IASTToken token : tokenList.getTokens()) {
+      // skip commas and such
       if (token.getTokenType() == 6) {
         continue;
       }
@@ -435,12 +436,15 @@ public class CXXLanguageFrontend extends LanguageFrontend {
     Expression expression;
     switch (token.getTokenType()) {
       case 1:
+        // a variable
         expression = newDeclaredReferenceExpression(code, UnknownType.getUnknownType(), code);
         break;
       case 2:
+        // an integer
         expression = newLiteral(Integer.parseInt(code), TypeParser.createFrom("int", true), code);
         break;
       case 130:
+        // a string
         expression =
             newLiteral(
                 code.length() >= 2 ? code.substring(1, code.length() - 1) : "",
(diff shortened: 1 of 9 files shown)


oxisto:
Any objections on merging this?

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontend.java
```diff
@@ -404,8 +415,8 @@ public void processAttributes(@NonNull Node node, @NonNull IASTAttributeOwner ow
     return list;
   }
 
-  private List<Expression> handleTokenList(IASTTokenList tokenList) {
-    List<Expression> list = new ArrayList<>();
+  private List<AnnotationMember> handleTokenList(IASTTokenList tokenList) {
+    List<AnnotationMember> list = new ArrayList<>();
 
     for (IASTToken token : tokenList.getTokens()) {
       if (token.getTokenType() == 6) {
```
konradweiss, line 422: Is there a constant for the token type that can be used for comparison or could we comment what the token type 6 means?

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontend.java
```diff
@@ -418,22 +429,26 @@ public void processAttributes(@NonNull Node node, @NonNull IASTAttributeOwner ow
     return list;
   }
 
-  private Expression handleToken(IASTToken token) {
+  private AnnotationMember handleToken(IASTToken token) {
     String code = new String(token.getTokenCharImage());
 
+    Expression expression;
     switch (token.getTokenType()) {
       case 1:
```
konradweiss, line 437: Same here for the token type, constants would be better for readability or else we need comments

oxisto: No, there isn't.

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontend.java
oxisto: No, there isn't.

oxisto: I have added comments

src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontend.java
oxisto: I have added comments

konradweiss:
The rest looks fine to me, this only handles usage of Annotations right? Creation of Annotations through @interface and similar is not supported? We can merge this and create an Issue that handles that as well.

PR #267, opened 2020-11-19
Trying to automate README.md update

oxisto:
Just trying this, it probably would make more sense to extract this into a separate GitHub action.


oxisto:
Will close this for now as it's better suited to be extracted into a GitHub workflow

PR #272, opened 2020-11-24
Improve property edges

vfsrfs:
Hotfix: @Relationship Annotations values are case sensitive.

`PropertyEdge` is now a generic class with its `T` type denoting the *end* node. In the future we might extend it to both start and end nodes.

Refactored copy/paste code into `PropertyEdge::unwrap` and most of the creation of property edges is now done in `DeclarationHolder::addIfNotContains`.


src/main/java/de/fraunhofer/aisec/cpg/graph/edge/PropertyEdge.java
```diff
@@ -19,7 +19,7 @@
   /** Required field for object graph mapping. It contains the node id. */
   @Id @GeneratedValue private Long id;
 
-  protected static final Logger log = LoggerFactory.getLogger(PropertyEdge.class);
+  @Transient protected static final Logger log = LoggerFactory.getLogger(PropertyEdge.class);
```
oxisto, line 22: Wouldn't it be better to just skip static fields in the Codyze - similarly to what we already do for Nodes? It is very error-prone to have every static field annotated with `@Transient`

src/main/java/de/fraunhofer/aisec/cpg/graph/edge/PropertyEdgeConverter.java
```diff
@@ -51,7 +51,7 @@
       if (deserializer.containsKey(entry.getKey())) {
         Object deserializedProperty = deserializer.get(entry.getKey()).apply(propertyValue);
         result.put(Properties.valueOf(entry.getKey()), deserializedProperty);
-      } else {
+      } else if (!entry.getKey().equals("sub-graph")) {
```
oxisto, line 54: Not sure what this is. But at least a comment explaining the use of this constant would be good.

vfsrfs: I agree with you. Currently the subgraph annotation is realized as an annotation and not with a property field so we cannot set it. This was just preliminary so that the codyze works for now.

src/main/java/de/fraunhofer/aisec/cpg/graph/edge/PropertyEdge.java
vfsrfs: I agree, however @Transient should also be supported.

oxisto:
Looks good

oxisto (approved):

oxisto:
It seems you accidentally added files that were not supposed to be there.

oxisto:
I've removed (almost) all the code smells and did a lot of cleanup up of error-prone boilerplate copy/paste code. I've extracted the common functionality of returning target nodes into `PropertyEdge::unwrap` and most of the creation of property edges is now done in `DeclarationHolder::addIfNotContains`.

PR #276, opened 2020-11-28
openCypher integration

oxisto:
This is work in progress. This adds a `Graph` class, which offers a simple interface to query its content with openCypher (https://opencypher.org). It is tied to the translation results using a Kotlin extension  `TranslationResult.graph` and can be used like this:

```Kotlin
val graph = result.graph

var nodes = graph.query("MATCH (m:MethodDeclaration) RETURN m")
```

Only a very limited subset of openCypher is supported for now. This includes

- Simple `MATCH` for nodes with a simple `WHERE` query, such as `MATCH (n) WHERE n.name = "myName" RETURN n`
- Very basic relationship queries, such as `MATCH (n:SomeType)-[:SOME_EDGE]->(m:OtherType) RETURN n`, also including simple `WHERE` constrains on the nodes, but not on the edge
- Can now either return the left or ride side variable of a relationship query, i.e. `MATCH (n:FunctionDeclaration)-[:PARAMETERS]->(m:ParamVariableDeclaration) RETURN m`
- Labels on multiple hierarchy levels are supported, i.e. looking for a `VariableDeclaration` should also find `ParamVariableDeclarations`

BEWARE: This is not optimised at all. It just stupidly uses the Java Stream API. I guess there is huge potential in optimising things, but this is not in scope at the moment.

The unit test https://github.com/Fraunhofer-AISEC/cpg/blob/opencypher-support/src/test/java/de/fraunhofer/aisec/cpg/graph/QueryTest.kt should give you an indication of what is supported.

Definitely NOT working (yet) is:
- `WHERE` conditions on edge properties
- an edge `MATCH` without specifying the edge label, such as `-[]->`. This is potentially not that easy, since it means that we need to gather all edge properties in a class. Also probably pretty slow without any optimisation
- Undirected edges or edges specifying both directions
- everything else in the openCypher language

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -165,6 +165,9 @@ dependencies {
     // CDT
     api("org.eclipse.cdt:core:6.11.1.202006011430")
 
+    // openCypher
+    api("org.opencypher:front-end-9.0:9.0.20190305")
+
     implementation(platform("org.jetbrains.kotlin:kotlin-bom"))
     implementation("org.jetbrains.kotlin:kotlin-stdlib")
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -166,7 +166,7 @@ dependencies {
     api("org.eclipse.cdt:core:6.11.1.202006011430")
 
     // openCypher
-    api("org.opencypher:front-end-9.0:9.0.20190305")
+    api("org.opencypher:parser-9.0:9.0.20190305")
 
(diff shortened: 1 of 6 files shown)


oxisto:
Marking this as blocked by #280

oxisto:
I think we could consider this as an initial, very experimental attempt, merge it and work on implementing the rest later.

konradweiss (approved):

PR #279, opened 2020-11-29
Fixed #278 by implementing single member annotations in Java

### LINKED ISSUE #278: Support implicit annotation style in Java
In Java you can either specifiy annotation values for keys directly, such as `@RequestMapping(value = "/test")` or indirectly using `@RequestMapping("hello")`. If you use the latter, then it will be assigned to an attribute named `value`, if it exists.

Currently, the CPG does just ignore this example; it should be parsed as an annotation with a entry of `value` and `"hello"`.

oxisto:
Looking also for single member annotations in Java now.

Closes #278


konradweiss:
Nice, I see no changes necessary in this PR so i will merge it.

PR #286, opened 2020-12-14
Fixes for Java static calls

### LINKED ISSUE #285: Build-pattern style member calls are recognised as static call expressions
i.e. 

```
target.path("/").get();
```

wrongly results in a `StaticCallExpression` with name `get` and fqn `target.path("/").get`. It should be a `MemberCallExpression` with further call expressions as a base.

oxisto:
This fixes #285 by properly parsing member call expressions.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -153,7 +153,7 @@ java {
 
 dependencies {
     api("org.apache.commons:commons-lang3:3.11")
-    api("org.neo4j:neo4j-ogm-core:3.1.7")
+    api("org.neo4j:neo4j-ogm-core:3.2.10")
     api("org.apache.logging.log4j:log4j-slf4j18-impl:2.14.0")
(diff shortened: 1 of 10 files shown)


konradweiss:
When fixing the last tests I found that the DeclaredReferenceExpression Added for unresolved MemberCallExpression  have no Location. Was that intentional? I see an Argument for not giving a location to nodes that are not originally coming from the parser.. On the other hand we can also view them as additional semantic information to the call and therefore give it the same location as the member call expression.

oxisto:
> When fixing the last tests I found that the DeclaredReferenceExpression Added for unresolved MemberCallExpression have no Location. Was that intentional? I see an Argument for not giving a location to nodes that are not originally coming from the parser.. On the other hand we can also view them as additional semantic information to the call and therefore give it the same location as the member call expression.

I think that is not intentional. I would add the same location as you suggested.

PR #287, opened 2020-12-15
compatibility to neo4j ogm

AndHager:
Loading a node from a Neo4j database that contains a PhysicalLocation throws a MappingException that is caused by a ClassCastException. 
The call:
session.query(Node::class.java,  "MATCH (n) RETURN n", HashMap<String, Any>())

The ClassCastException is caused by the conversion of a Long to a Integer in the LocationConverter and toIntExact expects a long value.
Therefore, I replaced the dangerous conversion with Long.parseLong and tested it.

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/helpers/LocationConverter.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/helpers/LocationConverter.java
@@ -33,19 +33,26 @@ import de.fraunhofer.aisec.cpg.sarif.Region;
 import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
+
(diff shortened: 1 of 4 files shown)


oxisto (approved):

konradweiss (approved):

oxisto:
Looks like we need to merge this manually because it was from an external repo and the restriction is that the "build" job is executed - which is not executed from an external repo, argh.

konradweiss:
> Looks like we need to merge this manually because it was from an external repo and the restriction is that the "build" job is executed - which is not executed from an external repo, argh.

I will find an "elegant" ;) solution for this problem

PR #294, opened 2021-01-08
Fix Incorrect call resolution for functions outside of a class

### LINKED ISSUE #292: Incorrect resolution of method invocation in C++
The resolution of functions invocation in C++ does not work correctly. The code snippet below calls the function `printname` in the main function. This call should be resolved to the `MethodDeclaration` `printname` from the `Geeks` `RecordDeclaration`. Instead an implicit `FunctionDeclaration`  `printname` is created which is directly connected with the TranslationUnitDeclaration.

## Snippet
```
// C++ program to demonstrate
// accessing of data members

#include <bits/stdc++.h>
using namespace std;
class Geeks
{
    // Access specifier
    public:

    // Data Members
    string geekname;

    // Member Functions()
    void printname()
    {
       cout << "Geekname is: " << geekname;
    }
};

int main() {

    // Declare an object of class geeks
    Geeks obj1;

    // accessing data member
    obj1.geekname = "Abhi";

    // accessing member function
    obj1.printname();
    return 0;
}
```

vfsrfs:
Fixes #292

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
@@ -239,6 +239,9 @@ public class CallResolver extends Pass {
           && ((HasType) member).getType() instanceof FunctionPointerType) {
         handleFunctionPointerCall(call, member);
         return;
+      } else {
+        handleMethodCall(curClass, call);
+        return;
       }
     }
 
@@ -276,7 +279,7 @@ public class CallResolver extends Pass {
   }
 
   private void handleNormalCalls(RecordDeclaration curClass, CallExpression call) {
-    if (curClass == null && this.currentTU != null && !(call instanceof MemberCallExpression)) {
+    if (curClass == null && this.currentTU != null) {
       // Handle function (not method) calls
       // C++ allows function overloading. Make sure we have at least the same number of arguments
       List<FunctionDeclaration> invocationCandidates =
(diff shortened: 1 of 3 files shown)


src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -276,7 +276,7 @@ private void resolveArguments(CallExpression call, RecordDeclaration curClass) {
   }
 
   private void handleNormalCalls(RecordDeclaration curClass, CallExpression call) {
-    if (curClass == null && this.currentTU != null) {
+    if (curClass == null && this.currentTU != null && !(call instanceof MemberCallExpression)) {
```
oxisto, line 279: I would actually prefer, if we would call `handleMethodCall` in the if-branch of the `handleCallExpression` function, something like this:

```
   if (call instanceof MemberCallExpression) {
      Node member = ((MemberCallExpression) call).getMember();
      if (member instanceof HasType
          && ((HasType) member).getType() instanceof FunctionPointerType) {
        handleFunctionPointerCall(call, member);
        return;
      } else {
        handleMethodCall(curClass, call);
        return;
      }
    }
```

I think that would be cleaner, since there is already this if, that decides how to proceed with member call expressions, instead of burying the decision in `handleNormalCalls`.

oxisto (changes requested):

oxisto (approved):

PR #306, opened 2021-01-16
Revert "Update dependency com.github.javaparser:javaparser-symbol-solver-core to v3.18.0"

### LINKED ISSUE #304: Analysis of Java files takes much longer with possible timeout
We observe a heavy time penalty (and potential timeout) with CPG v3.3.0 vs. v3.2.0 when analyzing [AESCTR.java](https://github.com/Fraunhofer-AISEC/codyze/blob/59da2a92a5d6623d0066efb4bbae8d860a8485aa/src/test/resources/java/jca/AESCTR.java). In CPG v3.3.0 the analysis takes more than **10 minutes** compared to ~ 1 second in v3.2.0.

The problem occurs somewhere in the transformation part (i.e. between https://github.com/Fraunhofer-AISEC/cpg/blob/d70539dbbb9c8939f07b79785d45548bbb159761/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontend.java#L133 and https://github.com/Fraunhofer-AISEC/cpg/blob/d70539dbbb9c8939f07b79785d45548bbb159761/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontend.java#L169).

 I'm guessing that there is a long running loop or deep recursion somewhere in the `ScopeManger` or in the `DeclarationHandler`. Sorry that I cannot be more precise for now.

oxisto:
Reverts Fraunhofer-AISEC/cpg#282

Fixes #304

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -156,7 +156,7 @@ dependencies {
     api("org.neo4j:neo4j-ogm-core:3.2.10")
     api("org.apache.logging.log4j:log4j-slf4j18-impl:2.14.0")
     api("org.slf4j:jul-to-slf4j:1.8.0-beta4")
-    api("com.github.javaparser:javaparser-symbol-solver-core:3.18.0")
+    api("com.github.javaparser:javaparser-symbol-solver-core:3.17.0")
 
     // Eclipse dependencies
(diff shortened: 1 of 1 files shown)


oxisto:
We should to a hot fix to 3.3.1 after this @konradweiss

PR #313, opened 2021-01-22
Prolongs dfg graphs to the root of a language feature 

konradweiss:
Currently we have the root node as branching node to keep the semantic information of the language-feature that branches while traversing the EOG. To keep EOG and DFG coherent and later build a propper PDG we prolong the dfg paths to the branching node.

I added the DFG edges where I added the EOG-Edges to the root node: in the EOG-pass- A strategy pattern would be nice to make the root node configurable for EOG and DFG but so far I could not find an easy way to integrate one for the EOG pass.

Root node of language feature if is a sink for the EOG of the condition and then branches out. For certain applications this rootnode allows traversals to use the root node like a jumpi 
eog_root_node

DFG edge from condition to the language-feature root node that acts as branching condition.
dfg_root_node

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
@@ -37,6 +37,7 @@ import de.fraunhofer.aisec.cpg.graph.statements.expressions.*;
 import de.fraunhofer.aisec.cpg.graph.types.Type;
 import de.fraunhofer.aisec.cpg.graph.types.TypeParser;
 import de.fraunhofer.aisec.cpg.helpers.SubgraphWalker;
+import de.fraunhofer.aisec.cpg.helpers.Util;
 import de.fraunhofer.aisec.cpg.passes.scopes.*;
 import java.util.*;
 import java.util.stream.Collectors;
@@ -781,7 +782,10 @@ public class EvaluationOrderGraphPass extends Pass {
     createEOG(doStatement.getStatement());
 
     createEOG(doStatement.getCondition());
+
+    doStatement.addPrevDFG(doStatement.getCondition());
     pushToEOG(doStatement); // To have semantic information after the condition evaluation
+
     connectCurrentToLoopStart();
     LoopScope currentLoopScope = (LoopScope) lang.getScopeManager().leaveScope(doStatement);
     if (currentLoopScope != null) {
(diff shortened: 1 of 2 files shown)


konradweiss:
I also tested this with codyze and the tests run through so it should be fine

src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java
```diff
@@ -781,7 +782,10 @@ private void handleDoStatement(@NonNull Node node) {
     createEOG(doStatement.getStatement());
 
     createEOG(doStatement.getCondition());
+
+    doStatement.addPrevDFG(doStatement.getCondition());
```
oxisto, line 786: I am semantically confused about an evaluation order pass, that adds data flow edges. Does that make sense?

vfsrfs (approved):

oxisto:
> Currently we have the root node as branching node to keep the semantic information of the language-feature that branches while traversing the EOG. To keep EOG and DFG coherent and later build a propper PDG we prolong the dfg paths to the branching node.
> 
> 
> 
> I added the DFG edges where I added the EOG-Edges to the root node: in the EOG-pass- A strategy pattern would be nice to make the root node configurable for EOG and DFG but so far I could not find an easy way to integrate one for the EOG pass.
> 
> 
> 
> Root node of language feature if is a sink for the EOG of the condition and then branches out. For certain applications this rootnode allows traversals to use the root node like a jumpi 
> 
> eog_root_node
> 
> 
> 
> DFG edge from condition to the language-feature root node that acts as branching condition.
> 
> dfg_root_node
> 
> 
> 
> 

Could you provide the code associated with the example graph? I am still not quite sure if I understand it correctly.

konradweiss:
> > Currently we have the root node as branching node to keep the semantic information of the language-feature that branches while traversing the EOG. To keep EOG and DFG coherent and later build a propper PDG we prolong the dfg paths to the branching node.
> > I added the DFG edges where I added the EOG-Edges to the root node: in the EOG-pass- A strategy pattern would be nice to make the root node configurable for EOG and DFG but so far I could not find an easy way to integrate one for the EOG pass.
> > Root node of language feature if is a sink for the EOG of the condition and then branches out. For certain applications this rootnode allows traversals to use the root node like a jumpi
> > eog_root_node
> > DFG edge from condition to the language-feature root node that acts as branching condition.
> > dfg_root_node
> 
> Could you provide the code associated with the example graph? I am still not quite sure if I understand it correctly.

`
if(num > maxSoFar){
...
}
`

PR #317, opened 2021-01-25
Introduce ParameterizedType for Graph Consistency

### LINKED ISSUE #293: Missing Generic Type Information
Types can be parameterized when generics are defined. This can be seen in the graph below, where the `Type Box<String>` contains a generic edge to the ObjectType `java.lang.String`. However, we also need to preserve this information of the generic type in the RecordDeclaration. This way we cannot only identify that a RecordDeclaration is parameterized, but also we can represent correctly the type of the `this` field and the type of the attribute `t` . 

I would suggest to introduce a new GenericType representing an abstract type, as it can be seen in the image below.

## Graph

new_generic_type

## Snippet

```
public class Box<T> {
    // T stands for "Type"
    private T t;

    public void set(T t) { this.t = t; }
    public T get() { return t; }
}
```

```
public class Main {
    public static void main(String[] args) {
        Box<String> stringBox = new Box<String>();
        stringBox.add(new String("Hello World"));
    }
}
```

vfsrfs:
This PR fixes #293 and introduces ParametrizedTypes. The main advantage is that the usage of parameters is visible in the graph for types that have not been initialized.

- We can identify if the type of a field `this` is parameterized
- The RecordDeclaration
- If we have to Different RecordDeclarations with the same generic-name (e.g. T) we don't merge them into one node, as they represent different types.

## Graph Changes
- Replaces the nodes (ObjectType) representing a generic node with ParameterizedType
- Add a new edge from a parameterized RecordDeclaration to the ParameterizedType
- Populate the generics edge from ObjectTypes that are not initialized such as the ObjectType from the `this` field

## API Changes
None

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
@@ -317,6 +317,11 @@ public class TypeManager {
       return true;
     }
 
+    // ObjectTypes can be passed as ReferenceTypes
+    if (superType instanceof ReferenceType) {
+      return isSupertypeOf(((ReferenceType) superType).getElementType(), subType);
+    }
+
     Optional<Type> commonType = getCommonType(new HashSet<>(List.of(superType, subType)));
     if (commonType.isPresent()) {
       return commonType.get().equals(superType);
--- a/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
@@ -62,6 +62,11 @@ public class TypeManager {
 
   @NonNull private Map<String, RecordDeclaration> typeToRecord = new HashMap<>();
 
+  /**
+   * Stores the relationship between parameterized RecordDeclarations (e.g. Classes using Generics)
+   * to the ParameterizedType to be able to resolve the Type of the fields, since ParameterizedTypes
+   * are unique to the RecordDeclaration and are not merged.
+   */
   @NonNull
   private Map<RecordDeclaration, List<ParameterizedType>> recordToTypeParameters = new HashMap<>();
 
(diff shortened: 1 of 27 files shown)


src/main/java/de/fraunhofer/aisec/cpg/graph/declarations/RecordDeclaration.java
```diff
@@ -64,6 +65,10 @@
   @SubGraph("AST")
   private List<PropertyEdge<RecordDeclaration>> records = new ArrayList<>();
 
+  @Relationship(value = "PARAMETERS", direction = "OUTGOING")
+  @SubGraph("AST")
+  private List<PropertyEdge<ParameterizedType>> parameters = new ArrayList<>();
```
oxisto, line 70: The name parameters is misleading. I would recommend something like typeParameters

vfsrfs: I have removed the typeParameters edges, since they will be modelled differently, due to the implementation of CPP Templates. They are now stored in the TypeManager.

src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
```diff
@@ -62,6 +62,9 @@
 
   @NonNull private Map<String, RecordDeclaration> typeToRecord = new HashMap<>();
 
+  @NonNull
```
konradweiss, line 65: I would add a descriptive comment here. For me comming from the PR it is obvious what type parameters do, but for someone that goes through the `TypeManager` it may be not.

konradweiss (changes requested):

src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
konradweiss: The rest looks fine to me, so we can merge later

oxisto:
LGTM

PR #340, opened 2021-02-18
Update README.md

oxisto:


Masrepus:
Didn't somebody insert the wildcard on purpose? I think to remember that the jar has the version number in its name

oxisto:
> Didn't somebody insert the wildcard on purpose? I think to remember that the jar has the version number in its name

The version is now only "injected" by the CI system. If you check out the master branch and build it locally, the version is not there.

Masrepus:
Ah I see!

PR #345, opened 2021-02-26
Deprecate old CFG pass

konradweiss:


oxisto:
Looks good, we can ignore the debt rating here, it should be still ok after we merge it.

Issue #43, opened 2020-01-23
Create a ThisExpression


oxisto:
Currently, `this` is handled as a dummy field of a record declaration. While this works, it is not very clean in my opinion and also leads to funny errors, such as `object.this.call()` is parsed. 

We should replace it with a `ThisExpression` that refers to the record declaration and can be used as a base for member field access and calls.

KuechA:
@oxisto has this been fixed by #835? Seems to be related

oxisto:
> @oxisto has this been fixed by #835? Seems to be related

Yes, this can be closed.

PR #350, opened 2021-03-06
Initial Go Language support

oxisto:
This is a work in progress PR that includes Go parsing in the CPG. It uses the Golang `ast` package through JNI, which might not necessarily be "fuzzy" enough for our purposes, but lets see. As far as I have seen, there are no decent ways to parse Go using either Java or Kotlin, thus the need for parsing it in Go itself. Additionally, usually it is most efficient / accurate to parse a language using tools written in the language itself.

# Features

It has some initial support for

- [x] structures
- [x] methods / function declarations
- [x] call expressions
- [x] member expressions
- [x] variable declarations
- [x] assignments / binary operators
- [x] if expressions
- [x] switch expression

# Behaviour Changes

The previous way how a call expression was resolved to a function declaration in the `CallResolver` was not aware of scopes, i.e., it was only looking at the TU level (global scope). This created a problem for Go which has *functions* outside of classes but on package scope, i.e., not directly on the global scope (TU). This is now fixed by moving the resolution into the scope manager and having it work similarly to resolving variables.

# Graph Changes

There is a small change in the `MethodDeclaration`. It now includes an additional `receiver` variable, which can be used to model receiver variables in method declarations that exist in multiple languages, such as Golang. Actually, they also exist implicitly in Java, as well, just the receiver variable is always called `this` and is modelled as a keyword rather than a variable. In the future we should also actually use this variable in the Java and C++ languages frontends instead of the dummy `this` field in the `RecordDeclaration` (as an alternative to #43).

There are some more intricate details of the Go language which we cannot easily model, i.e. multiple return values, which we do NOT yet change.


oxisto:
Just to clarify: a lot of the Go code is actually wrapping around JNI objects for convenience, in the future this code would be suited to be auto-generated out of the Java classes to also keep them in sync.

go-jni/frontend/handler.go
```diff
@@ -0,0 +1,939 @@
+package frontend
+
```
konradweiss, line 2: I would propose to name this file go_language_frontend or GoLanguageFrontend, or what the appropriate naming convention for go files is, to relate it to the frontend classes of Java and C++

go-jni/frontend/handler.go
```diff
@@ -0,0 +1,939 @@
...
+	// add it
+	scope.AddDeclaration((*cpg.Declaration)(p))
+
+	return
+}
+
+func (this *GoLanguageFrontend) handleDecl(fset *token.FileSet, decl ast.Decl) *cpg.Declaration {
+	this.LogDebug("Handling declaration (%T): %+v", decl, decl)
+
+	switch v := decl.(type) {
```
konradweiss, line 81: Looks like this PR uses explicit dispatching, in the other handlers the dispatching is done by the handler superclass and such switches are not necessary in the construction code.

go-jni/string.go
```diff
@@ -0,0 +1,39 @@
+package cpg
+
```
konradweiss, line 2: The purpose of this file seems to be to create constants of different type, can we rename this file to constants.go?

konradweiss (changes requested):
We should rename some of the files to give more information on what they do and consider to already do some refactoring that would be necessary in the future, e.g. separation of construction logic and handler dispatching to avoid switches in the code.

go-jni/frontend/handler.go
oxisto: We cannot use the handler super class since the handler system depends on dispatching handler functions based on a raw ast object that exist in the JVM, i.e., the JavaParser and Eclipse CDT objects. There are no equivalent Java objects in this case, so it is not possible, unless we create dummy objects in the JVM representing the Go objects. And that does not make sense in my opinion.

fwendland (changes requested):
Besides my comments, lgtm

go-jni/build.sh
```diff
@@ -0,0 +1,18 @@
...
+then
+    EXTENSION="dylib"
+else
+    EXTENSION="so"
+fi
+
+if [ "$JAVA_HOME" == "" ]
+then
+    JAVA_HOME=`/usr/libexec/java_home`
+fi
```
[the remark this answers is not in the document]
oxisto: I do not have access to a windows machine to check it. Feel free to add Windows support, However I would not see it as a priority until we decide to actually ship the native libraries with releases. Currently, libraries for experimental languages need to be built manually.

go-jni/frontend/handler.go
oxisto: The file contains 1000 LoC so i think for now this is ok.

go-jni/.vscode/tasks.json
```diff
@@ -0,0 +1,17 @@
...
+            "type": "shell",        
+            "problemMatcher": [],
+            "command": "./build.sh",
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            }
+        }
+    ]
+}
```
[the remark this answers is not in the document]
oxisto: I see no reason against that

konradweiss (approved):

PR #352, opened 2021-03-10
CallResolver Refinement for Calls with Implicit Cast

### LINKED ISSUE #325: Incorrect Call Resolution due to Implicit Cast
The CallResolver Pass is not able to resolve call if the function signature is modified by an implicit type conversion.

```
#include <iostream>
using namespace std;

int fixed_multiply (int val)
{
  return val * 3;
}


int main() {
  std::cout << fixed_multiply(10.0f) << '\n';
}
```
Screenshot 2021-02-04 at 14 33 19

vfsrfs:
This PR closes #325.

## Features
- The CallResolver is now able to resolve CallExpression, where an implicit cast is necessary to match the signature of the corresponding FunctionDeclaration

## Graph Changes
- An implicit CastExpression node is added when a CallExpression needs to perform a cast to be resolved

### DIFF
--- a/src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
+++ b/src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
@@ -329,6 +329,8 @@ public class CallResolver extends Pass {
           implicitCast.setExpression(call.getArguments().get(i));
           implicitCasts.add(implicitCast);
         } else {
+          // If no cast is needed we add null to be able to access the function signature list and
+          // the implicit cast list with the same index.
           implicitCasts.add(null);
         }
       }
@@ -337,7 +339,6 @@ public class CallResolver extends Pass {
     }
     return new ArrayList<>();
   }
-
 
   /**
    * modifies: call arguments by applying implicit casts
@@ -425,7 +426,8 @@ public class CallResolver extends Pass {
               .collect(Collectors.toList());
 
       if (invocationCandidates.isEmpty() && this.getLang() instanceof CXXLanguageFrontend) {
-        // If we don't find any candidate and our current language is c/c++ we check if there is a candidate with an implicit cast
+        // If we don't find any candidate and our current language is c/c++ we check if there is a
+        // candidate with an implicit cast
         invocationCandidates.addAll(resolveWithImplicitCast(call));
       }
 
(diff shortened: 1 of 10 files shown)


src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -278,6 +279,138 @@ private void resolveArguments(CallExpression call, RecordDeclaration curClass) {
...
+        Type callType = callSignature.get(i);
+        Type funcType = functionSignature.get(i);
+        if (callType.isPrimitive() && funcType.isPrimitive() && !(callType.equals(funcType))) {
+          CastExpression implicitCast = new CastExpression();
+          implicitCast.setImplicit(true);
+          implicitCast.setCastType(funcType);
+          implicitCast.setExpression(call.getArguments().get(i));
+          implicitCasts.add(implicitCast);
+        } else {
+          implicitCasts.add(null);
```
oxisto, line 331: Adding null to list can cause weird effects later. What's the point of it?

src/test/resources/calls/implicitcast/implicitcast.cpp
```diff
@@ -0,0 +1,12 @@
...
+using namespace std;
+
+int multiply (int val)
+{
+  return val * 3;
+}
+
+
+int main() {
+  std::cout << multiply(10.0) << '\n';
```
oxisto, line 11: Just to understand it correctly: We are not checking whether the call would actually succeed? I am a bit confused how we can decide which calls are ok and which ones are not. So the call would always get resolved if one primitive is used, regardless of which one?

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
vfsrfs: The idea is to have the list match the signature of the function, so it is possible to match the position of the list with the position of the argument. E.g: Assuming we have the following function and call
```
void test(int a, double b, float c) {}

int main() {
test(10.0, 10.0, 3);
}
```

we would generate the following implicit cast list: 
`(Cast(implicit=true, cast-type=int), null, Cast(implicit=true, cast-type=float))`

This means that the first argument of the CallExpression requires an implicit cast to int, the second one does not need an implicit cast and the third one needs an implicit cast to float

src/test/resources/calls/implicitcast/implicitcast.cpp
vfsrfs: If the CallResolver does not find any target to invoke before we create a dummy FunctionDeclaration, we first check if we can reach a FunctionDeclaration by modifying its signature with implicit casts.

I agree with you that we need to check if the language is C/C++, because other languages such as Java do not support this feature.

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
oxisto: Might I suggest that you add this comment (maybe in shorter form) to the code snippet above? thanks!

PR #357, opened 2021-03-15
CallResolver Refinement for CXX Call Resolution

### LINKED ISSUE #356: Incorrect Call Resolution with Default Arguments
The CallResolver Pass is not able to correctly resolve a call if the FunctionSignature contains default arguments, which are then omitted in the CallExpression.

Example:
```
#include <iostream>
using namespace std;

// defining the default arguments
void display(int = 3, char = '*');

int main() {
    // 3, * will be parameters
    display();
    return 0;
}

void display(int count, char c) {
   
}
```

default_args

vfsrfs:
This PR closes #356 

## Features
The CallResolver is now able to resolve a CallExpression which omits certain arguments, since they are predefined as defaults in the FunctionDeclaration.

Example:
```
#include <iostream>
using namespace std;

// defining the default arguments
void display(int = 3, char = '*');

int main() {
    // 3, * will be parameters
    display();
    return 0;
}

void display(int count, char c) {
   
}
```

new_graph

## Additional Features
- CallResolution for CallExpressions is refined to support default arguments.
- CallResolution for Methods and ConstructExpressions is refined to support implicit casts, default arguments and scoped resolution.
- As mentioned by @oxisto, call resolution has to traverse the scope to check for possible redefinitions (especially with a redefinition of the default values). This is now supported by the CallResolver.
- The CallResolver can now identify if a function is declared above its usage or not. If not a dummy function will be created.
- The CallResolver can correctly resolve calls where a base class contains multiple overloaded methods, but one of them is overridden in the inherited class.
- Add tests for all of the changes mentioned above.

## Graph Changes
- This PR introduces a new edge called `default` that starts at a ParamVariableDeclaration and ends at an Expression, representing the default value of the argument.
- The omitted arguments are added to the CallExpression
- The defines edge is extended to all functions, not only to methods
- A `ConstructExpression` is now derived from a `CallExpression`


vfsrfs:
Merging is blocked until #352 is merged

oxisto:
Is it intentional, that you merged the branch of #352 also into this PR?

vfsrfs:
Yes it is, because I have implemented the default arguments on top of the modifications made for implicit casts, since it is possible to encounter both features. E.g.
```
void test(int a, int b=3) {}

int main() {
    test(10.0); // apply implicit cast to int and use default value for 2nd argument
}
```

oxisto:
> Yes it is, because I have implemented the default arguments on top of the modifications made for implicit casts, since it is possible to encounter both features. E.g.
> 
> ```
> void test(int a, int b=3) {}
> 
> int main() {
>     test(10.0); // apply implicit cast from double to int and use default value for 2nd argument
> }
> ```

Makes sense. I didn't see the `blocked` label.

konradweiss (approved):

oxisto:
This conflicts with the current master.

oxisto:
One last thing ... I wonder whether is a good idea to add the default arguments to the call expression. For one, they should definitely be marked as implicit. Because we might be interested in catching call expression to a function based on a specified set of arguments and if you suddenly add arguments that were not existing in the original source code, this creates problem in querying later. With the implicit property, we could at least filter them out. But I am still not quite convinced if it makes sense to have them as part of the call expressions. Aren't they usually assigned at the start of the function (depending on the language of course) rather than added to the argument call?

oxisto:
Looking through the specification, it is still not 100 % clear how they work but it seems to lean towards just being syntactic sugar for adding the expressions to the call where they are omitted, but then I would still argue that we need to set `implicit` to true, to catch that in the analysis.

oxisto:
Ah. No, I have to retract my previous statement. This is so confusing tbh. Have a look at the example in http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4713.pdf in 11.3.6 - which I cannot copy here for ISO copyright reasons

vfsrfs:
In my opinion the default arguments are always part of the CallExpression because the expression needs to be evaluated before the function is invoked. This is the case for C++ (I don't know how it is handled in other languages). This can be observed in the compiler output:

```
|   |-CallExpr 0x7ffa05d86f50 <line:12:5, col:13> 'void'
|   | |-ImplicitCastExpr 0x7ffa05d86f38 <col:5> 'void (*)(int, char)' <FunctionToPointerDecay>
|   | | `-DeclRefExpr 0x7ffa05d86ef0 <col:5> 'void (int, char)' lvalue Function 0x7ffa05d86c78 'display' 'void (int, char)'
|   | |-CXXDefaultArgExpr 0x7ffa05d86f80 <<invalid sloc>> 'int'
|   | `-CXXDefaultArgExpr 0x7ffa05d86fa0 <<invalid sloc>> 'char'
```

Therefore, I think that we should add the default arguments to the CallExpression. 

I agree with you that it might create a problem when querying later, since it does not match the original source code. However, I don't think it would be better to add the `implicit` property to the nodes, because from the point of view of the FunctionDeclaration the default arguments exist in the source code and as a consequence they should not implicit.

What do you think of adding an `isDefault` property to the `ARGUMENTS` edge that leaves the CallExpression? This way we could filter out the default arguments when querying, but we do not need to mark the Nodes as `implicit`.

With respect to the example you mentioned, I wasn't aware, that the call resolution with defaults depends on the current scope. Thank you for pointing it out. I will mark this PR as draft until I have implemented the changes.

oxisto:
Yeah, you are right. I was checking the same thing yesterday on ASM level and it turns out that the contents of the default argument is evaluated before calling the function. 

The idea of adding `isDefault` is good, however I fear that this opens a can of worms because it means extending the `PropertyEdge` class with something like a `ArgumentEdge` class - otherwise we would add the property to all edges (which could also be ok for now and lets refactor later).

vfsrfs:
As you already mentioned I am also not sure if we can achieve the scoping resolution with 100% accuracy. However, I don't think that we will encounter redefinitions of defaults that require a resolution which consider the scope very often. Nevertheless, I will take a look at how to implement the resolution using the current ScopeManager. 

Regarding the `isDefault` property it is not necessary to create an `ArgumentEdge` and we also do not need to add the property to all edges.

The PropertyEdges are designed in such a way that you can dynamically add properties to an edge without adding the property to the other edges. The way it is implemented, is that every PropertyEdge has its own Map containing the properties. So if the key is not present, the edge has no such property.  The only thing that needs to be done is to declare the property in the `Properties` Enum, but this is because codyze requires this information beforehand.

As an example you could take a look at the `BRANCH` EdgeProperty, which was introduced only for certain `eog` edges (which handle branching contidions) and not to the other edges. The `isDefault` property could be implemented the same way.

oxisto:
> As you already mentioned I am also not sure if we can achieve the scoping resolution with 100% accuracy. However, I don't think that we will encounter redefinitions of defaults that require a resolution which consider the scope very often. Nevertheless, I will take a look at how to implement the resolution using the current ScopeManager.

I would suggest not spending too much time on it, if its not easily doable, because to be honest having a default argument which calls a function is just plain crazy in my opinion :)

> 
> Regarding the `isDefault` property it is not necessary to create an `ArgumentEdge` and we also do not need to add the property to all edges.
> 
> The PropertyEdges are designed in such a way that you can dynamically add properties to an edge without adding the property to the other edges. The way it is implemented, is that every PropertyEdge has its own Map containing the properties. So if the key is not present, the edge has no such property. The only thing that needs to be done is to declare the property in the `Properties` Enum, but this is because codyze requires this information beforehand.
> 
> As an example you could take a look at the `BRANCH` EdgeProperty, which was introduced only for certain `eog` edges (which handle branching contidions) and not to the other edges. The `isDefault` property could be implemented the same way.

Ok

oxisto:
While you are at it... Might it be possible to add a `NAME` property to edges as well? This might be a nice way to model named arguments, as for example possible in languages such as python. Although it could make more sense to add this in a separate PR.

vfsrfs:
> While you are at it... Might it be possible to add a `NAME` property to edges as well? This might be a nice way to model named arguments, as for example possible in languages such as python. Although it could make more sense to add this in a separate PR.

Yes I can do that. How should we label argument edges that are unnamed? Empty String or simply omit the property and only have it on those edges that have a name?

oxisto:
> > While you are at it... Might it be possible to add a `NAME` property to edges as well? This might be a nice way to model named arguments, as for example possible in languages such as python. Although it could make more sense to add this in a separate PR.
> 
> Yes I can do that. How should we label argument edges that are unnamed? Empty String or simply omit the property and only have it on those edges that have a name?

Yes I would omit it on those and only set the `INDEX` property. It probably gets a bit mixed up in the end because the index is currently set according to the index in the `arguments` array list. So if you have arguments 1 and 2 unnamed and two more named arguments, they will probably end up having the index 3 and 4 currently.

oxisto:
> > > While you are at it... Might it be possible to add a `NAME` property to edges as well? This might be a nice way to model named arguments, as for example possible in languages such as python. Although it could make more sense to add this in a separate PR.
> > 
> > 
> > Yes I can do that. How should we label argument edges that are unnamed? Empty String or simply omit the property and only have it on those edges that have a name?
> 
> Yes I would omit it on those and only set the `INDEX` property. It probably gets a bit mixed up in the end because the index is currently set according to the index in the `arguments` array list. So if you have arguments 1 and 2 unnamed and two more named arguments, they will probably end up having the index 3 and 4 currently.

To keep it simple I would suggest the following, @vfsrfs: I will do another branch preparing the `DEFAULT` and `INDEX` properties as well as the necessary modifications to the `CallExpression` and we will keep this purely about resolving calls with default arguments. Resolving the calls with named arguments will take another swing at it.

oxisto:
> > > > While you are at it... Might it be possible to add a `NAME` property to edges as well? This might be a nice way to model named arguments, as for example possible in languages such as python. Although it could make more sense to add this in a separate PR.
> > > 
> > > 
> > > Yes I can do that. How should we label argument edges that are unnamed? Empty String or simply omit the property and only have it on those edges that have a name?
> > 
> > 
> > Yes I would omit it on those and only set the `INDEX` property. It probably gets a bit mixed up in the end because the index is currently set according to the index in the `arguments` array list. So if you have arguments 1 and 2 unnamed and two more named arguments, they will probably end up having the index 3 and 4 currently.
> 
> To keep it simple I would suggest the following, @vfsrfs: I will do another branch preparing the `DEFAULT` and `INDEX` properties as well as the necessary modifications to the `CallExpression` and we will keep this purely about resolving calls with default arguments. Resolving the calls with named arguments will take another swing at it.

See #395

oxisto:
This is ready for review.

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -673,29 +929,77 @@ private ConstructorDeclaration createConstructorDummy(
...
     Pattern namePattern =
-        Pattern.compile("(" + Pattern.quote(record.getName()) + "\\.)?" + Pattern.quote(name));
-    return record.getMethods().stream()
-        .filter(m -> namePattern.matcher(m.getName()).matches() && m.hasSignature(signature))
-        .map(FunctionDeclaration.class::cast)
-        .collect(Collectors.toList());
+        Pattern.compile(
+            "(" + Pattern.quote(recordDeclaration.getName()) + "\\.)?" + Pattern.quote(name));
+
+    if (lang instanceof CXXLanguageFrontend) {
```
oxisto, line 938: I think this may also be relevant to the Python frontend

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -673,29 +929,77 @@ private ConstructorDeclaration createConstructorDummy(
...
       List<FunctionDeclaration> firstLevelCandidates =
           possibleTypes.stream()
-              .map(r -> getInvocationCandidatesFromRecord(r, name, signature))
+              .map(r -> getInvocationCandidatesFromRecord(r, name, call))
               .flatMap(Collection::stream)
               .collect(Collectors.toList());
-      if (firstLevelCandidates.isEmpty()) {
-        return possibleTypes.stream()
+
+      if (lang instanceof CXXLanguageFrontend) {
```
oxisto, line 994: Why is this specific for C++? Maybe a short comment would be good here

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -704,19 +1008,127 @@ private ConstructorDeclaration createConstructorDummy(
...
+        applyImplicitCastToArguments(constructExpression, implicitCasts);
+        addDefaultArgsToCall(constructorDeclaration, constructExpression);
+        return constructorDeclaration;
+      }
+    }
+    return null;
+  }
+
+  @NonNull
+  private ConstructorDeclaration getConstructorDeclarationCXX(
```
oxisto, line 1100: Why does it have CXX in the name? It seems only a part of this is C++ specific. Can we add a short javadoc saying what this function is doing?

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -704,19 +1008,127 @@ private ConstructorDeclaration createConstructorDummy(
...
+            constructExpression.addArgument(argument, true);
+          }
+          return constructor;
+        }
+      }
+    }
+    return null;
+  }
+
+  private ConstructorDeclaration resolveConstructorWithImplicitCast(
```
oxisto, line 1065: For new functions, please add a short javadoc

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -704,19 +1008,127 @@ private ConstructorDeclaration createConstructorDummy(
...
+      List<Type> signature, RecordDeclaration recordDeclaration) {
+    for (ConstructorDeclaration constructor : recordDeclaration.getConstructors()) {
+      if (constructor.hasSignature(signature)) {
+        return constructor;
+      }
+    }
+    return null;
+  }
+
+  private ConstructorDeclaration resolveConstructorWithDefaults(
```
oxisto, line 1043: For new functions, please add a short javadoc

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -339,51 +340,175 @@ private boolean compatibleSignatures(List<Type> callSignature, List<Type> functi
...
+
+    // Apply default arguments
+    for (FunctionDeclaration functionDecl : invocationTargetsWithImplicitCastAndDefaults) {
+      addDefaultArgsToCall(functionDecl, call);
+    }
+
+    return invocationTargetsWithImplicitCastAndDefaults;
+  }
+
+  private List<FunctionDeclaration> resolveWithImplicitCastFunc(CallExpression call) {
```
oxisto, line 485: Almost the same as `resolveWithImplicitCastMethod` but with a `definedBefore`?

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -412,21 +543,112 @@ private void applyImplicitCastToArguments(
...
+    for (FunctionDeclaration functionDeclaration : initialInvocationCandidates) {
+      addDefaultArgsToCall(functionDeclaration, call);
+      if (functionDeclaration.hasSignature(call.getSignature())) {
+        invocationCandidatesDefaultArgs.add(functionDeclaration);
+      }
+    }
+    return invocationCandidatesDefaultArgs;
+  }
+
+  private List<FunctionDeclaration> resolveWithDefaultArgsFunc(CallExpression call) {
```
oxisto, line 585: Why is it different for funcs and methods?

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -339,51 +340,175 @@ private boolean compatibleSignatures(List<Type> callSignature, List<Type> functi
...
+    return callSignature;
+  }
+
+  /**
+   * @param constructExpression ConstructExpression
+   * @param constructorDeclaration ConstructorDeclaration the ConstructExpression was resolved to
+   * @return list containing the signature containing all argument types including the default
+   *     arguments
+   */
+  private List<Type> getCallSignatureWithDefaults(
```
oxisto, line 370: Probably redundant, now that a construct expression is a call expression.

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
vfsrfs: In C++ the order of the functions matters, but not the order of the methods. This means a function that is declared below the usage cannot be resolved and the compiler will create an implicit function definition. In those cases we want a dummy function to be the target of the resolved. In methods the definition of a method can be below the usage and it is still correctly resolved.

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
vfsrfs: Yes that is possible. I already mentioned this to @konradweiss and we should refactor the CallResolver to handle the different languages. But that will be part of a different pr

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
oxisto: Well... the *declaration* of a method must always be known beforehand. The *definition* can be done later in the code. There is no difference in this regards whether this is a function or a method.

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
vfsrfs: Yes I will add a comment. As far as I know this only applies to C++. An example would be the following:
```
class Base
{
public:
    int calc(int i)
    {
        cout << "int: ";
        return i+6;
    }
};
class Overload : public Base
{
public:
    double calc(double d)
    {
        cout << "double: ";
        return d+9.9;
    }
};
int main()
{
    Overload overloaded;
    cout << overloaded.calc(1) << '\n';
    cout << overloaded.calc(1.1) << '\n';
    return 0;
}
```
In C++ function overloading cannot happen at different levels in the hierarchy. That is the reason why both of those calls will invoke the `double calc(double d)` function. In java the overload class would have both overloaded methods and therefore the first call would call the method with the `int` as parameter

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
vfsrfs: There is a difference. Consider the following example:
```
#include <iostream>
using namespace std;
class Base
{
public:
    int calc(int i)
    {
        cout << "int: " << doSmth(i,1) << "\n";
        return i;
    }

    int doSmth(int a, int b) {
        return a+b;
    }
};

int main()
{
    Base base;
    base.calc(1);
    return 0;
}
```
`doSmth` is defined below `calc` but this will compile fine and work as intended

src/test/java/de/fraunhofer/aisec/cpg/enhancements/calls/CallResolverTest.java
```diff
@@ -390,8 +390,10 @@ private void testDefaultArgumentsInDefinition() throws Exception {
     // Check defaults edge of ParamVariableDeclaration
     assertTrue(displayFunction.getDefaultParameters().get(0) instanceof Literal);
     assertTrue(displayFunction.getDefaultParameters().get(1) instanceof Literal);
-    assertEquals("*", ((Literal) displayFunction.getDefaultParameters().get(0)).getValue());
-    assertEquals("3", ((Literal) displayFunction.getDefaultParameters().get(1)).getValue());
+    assertEquals(
```
oxisto, line 393: You should use the correct type here instead of casting `toString`. In this case it is a character '*' and the number 3.

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
oxisto: Ah, yes. This seems to be a special case that methods of the same class can be used (in inline methods) before they are declared. Pretty sure this only applies to this very special case.

oxisto: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf

Chapter 3.4.1 specifies this in rather long and boring details. Relevant for this is Paragraph 8. which applies to a "name" used in a function body (within a class). The name must either be declared before its use or it must be a member of that class.

oxisto: The general challenge that I see here that this is getting way to specific for C++ in the end. So I would propose not going into the very peculiar specifics of C++ in this generic class. We might have to tackle this differently in the long run  if we want to be very accurate for specific languages, but this is not our intention at the moment.

What do you think, @konradweiss ?

vfsrfs: As far as I know we cannot distinguish between methods that are defined within a class or not, so since this case is possible I decided to not enforce the rule that a function/method should be declared before its use when dealing with methods.

I agree with you, that this is very specific and it shouldn't be in this generic class, especially because it makes it very hard to maintain. I already discussed this problem with @konradweiss (see #406). I would suggest to keep these changes for now, as sooner or later we will want to have a more precise call resolution and do the refactoring as soon as possible (until we know how to implement the specific CallResolvers) in a separate PR.

src/main/java/de/fraunhofer/aisec/cpg/graph/declarations/FunctionDeclaration.java
```diff
@@ -100,6 +101,12 @@ public String getSignature() {
         + Objects.requireNonNullElse(this.type, UnknownType.getUnknownType()).getTypeName();
   }
 
+  public boolean hasSameSignature(FunctionDeclaration targetFunctionDeclaration) {
+    return targetFunctionDeclaration.getName().equals(this.name)
+        && targetFunctionDeclaration.getType().equals(this.type)
```
konradweiss, line 106: The return-type is normally not part of the function signature that influences call resolution. At least not in C++ or Java. This may lead to issues if we use it for call resolution. Is the check for type necessary for some other check?

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -339,51 +340,175 @@ private boolean compatibleSignatures(List<Type> callSignature, List<Type> functi
...
+   * @param call CallExpression which does not contain all necessary arguments and uses the default
+   *     arguments
+   */
+  private void addDefaultArgsToCall(FunctionDeclaration functionDeclaration, CallExpression call) {
+    if (functionDeclaration.hasSignature(getCallSignatureWithDefaults(call, functionDeclaration))) {
+      for (Expression expression :
+          functionDeclaration
+              .getDefaultParameters()
+              .subList(
+                  call.getArguments().size(), functionDeclaration.getDefaultParameters().size())) {
```
konradweiss, line 399: Why are we adding defaults as arguments to the call-expression itself? I would strictly oppose this if it is not necessary as the arguments are AST-Nodes. if we add the defaults to the arguments we will not be able to search the CallExpression by its syntax appearing in the original code and create potential loobs in subgraphs that connect arguments and Parameters/Initializers of Parameters.

Is there a good reason that cannot be solved in another way?

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
```diff
@@ -339,51 +340,175 @@ private boolean compatibleSignatures(List<Type> callSignature, List<Type> functi
...
+      ConstructorDeclaration constructorDeclaration, ConstructExpression constructExpression) {
+    if (constructorDeclaration.hasSignature(
+        getCallSignatureWithDefaults(constructExpression, constructorDeclaration))) {
+      for (Expression expression :
+          constructorDeclaration
+              .getDefaultParameters()
+              .subList(
+                  constructExpression.getArguments().size(),
+                  constructorDeclaration.getDefaultParameters().size())) {
+        constructExpression.addArgument(expression, true);
```
konradweiss, line 423: Same here, I would not create new AST-Nodes if it can be avoided

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
oxisto: We had this discussion (somewhere - not sure where exactly now). They way it works for most language, is that default arguments are resolved before the call and then implicitly added to the call (expression). The original syntax can be retrieved by omitting those arguments that have a `DEFAULT` property set on the edge.

I am not sure, if I follow your example w.r.t. loops. Can you give a concrete example?

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
konradweiss: Yes, at some point we may have to introduce strategies or LanguageSpecific passes, the question is how to best integrate them such that we can in theory parse multiple classes and therefor have to run/apply multiple strategies/or passes.

konradweiss (changes requested):
I would prefere not to add default initializers of parameters to the list or arguments in a call expression. As it creates AST-Nodes that were not there previously and this is an issue when you search for a call by the signature that you see in the original source code. 

I would at least discuss this here to see if we can change it.

oxisto:
> I would prefere not to add default initializers of parameters to the list or arguments in a call expression. As it creates AST-Nodes that were not there previously and this is an issue when you search for a call by the signature that you see in the original source code.
> 
> I would at least discuss this here to see if we can change it.

Any other suggestions? I would suggest to "clone" the expressions rather connecting it to the original and then marking them as `implicit`-

src/main/java/de/fraunhofer/aisec/cpg/passes/CallResolver.java
konradweiss: Yes, but that only works if someone knows that his call will get some added arguments. That may lead to unexpected results if someone does not know that he/she has to ignore does.

With regards to the loop it is not something that I don't know if it may be an issue.

konradweiss:
> > I would prefere not to add default initializers of parameters to the list or arguments in a call expression. As it creates AST-Nodes that were not there previously and this is an issue when you search for a call by the signature that you see in the original source code.
> > I would at least discuss this here to see if we can change it.
> 
> Any other suggestions? I would suggest to "clone" the expressions rather connecting it to the original and then marking them as `implicit`-

Would not solve the AST-Node issue. Only, the loop Issue that is not tangible enough to be considered a problem now. If we can't find a better solution I am fine with the current way we are doing it but I would pref not adding them as AST-Nodes. Are we getting a context issue when we compute the EOG of different call sites? So my thought is that we have an EOG Path in the args evaluation of two calls:
`
1. foo(1);
2. foo(2);

3. void foo(int i, int j = 3){ .. }
`

and we get the path 1 -> 3 -> foo();in line 1->  2-> 3 -> foo(); in line 2
                                                                                  -> foo(); in line 1

so we get a loop in the intraprocedural EOG if it goes over the initializers. I have to check if we get this problem.
Then copying the expression would actually be the solution

konradweiss:
> > > I would prefere not to add default initializers of parameters to the list or arguments in a call expression. As it creates AST-Nodes that were not there previously and this is an issue when you search for a call by the signature that you see in the original source code.
> > > I would at least discuss this here to see if we can change it.
> > 
> > 
> > Any other suggestions? I would suggest to "clone" the expressions rather connecting it to the original and then marking them as `implicit`-
> 
> Would not solve the AST-Node issue. Only, the loop Issue that is not tangible enough to be considered a problem now. If we can't find a better solution I am fine with the current way we are doing it but I would pref not adding them as AST-Nodes. Are we getting a context issue when we compute the EOG of different call sites? So my thought is that we have an EOG Path in the args evaluation of two calls:
> `
> 
> 1. foo(1);
> 2. foo(2);
> 3. void foo(int i, int j = 3){ .. }
>    `
> 
> and we get the path 1 -> 3 -> foo();in line 1-> 2-> 3 -> foo(); in line 2
> -> foo(); in line 1
> 
> so we get a loop in the intraprocedural EOG if it goes over the initializers. I have to check if we get this problem.
> Then copying the expression would actually be the solution

Loops_in_EOG_through_defaults reusage in args

`
void foo(int a, int b=3) {}

int main() {
    foo(1);
    foo(2);
    foo(3);
}
`
So we get loops here. if we have to add these new ast nodes we have to make copies. @vfsrfs Considering that we want to capture the evaluation of the default argument, i think we actually have to add these subnodes unfortunately

oxisto:
Just another thought in this, @konradweiss. The way clang handles this, is it wraps the default arguments in a [CXXDefaultArgExpr](https://clang.llvm.org/doxygen/classclang_1_1CXXDefaultArgExpr.html#details) node, which then links to the default parameter and the expression. Not quite sure, if this would solve our problem though.

konradweiss (approved):

oxisto:
LGTM

PR #358, opened 2021-03-17
Initial Python Support

maximiliankaul:
This introduces basic support for Python.

# Graph changes

No breaking changes, but new types were added.

- Added a `KeyValueExpression` for key value pairs, i.e. as part of a dictionary


oxisto:
Before we continue @maximiliankaul could you co a quick check on the license of JEP?

src/main/java/de/fraunhofer/aisec/cpg/frontends/python/JepSingleton.kt
```diff
@@ -0,0 +1,13 @@
+package de.fraunhofer.aisec.cpg.frontends.python
+
+import jep.MainInterpreter
+
+// Jep's configuration must be set before the first instance is created. Later calls to setJepLibraryPath and co result in failures.
+object JepSingleton {
+    init {
+        // TODO
+        MainInterpreter.setJepLibraryPath("/home/maximilian/.virtualenvs/jep/lib/python3.9/site-packages/jep/libjep.so")
```
oxisto, line 9: look in system directories and current path?

src/main/java/de/fraunhofer/aisec/cpg/graph/declarations/FunctionDeclaration.java
```diff
@@ -247,6 +247,10 @@ public void addParameter(ParamVariableDeclaration paramVariableDeclaration) {
     this.parameters.add(propertyEdge);
   }
 
+  public void addParameterList(List<ParamVariableDeclaration> params) {
```
oxisto, line 250: This should not be done directly but rather through the scope manager's `addDeclaration`. This will automagically add it to the correct declaration "container"

python/main.py
```diff
@@ -0,0 +1,583 @@
...
+
+
+    ### CONTROL FLOW ###
+    def visit_If(self, node):
+        debug_print(ast.dump(node))
+        stmt = IfStatement()
+        self.add_loc_info(node, stmt)
+        stmt.setCondition(self.visit(node.test))
+        if isinstance(node.body, list):
+            body = CompoundStatement()
```
oxisto, line 377: this should be handled generically by a method that handles the body and returns a compounds statement. I guess this is needed in a couple of places

python/main.py
```diff
@@ -0,0 +1,583 @@
...
+            raise NotImplementedError
+        #if node.starargs is not None:
+        #    raise NotImplementedError
+        #if node.kwargs is not None:
+        #    raise NotImplementedError
+        # body
+        for b in node.body:
+            if isinstance(b, ast.FunctionDef):
+                fd = self.visit_FunctionDef(b, returnMethod=True)
+                rec.addMethod(fd)
```
oxisto, line 537: rather than adding declarations directly, the following flow should be used:

The the scope of the class / record declaration by using ScopeManager::enterScope(rec). Then afterwards within the scope of the record, use ScopeManager::addDeclaration(fd). This applies to all *declarations*. The scope manager then handles the rest automagically, by dispatching it back to the class associated with the scope, i.e. a RecordDeclaration, which then decides how to add it. This is not really documented, maybe we need a "How to write a language frontend" guide somewhere @konradweiss

python/main.py
```diff
@@ -0,0 +1,583 @@
...
+    def visit_AsyncWith(self, node):
+        debug_print(ast.dump(node))
+        raise NotImplementedError
+
+    ### MISC ###
+    def visit_Module(self, node):
+        debug_print(ast.dump(node))
+        '''The AST is wrapped in a 'Module' with 'body' list. Visit the body.'''
+        for n in node.body:
+            self.tud.addDeclaration(self.visit(n))
```
oxisto, line 568: Same as above. the global layer is a special case. In this case, enterScope should not be used, but instead ScopeManager::resetGlobal(tu). This will then tie the current global scope to the current translation unit.

BTW: `tu` is mostly used as abbreviation for a translation unit in other langues.

oxisto (changes requested):

maximiliankaul:
> Before we continue @maximiliankaul could you co a quick check on the license of JEP? It seems to refer to a libpng license, we should stick to something which is compatible with Apache 2.0

https://github.com/ninia/jep/blob/master/LICENSE Looks compatible to me. However, I can only find sources that state zlib is compatible to GPL.

python/main.py
maximiliankaul: Yes, I talked with @konradweiss and I'm using the ScopeManager now.

oxisto:
Note (to myself): This (currently) needs Python 3.9

oxisto:
This PR is already getting quite long. I would suggest doing a merge-in soon and then continue on incrementally from there. Do we have any real blockers on this, @konradweiss, @maximiliankaul?

oxisto:
> > Before we continue @maximiliankaul could you co a quick check on the license of JEP? It seems to refer to a libpng license, we should stick to something which is compatible with Apache 2.0
> 
> https://github.com/ninia/jep/blob/master/LICENSE Looks compatible to me. However, I can only find sources that state zlib is compatible to GPL.

Yes, looks ok, also confirmed by @schanzen

src/main/python/main.py
```diff
@@ -0,0 +1,1123 @@
...
+from de.fraunhofer.aisec.cpg.sarif import PhysicalLocation
+from de.fraunhofer.aisec.cpg.sarif import Region
+from java.net import URI
+from java.util import ArrayList
+
+
+#############################
+# PROBLEMS / OPEN QUESTIONS #
+#############################
+# 1. Wie gehen wir mit Expression außerhalb von Funktionen um? Für Funktionen gibt es die FunctionDeclaration. Was gibt es für Expressions?
```
oxisto, line 43: Cleanup open issue list, i.e. make it a bit more presentable

src/main/python/main.py
```diff
@@ -0,0 +1,1123 @@
...
+                res.append(self.lines[lineno][:])
+                lineno += 1
+
+            # last line is partially read
+            res.append(self.lines[end_lineno][:end_col_offset])
+
+            return "\n".join(res)
+
+
+class MyWalker(ast.NodeVisitor):
```
oxisto, line 94: Rename MyWalker to something more meaningful

oxisto: In the long run, we might want to split it into separate files and walkers. Let's leave that for a future PR.

src/main/python/main.py
```diff
@@ -0,0 +1,1123 @@
...
+        debug_print(ast.dump(node))
+        # TODO: implement this - how?
+        # raise NotImplementedError
+        return Expression()
+
+    def visit_Attribute(self, node):
+        debug_print(ast.dump(node))
+
+        mem = MemberExpression()
+        # self.add_loc_info(node, mem)
```
oxisto, line 541: why is this commented out?

src/main/python/main.py
```diff
@@ -0,0 +1,1123 @@
...
+    def visit_Assign(self, node: ast.Assign):
+        debug_print(ast.dump(node))
+
+        if len(node.targets) != 1:
+            raise NotImplementedError
+
+        target = node.targets[0]
+
+        # if isinstance(target, ast.Attribute):
+        #    debug_print("HERE")
```
oxisto, line 616: remove

src/main/python/main.py
```diff
@@ -0,0 +1,1123 @@
...
+
+            stmt.setStatement(body)
+        return stmt
+
+    def visit_While(self, node):
+        debug_print(ast.dump(node))
+        w = WhileStatement()
+        self.add_loc_info(node, w)
+        w.setCondition(self.visit(node.test))
+        body = CompoundStatement()
```
oxisto, line 792: use `make_compound_statement`

src/main/python/main.py
```diff
@@ -0,0 +1,1123 @@
...
+        fd.addAnnotations(annotations)
+
+        # debug_print(node.de)
+
+        # handle args
+        # scopeManager adds them to fd
+        self.visit_arguments(node.args, fd)
+
+        # handle body
+        body = CompoundStatement()
```
oxisto, line 904: Use `make_compound_statement`

src/main/python/main.py
```diff
@@ -0,0 +1,1123 @@
...
+        return Expression()
+
+    ### VARIABLES ###
+    def visit_Name(self, node):
+        debug_print(ast.dump(node))
+        ref = DeclaredReferenceExpression()
+        self.add_loc_info(node, ref)
+        ref.setName(node.id)
+
+        # resolved_ref = self.scopemanager.resolve(ref)
```
oxisto, line 197: It *might* be necessary to call `ScopeManager::connectToLocal` with the `ref`. We do this for Java and Golang, but not for C++ for some reason. This is unfortunately very inconsistent, but @konradweiss is aware of this.

src/main/python/main.py
```diff
@@ -0,0 +1,1123 @@
...
+        return self.visit(node.value)
+
+    def visit_UnaryOp(self, node):
+        debug_print(ast.dump(node))
+        unop = UnaryOperator()
+        self.add_loc_info(node, unop)
+        unop.setOperatorCode(node.op)
+        unop.setInput(self.visit(node.operand))
+        if isinstance(node.op, ast.UAdd):
+            unop.setName("UAdd")
```
oxisto, line 233: `operatorCode` and `name` should be set to the same value. Additionally, it should be set to the symbols, e.g. `+` instead of a name. There is some logic code that depends on that.

src/main/python/main.py
```diff
@@ -0,0 +1,1123 @@
...
+            binop.setOperatorCode("//")
+            binop.setName("//")
+        elif isinstance(node.op, ast.Mod):
+            binop.setOperatorCode("%")
+            binop.setName("%")
+        elif isinstance(node.op, ast.Pow):
+            binop.setOperatorCode("Pow")
+            binop.setName("Pow")
+        elif isinstance(node.op, ast.LShift):
+            binop.setOperatorCode("LShift")
```
oxisto, line 286: Convert the rest of the names to symbols

oxisto (changes requested):

oxisto:
As discussed internally, this will be the last item on the 3.5 release schedule.

src/main/python/main.py
maximiliankaul: Resolved in 5ec26c85f37b9d3fda77d70d502110e9e53196a4

src/main/python/main.py
maximiliankaul: Resolved in 5ec26c85f37b9d3fda77d70d502110e9e53196a4

src/main/python/main.py
maximiliankaul: Resolved in 5ec26c85f37b9d3fda77d70d502110e9e53196a4

oxisto (approved):
I think we can go ahead and merge this initial support branch!

PR #368, opened 2021-03-22
Template Support

oxisto:
Just to keep track of it.


src/main/java/de/fraunhofer/aisec/cpg/graph/statements/expressions/TemplateCallExpression.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.*;
+import org.neo4j.ogm.annotation.Relationship;
+
+public class TemplateCallExpression extends CallExpression implements HasType.SecondaryTypeEdge {
+  @Relationship(value = "TEMPLATE_PARAMETERS", direction = "OUTGOING")
+  @SubGraph("AST")
+  private List<PropertyEdge<Node>> templateParameters = new ArrayList<>();
+
+  @Relationship(value = "INSTANTIATION", direction = "OUTGOING")
+  private FunctionTemplateDeclaration instantiation;
```
konradweiss, line 45: This field needs some documentation, it does not come naturally to me what it will point to.

vfsrfs:
This PR adds template support as discussed in #363


## Features
- Supports Type Parameters and Value Parameters as Parameters for the TemplateDeclaration
- Supports the resolution with default parameters
- Supports use of a previously defined template parameter as a default value
- Creates dummy elements when needed
- CallExpression is extended to include an `instantiation` Edge to the template that is being instantiated by the call and a `template parameter` edge that connects the CallExpression to the parameters the template is instantiated with
- Introduces TypeExpression to model the type keywords provided when instantiating a template

### ClassTemplate
- Supports templates applied to classes and structs
- Templates can be instantiated either on the stack or on the heap using the `new` keyword

Example:

classtemplate


```
template<class Type1, class Type2>
class Pair
{
    public:
        Type1 first;
        Type2 second;
};

int main()
{
    Pair<int,int> point1;
}
```


## FunctionTemplate
- Can resolve template parameters using auto deduction or implicit casts
- Prioritizes matching function that already exist instead of instantiating a template
- Supports function templates within a record

Example:

functiontemplate


(ObjectType `int` removed for better visualization)

```
template <class T=int, int N=5>
T fixed_multiply (T val)
{
  auto x = val * N;
  return x;
}

int main() {
  fixed_multiply<int,2>(10);
}
```

### Future Changes
There are some request that weren't addressed in this PR, but we will have to take care of them in the future:
- Add Support for Template Specialization
- Add TemplateExpansionPass to realize every template instantiation

Issue #215, opened 2020-08-27
Create Namespace node that references all NamespaceDeclarations which add declarations to its Scope


konradweiss:
Currently, we add a Namespace Declaration whenever we see one, however, this means that in java we create a NamespaceDeclaration for every file in a package, and the declarations in the package don't know each other, neither do the NamespaceDeclarations. In C++ a Namespace can be declared multiple times and here we have the same problem, both NamespaceDeclarations add Declarations to the same Namespace, but there is no way of querying the graph for this relation. The query could be written as all Declarations in a NamespaceDeclaration of a specific Name, but no concrete Entity of a Namespace is present.

oxisto:
The added problem is, that current each `NamespaceDeclaration` has its own `NameSpace`, making resolution of functions that are in different files but the same namespace impossible. 

Thus, in #378 I have created a workaround that very naivly looks up exist namespaces in the current scope and uses their `NameScope`, only creating a new name scope if necessary.

PR #378, opened 2021-03-26
Improvements in the Golang parser

oxisto:
- Correctly setting base of member call expression 
- Trying to resolve functions in the same namespace across files (WORKAROUND until we tackle #215)


oxisto:
Can we merge this? @konradweiss

PR #382, opened 2021-04-01
Exclude experimental languages from default build

### LINKED ISSUE #381: Build flag for conditional compilation of experimental features
The new support for Go using JNI requires the whole build chain for Go. I couldn't find a simple way to disable the experimental Go intergration during compilation. I was able to remove the compile task but there were still test cases that then failed during build/test. This is somewhat annoying if I just want to extend the Java part.

Would it be possible to add build flags to conditionally disable certain experimental features?

oxisto:
Fixes #381


fwendland (approved):
LGTM. Thank you for your work.

PR #389, opened 2021-04-09
Make language frontends configurable

oxisto:
Adding a new `registerLanguage` and `defaultLanguages` features to the translation config builder instead of a fixed factory.

This changes the API and definitely deserves a bump in the version number, maybe even a 4.0.0.


konradweiss (approved):
As far as I can see this looks mergeable :)

oxisto:
Ok, how should we proceed? Merge this and do a 4.0.0 release next? This completely breaks the existing API.

oxisto:
As discussed internally, we will wait for the 3.5 release before merging this. Setting this to blocked now.

PR #396, opened 2021-04-16
Added a small python fix

oxisto:
* Added support for conditional expressions
* Do not add null statements to compound statement, resulting in neo4j errors


maximiliankaul (approved):
Added location and a test. Looks good to me.

PR #411, opened 2021-04-27
Trying to fix neo4j by converting even more classes to Kotlin

### LINKED ISSUE #410: Cannot persist to Neo4J anymore
It seems that #395 introduced a bug because of the conversion to a Kotlin class, so that the CPG could not persist to neo4j anymore (not sure if this also affected Codyze). Basically persisting into graph ended up in an endless recursion.

oxisto:
Well.. this commit goes further down the rabbit hole in converting stuff to Kotlin. It seems that neo4j cannot properly handle entity classes that are written in Kotlin but derive from Java classes. So... this commit converts `Node` and `PropertyEdge` classes to Kotlin as well.

Not sure if this is really the best way... However, reverting #395 is also not ideal since the improved `CallExpression` class is really useful for #357. 

On the plus side, having `Node` and `PropertyEdge` in Kotlin would allow for some additional magic.

Fixes #410


vfsrfs:
What are the main benefits with regards to #357?

As far as the `is_default` property edge is concerned, we will probably not need it anymore, since we will have deal with the default arguments differently in order to fix the issue with the loops @konradweiss mentioned.

I would suggest to discuss this on Thursday, since those are major changes and may entail other consequences.

oxisto:
Well in the end it comes down to personal preferences I would say. I wouldn’t mind having the ability to extend nodes with Kotlin classes and it looks like the only way to do that is to have the base class written on kotlin. Not quite sure why. The intermediate classes seem to work fine if they are Java. 

I would prefer Kotlin over Java for most, if not all parts of the project, but that is as I said personal preference and I will not start a Kotlin vs. Java argument here.

konradweiss (approved):

PR #420, opened 2021-05-03
Converting call expressions back to Java

### LINKED ISSUE #410: Cannot persist to Neo4J anymore

oxisto:
Hopefully fixes #410 (again)


vfsrfs (approved):
Neo4j works now 👍

PR #431, opened 2021-05-20
Added simple inference for record declarations

oxisto:
Still a lot of TODOs:
- decide whether it is a class or a struct
- maybe make it optional to not break things
- where to place the additional declaration in the AST?

### DIFF
--- a/cpg-library/src/main/java/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.java
+++ b/cpg-library/src/main/java/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.java
@@ -25,6 +25,7 @@
  */
 package de.fraunhofer.aisec.cpg.passes;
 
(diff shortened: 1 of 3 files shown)


peckto:
How should we tag the inferred declarations? 
Is this a `dummy` or `implicit` or something else?
If I see it correctly, the nodes are currently tagged as `implicit`.

oxisto:
> How should we tag the inferred declarations?
> Is this a `dummy` or `implicit` or something else?
> If I see it correctly, the nodes are currently tagged as `implicit`.

There was a lot of inconsistencies between those two and this PR might be a chance to rectify some mistakes here. An `implicit` node should be something which is *definitely* there, but not part of the AST, for example an implicit `this`. The `dummy` property originally also included these implicit definitions, but also the inferences. I would propose removing the `dummy` one (it is already deprecated) and introduce an `inferred` property for those.

PR #434, opened 2021-05-21
Enable experimental languages in `cpg-neo4j`

oxisto:
This enables the usage of experimental languages, such as Python in `cpg-neo4j`, through an extra flag `--enable-experimental-python` and `--enable-experimental-go`. This is considered a workaround until we have a stable release strategy for those languages.

I have also included a short introduction, how to make it work. Follow the instructions on the main README to install jep either globally or using a virtualenv. After building the distribution, use the following steps to analyse a python file.

```bash
cd ./build/install/cpg-neo4j
bin/cpg-neo4j --enable-experimental-python ../../../src/test/resources/test.py
```
It is important to change into the directory where the distribution is installed, otherwise the `cpg.py` will not be found.

Closes #434


maximiliankaul (approved):
I'm not entirely happy with the "possibleLocations" approach for the cpg.py file but this will be resolved anyways once we ship a python module.
Otherwise: LGTM

PR #435, opened 2021-06-01
CPG Console

oxisto:
Added a simple console based on `ki`, the kotlin interpreter.


cpg-console/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+```bash
+../gradlew installDist
+build/install/cpg-console/bin/cpg-console
+```
+
+The following example snippet can be used:
+
+```kotlin
+:a src/test/resources/array.cpp
```
konradweiss, line 11: the :a command should be replaced with  :tr I think.

PR #456, opened 2021-07-03
Fixed EOG handling in function declaration statements

### LINKED ISSUE #455: Declaration Statements with Function Declarations break `EOG` edges
It seems that call expressions are missing `EOG` edges. I suspect, that this may have been introduced by #357.

Consider the following snippet from the Codyze test: 
```cpp
class Test {
public:
  int call(int a) {
    return a + 1;
  }
}

int main() {
  int foo = 42;
  foo = 3;
  Test t();
  t.call(foo);
}
```

This leads to the following EOG, which as you can see, contains no EOG edges to the call expression.

image

oxisto:
Fixes #455 

Also addresses problems found in https://github.com/Fraunhofer-AISEC/codyze/pull/184

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -35,7 +35,7 @@ plugins {
     id("org.sonarqube") version "3.3"
     id("com.diffplug.spotless") version "5.14.0"
     id("com.github.johnrengelman.shadow") version "7.0.0" apply false
-    kotlin("jvm") version "1.4.32" apply false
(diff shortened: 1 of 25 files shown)


oxisto:
@konradweiss Can you please have a look at this. After my fix, a lot of other tests that test the EOG order fail. But, the questions is, whether those tests are actually correct.

PR #462, opened 2021-07-07
TypeScript language frontend

oxisto:
This PR adds *experimental* support for a TypeScript parser. It is definitely not feature complete, but can be used to parse simple typescript snippets through the official `typescript` parser written in TypeScript / nodejs. It includes a simple nodejs script that invokes this parser in `src/main/nodejs`. It basically dumps the AST in a JSON structure on stdout and this input is parsed by `TypeScriptLanguageFrontend`.

This frontend should be able to parse the following languages / dialects, but only TypeScript has been tested extensively:

- JavaScript (`.js`)
- JavaScript React (`.jsx`)
- TypeScript (`.ts`)
- TypeScript React (`.tsx`)

Work that still needs to be done:
- [x] Support basic expressions, such as calls and field access
- [x] Support for simple classes and interfaces (no inheritance)
- [x] Very basic React support, it will generate generic `Expression`s with the JSX content
- [x] Adjust README with instructions
- [x] Make it build on GitHub CI
- [x] Bundle nodejs parser when building on JitPack
- [x] Parse function-based React components
- [x] Support annotation parser (aka Decorators)


oxisto:
A lot of React components are function-based and are directly declared as a `VariableDeclaration` statement on file-level. This is currently blocked by #464, so I suggest to wait for the following PR(s) to merged first:
- #468

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/typescript/ExpressionHandler.kt
```diff
@@ -0,0 +1,243 @@
...
+                    fqn,
+                    memberExpression.base,
+                    member,
+                    ".",
+                    this.lang.getCodeFromRawNode(node)
+                )
+        } else {
+            val name = this.lang.getIdentifierName(node)
+
+            // TODO: fqn - how?
```
konradweiss, line 229: Is this solved? Is the issue that we don't know how to get the fqn or that in the typescript names are hard to map to the fqn concept?

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/typescript/TypeScriptLanguageFrontend.kt
```diff
@@ -0,0 +1,190 @@
...
+            return astNode.code
+        } else {
+            null
+        }
+    }
+
+    override fun <T : Any?> getLocationFromRawNode(astNode: T): PhysicalLocation? {
+        return if (astNode is TypeScriptNode) {
+            // TODO: LSP region
+            val location = PhysicalLocation(File(astNode.location.file).toURI(), Region())
```
konradweiss, line 118: Is this still using the default region?

konradweiss (changes requested):
Two comment should be resolved or if not relevant, deleted. If the have more complex implications to be solved we can still merge without them us it is an experimental branch

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/typescript/ExpressionHandler.kt
oxisto: Not really solved, for now, I assume `fqn = name`.

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/typescript/TypeScriptLanguageFrontend.kt
oxisto: Yes, this is a bit more complicated because of the different code region formats. I would do this in a separate PR.

konradweiss (approved):

Issue #464, opened 2021-07-07
Store Record bodies / TranslationUnit bodies in the graph


maximiliankaul:
Currently, we cannot handle code like the following python snippet:
```Python
print("HELLO")
class Foo:
  print("WORLD")
```
Neither of the two print statements can be stored in CPG as the RecordDeclaration, NamespaceDeclaration, TranslationUnitDeclaration do not supports a `body` (or similar) container.
This issue is likely relevant for other scripting languages besides python, too.

oxisto:
I am also running into this issue with TypeScript - or any other script-based language for that matter. I will try to set up a call next week once every one is back in office.

Issue #432, opened 2021-05-20
Cannot build with -Pexperimental


jiechenjiechen:
I tried to build with the following command under bash in Mac:

`CPG_PYTHON_VIRTUALENV=py39; gradlew -Pexperimental build`

where py39 is environment where I have installed jep. My default java version is 15

```
$ java -version
java version "15.0.2" 2021-01-19
Java(TM) SE Runtime Environment (build 15.0.2+7-27)
Java HotSpot(TM) 64-Bit Server VM (build 15.0.2+7-27, mixed mode, sharing)
```

When building, the following errors occur

```
> Task :cpg-library:test

PythonFrontendTest > testSimpleClass() FAILED
    java.util.concurrent.ExecutionException at PythonFrontendTest.kt:251
        Caused by: java.lang.UnsatisfiedLinkError at ClassLoader.java:2447

PythonFrontendTest > testCtor() FAILED
    java.util.concurrent.ExecutionException at PythonFrontendTest.kt:455
        Caused by: java.lang.UnsatisfiedLinkError at Jep.java:-2

PythonFrontendTest > testSelf() FAILED
    java.util.concurrent.ExecutionException at PythonFrontendTest.kt:369
        Caused by: java.lang.UnsatisfiedLinkError at Jep.java:-2

PythonFrontendTest > testIf() FAILED
    java.util.concurrent.ExecutionException at PythonFrontendTest.kt:207
        Caused by: java.lang.UnsatisfiedLinkError at Jep.java:-2

PythonFrontendTest > testLiteral() FAILED
    java.util.concurrent.ExecutionException at PythonFrontendTest.kt:56
        Caused by: java.lang.UnsatisfiedLinkError at Jep.java:-2

PythonFrontendTest > testFunctionDeclaration() FAILED
    java.util.concurrent.ExecutionException at PythonFrontendTest.kt:111
        Caused by: java.lang.UnsatisfiedLinkError at Jep.java:-2

PythonFrontendTest > testIfExpr() FAILED
    java.util.concurrent.ExecutionException at PythonFrontendTest.kt:312
        Caused by: java.lang.UnsatisfiedLinkError at Jep.java:-2

225 tests completed, 7 failed
```

I traced the errors, for example, the first one. The first line reads:

`java.util.concurrent.ExecutionException: java.lang.UnsatisfiedLinkError: no jep in java.library.path: /Users/chenjie/lab/SDG/cpg/cpg-library/src/main/golang`

[...]

PR #465, opened 2021-07-07
Python: add new Python features

maximiliankaul:
This branch adds new Python features and fixes bugs identified with the code snippet in #432 
This is not a complete fix, as we currently do not store expressions / statements outside of records / functions in the cpg (see issue #464)
Changes:
- return statements can now hold a value
- lists of statements are now either wrapped in a compound statement (size > 1) or directly returned
- more tests


oxisto:
Is there a particular reason to not wrap statement of length in a CompoundStatement. Is it just for convinence or did it cause any of the bugs in the mentioned issue? Not 100% sure, how consistently we do this in other languages. The CPG structure explicitly allows this, so I guess it should be ok.

cpg-library/src/main/python/cpg.py
```diff
@@ -1178,6 +1194,7 @@ def visit_ClassDef(self, node):
                 # TODO what to do about expressions inside a class?
                 self.visit(b)
             else:
+                print(b)
```
oxisto, line 1197: Remove print

cpg-library/src/test/java/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt
```diff
@@ -506,4 +499,104 @@ class PythonFrontendTest : BaseTest() {
...
+        assertEquals(clsC1.fields.size, 1)
+        val field = clsC1.fields[0]
+        assertNotNull(field)
+        assertEquals(field.name, "total")
+        // TODO assert initializer "total = 0"
+        val meth = clsC1.methods[0]
+        assertNotNull(meth)
+        assertEquals(meth.name, "inc")
+        assertEquals(meth.recordDeclaration, clsC1)
+        val selfReceiver = meth.receiver
```
oxisto, line 573: some newlines now and then would be good for readability.

maximiliankaul:
> Is there a particular reason to not wrap statement of length == 1 in a CompoundStatement? Is it just for convinence or did it cause any of the bugs in the mentioned issue? Not 100% sure, how consistently we do this in other languages. The CPG structure explicitly allows this, so I guess it should be ok.

I just thought this would be nicer to work with. There is no reason that I can think of not to wrap single statements in a CompoundStatement. Do you prefer to have everything in a CompoundStatement?

oxisto:
> > Is there a particular reason to not wrap statement of length == 1 in a CompoundStatement? Is it just for convinence or did it cause any of the bugs in the mentioned issue? Not 100% sure, how consistently we do this in other languages. The CPG structure explicitly allows this, so I guess it should be ok.
> 
> I just thought this would be nicer to work with. There is no reason that I can think of not to wrap single statements in a CompoundStatement. Do you prefer to have everything in a CompoundStatement?

I do not have a strong preference for or against it I would say. The only problem we could run into is if one manages to declare variables within the one statement and also uses it. Not sure if this is possible. Because variable resolution depends on a scope (of the compound statement) and with just a single statement, there is also no scope.

maximiliankaul:
> > > Is there a particular reason to not wrap statement of length == 1 in a CompoundStatement? Is it just for convinence or did it cause any of the bugs in the mentioned issue? Not 100% sure, how consistently we do this in other languages. The CPG structure explicitly allows this, so I guess it should be ok.
> > 
> > 
> > I just thought this would be nicer to work with. There is no reason that I can think of not to wrap single statements in a CompoundStatement. Do you prefer to have everything in a CompoundStatement?
> 
> I do not have a strong preference for or against it I would say. The only problem we could run into is if one manages to declare variables within the one statement and also use it. Not sure if this is possible. Because variable resolution depends on a scope (of the compound statement) and with just a single statement, there is also no scope.

I would prefer to have it as it is currently implemented. And it will be an easy fix, if we decide to change this in the future as only the function `make_compound_statement` (currently a very misleading name...) needs to be adjusted.

oxisto (approved):

maximiliankaul:
I would like to merge this PR as some of my other work depends on this. However, there is still the bug with statements outside of functions / records that depends on #464 .

oxisto:
> I would like to merge this PR as some of my other work depends on this. However, there is still the bug with statements outside of functions / records that depends on #464 .

Agreed. We will merge this, so you can continue

PR #477, opened 2021-07-14
Smart guessing of a cast expression vs. call expression, if the type is known

### LINKED ISSUE #474: Java Cast Error
When loading the following C code with the cpg, I get an Exception:

```c
int main() {
	size_t count = (size_t)(42);
	return 0;
}
```

```
java.lang.ClassCastException: class 
org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTUnaryExpression cannot be cast to class 
org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTIdExpression
```

Whereas the following code can be loaded successfully:
```c
int main() {
	size_t count = (size_t)42;
	return 0;
}
```

I'm using the cpg and cpg-neo4j from master branch.
Can you reproduce the error?

oxisto:
C/C++ has some ambuigities, for example in the expression `(size_t)(42)`. If CDT has not seen the type, it will guess, that this is a call expression. We try to be a little bit smarter than CDT and check our type manager, if the type is already known, for example because it was used as a type in a LHS of an assignment. This would also allow us to pre-fill the `TypeManager` with some common types.

Fixes #474

### DIFF
--- a/cpg-library/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
+++ b/cpg-library/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
@@ -240,6 +240,10 @@ public class TypeManager {
     return secondOrderTypes;
   }
 
+  public boolean typeExists(String name) {
+    return firstOrderTypes.stream().anyMatch(type -> type.getRoot().getName().equals(name));
+  }
+
   private TypeManager() {}
 
   public static TypeManager getInstance() {
--- a/cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
+++ b/cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.java
@@ -446,12 +446,14 @@ public class DeclarationHandler extends Handler<Declaration, IASTDeclaration, CX
       handleTemplateUsage((CPPASTNamedTypeSpecifier) declSpecifier, ctx, sequence);
     } else {
       for (IASTDeclarator declarator : ctx.getDeclarators()) {
+        String typeString = getTypeStringFromDeclarator(declarator, ctx.getDeclSpecifier());
+
+        // make sure, the type manager knows about this type before parsing the declarator
+        Type result = TypeParser.createFrom(typeString, true, lang);
+
         ValueDeclaration declaration =
             (ValueDeclaration) this.lang.getDeclaratorHandler().handle(declarator);
 
-        String typeString = getTypeStringFromDeclarator(declarator, ctx.getDeclSpecifier());
-
-        Type result = TypeParser.createFrom(typeString, true, lang);
         declaration.setType(result);
 
         // cache binding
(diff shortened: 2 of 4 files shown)


peckto:
Involving the `TypeManager` to solve ambiguities is a good idea!
This could also be handy to solve other C/C++ ambiguities.

I used the following code to test this PR:
```c
#include <stddef.h>
#include <stdint.h>

int foo(int a) {
	return a;
}


int main() {
	size_t i7 = (uint8_t)(42);  // cast
	size_t i8 = (uint8_t)42;    // cast
	size_t i1 = (size_t)(42);   // cast
	size_t i2 = (size_t)42;     // cast
	size_t i3 = (foo)(42);      // call
	size_t i4 = foo(42);        // call
	uint8_t i5 = (bar)(42);     // call
	uint8_t i6 = bar(42);       // call
	return 0;
}
```

Without loading the includes, only the `(uint8_t)(42)` is interpreted wrong (as CallExpression), but this is expected as the type is not known at this position.
When loading the includes, all expressions are parsed correctly.

One step further would be, to run the parsing a second time, preserving the `TypeManager` state.

How does the `TypeManager` handle different scopes? (eg. namespace, translation unit)

oxisto:
> Involving the `TypeManager` to solve ambiguities is a good idea!
> This could also be handy to solve other C/C++ ambiguities.

I am still not quite sure if it is or not :) Or rather whether this should actually be done in a later pass.

> 
> I used the following code to test this PR:
> 
> ```c
> #include <stddef.h>
> #include <stdint.h>
> 
> int foo(int a) {
> 	return a;
> }
> 
> 
> int main() {
> 	size_t i7 = (uint8_t)(42);  // cast
> 	size_t i8 = (uint8_t)42;    // cast
> 	size_t i1 = (size_t)(42);   // cast
> 	size_t i2 = (size_t)42;     // cast
> 	size_t i3 = (foo)(42);      // call
> 	size_t i4 = foo(42);        // call
> 	uint8_t i5 = (bar)(42);     // call
> 	uint8_t i6 = bar(42);       // call
> 	return 0;
> }
> ```
> 
> Without loading the includes, only the `(uint8_t)(42)` is interpreted wrong (as CallExpression), but this is expected as the type is not known at this position.
> When loading the includes, all expressions are parsed correctly.
> 
> One step further would be, to run the parsing a second time, preserving the `TypeManager` state.
> 
> How does the `TypeManager` handle different scopes? (eg. namespace, translation unit)

Unfortunately, it does not at all, so this is more of a work around, because currently, all types are "global"

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
```diff
@@ -240,6 +240,10 @@ public ParameterizedType createOrGetTypeParameter(
     return secondOrderTypes;
   }
 
+  public boolean typeExists(String name) {
+    return firstOrderTypes.stream().anyMatch(type -> type.getRoot().getName().equals(name));
```
oxisto, line 244: Maybe needs to be synchronised or in a synchronised block, once we merge in #444

PR #489, opened 2021-07-30
Improvements to the Go language frontend

oxisto:
* Added support for parsing `ForStatement` in Go. This is also graph changing, as we needed to change the type of the iteration expression to `Statement`.
* Handling `IncDecStmt` as `UnaryOperator`.
* Preparing Go modules support. This is not active yet, just the prep work
* Proper resolution of calls across packages, based on package names. This could be re-used in other language frontends for scoped resolution, but is currently limited to Go

### DIFF
--- a/cpg-library/src/main/golang/frontend/handler.go
+++ b/cpg-library/src/main/golang/frontend/handler.go
@@ -30,10 +30,14 @@ import (
 	"fmt"
 	"go/ast"
 	"go/token"
+	"io/ioutil"
 	"log"
+	"os"
+	"path"
 	"strconv"
 	"strings"
 
+	"golang.org/x/mod/modfile"
 	"tekao.net/jnigi"
 )
 
@@ -46,6 +50,36 @@ func getImportName(spec *ast.ImportSpec) string {
 	var paths = strings.Split(path, "/")
 
 	return paths[len(paths)-1]
+}
+
+func (frontend *GoLanguageFrontend) ParseModule(topLevel string) (exists bool, err error) {
+	frontend.LogInfo("Looking for a go.mod file in %s", topLevel)
+
+	mod := path.Join(topLevel, "go.mod")
+
+	if _, err := os.Stat(mod); err != nil {
+		if os.IsNotExist(err) {
+			frontend.LogInfo("%s does not exist", mod)
+
+			return false, nil
+		}
+	}
+
+	b, err := ioutil.ReadFile(mod)
+	if err != nil {
+		return true, fmt.Errorf("could not read go.mod: %w", err)
+	}
+
+	module, err := modfile.Parse(mod, b, nil)
+	if err != nil {
+		return true, fmt.Errorf("could not parse mod file: %w", err)
+	}
+
+	frontend.Module = module
+
+	frontend.LogInfo("Go application has module support with path %s", module.Module.Mod.Path)
+
+	return true, nil
 }
 
 func (this *GoLanguageFrontend) HandleFile(fset *token.FileSet, file *ast.File, path string) (tu *cpg.TranslationUnitDeclaration, err error) {
--- a/cpg-library/src/main/golang/frontend/handler.go
+++ b/cpg-library/src/main/golang/frontend/handler.go
@@ -440,6 +440,36 @@ func (this *GoLanguageFrontend) handleBlockStmt(fset *token.FileSet, blockStmt *
 	return c
 }
 
+func (this *GoLanguageFrontend) handleForStmt(fset *token.FileSet, forStmt *ast.ForStmt) *cpg.ForStatement {
+	this.LogDebug("Handling for statement: %+v", *forStmt)
+
+	f := cpg.NewForStatement(fset, forStmt)
+
+	if statement := this.handleStmt(fset, forStmt.Init); statement != nil {
+		f.SetInitializerStatement(statement)
+	}
+
+	if condition := this.handleExpr(fset, forStmt.Cond); condition != nil {
+		f.SetCondition(condition)
+	}
+
+	var scope = this.GetScopeManager()
+
+	scope.EnterScope((*cpg.Node)(f))
+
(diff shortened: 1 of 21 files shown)


oxisto:
I created another issue that the scope manager needs harmonising, see #499.

cpg-library/src/main/golang/frontend/handler.go
```diff
@@ -440,6 +474,36 @@ func (this *GoLanguageFrontend) handleBlockStmt(fset *token.FileSet, blockStmt *
 	return c
 }
 
+func (this *GoLanguageFrontend) handleForStmt(fset *token.FileSet, forStmt *ast.ForStmt) *cpg.ForStatement {
+	this.LogDebug("Handling for statement: %+v", *forStmt)
+
+	f := cpg.NewForStatement(fset, forStmt)
+
+	if statement := this.handleStmt(fset, forStmt.Init); statement != nil {
```
maximiliankaul, line 482: statement -> initStatement would be less confusing

cpg-library/src/main/golang/frontend/handler.go
```diff
@@ -440,6 +474,36 @@ func (this *GoLanguageFrontend) handleBlockStmt(fset *token.FileSet, blockStmt *
...
+		f.SetInitializerStatement(statement)
+	}
+
+	if condition := this.handleExpr(fset, forStmt.Cond); condition != nil {
+		f.SetCondition(condition)
+	}
+
+	var scope = this.GetScopeManager()
+
+	scope.EnterScope((*cpg.Node)(f))
```
maximiliankaul, line 492: shouldn't this be before the initializer statement?

cpg-library/src/main/golang/frontend/handler.go
```diff
@@ -440,6 +474,36 @@ func (this *GoLanguageFrontend) handleBlockStmt(fset *token.FileSet, blockStmt *
...
+
+	if body := this.handleStmt(fset, forStmt.Body); body != nil {
+		f.SetStatement(body)
+	}
+
+	if iter := this.handleStmt(fset, forStmt.Post); iter != nil {
```
maximiliankaul, line 498: moving this in front of handling the body would feel more natural to me

cpg-library/src/main/golang/frontend/handler.go
```diff
@@ -460,6 +524,26 @@ func (this *GoLanguageFrontend) handleReturnStmt(fset *token.FileSet, returnStmt
 	return r
 }
 
+func (this *GoLanguageFrontend) handleIncDecStmt(fset *token.FileSet, incDecStmt *ast.IncDecStmt) *cpg.UnaryOperator {
+	this.LogDebug("Handling decimal increment statement: %+v", *incDecStmt)
+
+	u := cpg.NewUnaryOperator(fset, incDecStmt)
+
+	if incDecStmt.Tok == token.INC {
+		u.SetOperatorCode("++")
```
maximiliankaul, line 533: cpg already provides this as a public field of UnararyOperator: OPERATOR_POSTFIX_INCREMENT
shouldn't this be used instead of hardcoding "++"?

cpg-library/src/main/golang/frontend/handler.go
```diff
@@ -460,6 +524,26 @@ func (this *GoLanguageFrontend) handleReturnStmt(fset *token.FileSet, returnStmt
...
+	}
+
+	if incDecStmt.Tok == token.DEC {
+		u.SetOperatorCode("--")
```
maximiliankaul, line 537: see "++" above

cpg-library/src/main/golang/frontend/handler.go
```diff
@@ -708,7 +799,20 @@ func (this *GoLanguageFrontend) handleCallExpr(fset *token.FileSet, callExpr *as
...
+		pos := strings.LastIndex(name, ".")
+		if pos != -1 {
+			fqn := name
+
+			c.SetFqn(fqn)
+
+			// need to have the short name
+			c.SetName(name[pos+1:])
+		} else {
+			c.SetName(name)
```
maximiliankaul, line 813: is it ok to not set fqn at all?

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/ScopeManager.java
```diff
@@ -563,8 +564,45 @@ public ValueDeclaration resolve(DeclaredReferenceExpression ref) {
...
+   * Tries to resolve a function in a call expression.
+   *
+   * @param call the call expression
+   * @return a list of possible functions
+   */
   public List<FunctionDeclaration> resolveFunction(CallExpression call) {
-    return resolveFunction(currentScope, call);
+    var scope = currentScope;
+
+    // First, we need to check, whether we have some kind of scoping. this is currently only limited
```
maximiliankaul, line 576: Do we want to have language specific code here?

maximiliankaul (commented):
Don't know much about go. LGTM in general.

cpg-library/src/main/golang/frontend/handler.go
oxisto: In theory yes. Accessing this in the go frontend tends to be tedious because I need to go up the stack from Go to Java again. Does not seem to be satisfied for a simple "++" I would say.

cpg-library/src/main/golang/frontend/handler.go
oxisto: To be honest: I have no idea :)

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/ScopeManager.java
oxisto: In the long run: No, that's why I created #499. I can do a quick check, whether this would destroy anything if I remove the language frontend guard. On the other hand, we do already have a language specific resolving for CXX.

PR #500, opened 2021-08-04
ForEachStatement: `variable` is now a `Statement` instead of a `Declaration`

maximiliankaul:
This PR changes the graph: `ForEachStamtents`'s `variable` field becomes a `Statement` (was a `Declaration`) to allow for languages that do not require new variables in for loops but also allow references.

Example:
```python
iter = 1
for iter in [1,2,3]:
  pass
# iter is now 3
```

### DIFF
--- a/cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/StatementHandler.java
+++ b/cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/StatementHandler.java
@@ -25,6 +25,8 @@
  */
 package de.fraunhofer.aisec.cpg.frontends.cpp;
 
+import static de.fraunhofer.aisec.cpg.graph.NodeBuilder.newDeclarationStatement;
+
 import de.fraunhofer.aisec.cpg.frontends.Handler;
 import de.fraunhofer.aisec.cpg.graph.NodeBuilder;
 import de.fraunhofer.aisec.cpg.graph.TypeManager;
@@ -272,7 +274,7 @@ class StatementHandler extends Handler<Statement, IASTStatement, CXXLanguageFron
     lang.getScopeManager().enterScope(statement);
 
     Declaration decl = this.lang.getDeclarationHandler().handle(ctx.getDeclaration());
-    DeclarationStatement var = new DeclarationStatement();
+    DeclarationStatement var = newDeclarationStatement(decl.getCode());
     var.setSingleDeclaration(decl);
     Statement iterable = this.lang.getExpressionHandler().handle(ctx.getInitializerClause());
 
@@ -305,8 +307,7 @@ class StatementHandler extends Handler<Statement, IASTStatement, CXXLanguageFron
       TypeManager.getInstance().handleTypedef(ctx.getRawSignature());
       return null;
     } else {
-      DeclarationStatement declarationStatement =
-          NodeBuilder.newDeclarationStatement(ctx.getRawSignature());
+      DeclarationStatement declarationStatement = newDeclarationStatement(ctx.getRawSignature());
 
       Declaration declaration = this.lang.getDeclarationHandler().handle(ctx.getDeclaration());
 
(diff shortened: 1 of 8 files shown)


cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/StatementHandler.java
```diff
@@ -271,7 +271,9 @@ private ForEachStatement handleForEachStatement(CPPASTRangeBasedForStatement ctx
     ForEachStatement statement = NodeBuilder.newForEachStatement(ctx.getRawSignature());
     lang.getScopeManager().enterScope(statement);
 
-    Declaration var = this.lang.getDeclarationHandler().handle(ctx.getDeclaration());
+    Declaration decl = this.lang.getDeclarationHandler().handle(ctx.getDeclaration());
+    DeclarationStatement var = new DeclarationStatement();
```
oxisto, line 275: It would be better to use the `NodeBuilder` pattern. We discourage directly creating these objects, as the node builder sets some "required" parameters. And yes I am aware how your python code works, let's keep that as a future issue ;)

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/StatementAnalyzer.java
```diff
@@ -184,12 +184,11 @@ private ForEachStatement handleForEachStatement(Statement stmt) {
     de.fraunhofer.aisec.cpg.graph.statements.Statement iterable =
         lang.getExpressionHandler().handle(forEachStmt.getIterable());
 
-    if (!(var instanceof DeclarationStatement
-        && ((DeclarationStatement) var).isSingleDeclaration())) {
-      log.error("more or unknown decl in foreach");
-    } else {
-      statement.setVariable(((DeclarationStatement) var).getSingleDeclaration());
-    }
+    if (!(var instanceof DeclarationStatement))
```
oxisto, line 187: Minor nitpick: please always use `{` even for one-line ifs.

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/StatementAnalyzer.java
```diff
@@ -184,12 +184,11 @@ private ForEachStatement handleForEachStatement(Statement stmt) {
...
+      log.error("Expected a DeclarationStatement but received: {}", var.getName());
+    else if (!((DeclarationStatement) var).isSingleDeclaration())
```
oxisto, line 189: I think you can remove this, I think it is ok to check whether this is a declaration statement, but any more checks are not necessary

cpg-library/src/test/java/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontendTest.java
```diff
@@ -137,10 +137,15 @@ void testForeach() throws Exception {
     assertEquals(ls, ((DeclaredReferenceExpression) forEachStatement.getIterable()).getRefersTo());
 
     // should declare String s
-    VariableDeclaration s = (VariableDeclaration) forEachStatement.getVariable();
+    Statement s = forEachStatement.getVariable();
```
oxisto, line 140: Just use `var´ in case we change the field type again.

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/java/StatementAnalyzer.java
```diff
@@ -184,12 +184,11 @@ private ForEachStatement handleForEachStatement(Statement stmt) {
...
+      log.error("Expected a single DeclarationStatement.");
+    else statement.setVariable(var);
```
oxisto, line 191: While you are at it: Could you please rename `var` to `variable`, so that it does not clash with the keyword var of later Java version / Kotlin.

oxisto (commented):
Just some minor things

oxisto (approved):

PR #513, opened 2021-08-13
Mk/python rework

maximiliankaul:
This is a WIP PR. The aim is to rewrite the python frontend to use a similar approach as the other frontends (`handle_statement` / `handle_expresssion` / ... instead of recursively traversing the AST). The old approach made some changes quiet difficult to implement.

TODOs before this can be merged:
- [ ] write comments
- [ ] transfer all old functionality to this branch
- [ ] add source code to CPG nodes
- [ ] fix the last test (9 out of 10 currently pass)
- [ ] move more stuff into `handle_xxx()` functions
- [ ] find an elegant solution to build and install the python package
- [ ] ...

How to try out this PR on your system:
1. you have to be in the same virtualenv that the cpg is using
2. navigate to `cpg-library/src/main/python/CPGPython/`
3. optionally remove an old package ` pip3 uninstall CPGPython`
4. build the python package: `python3 -m build`
5. install the package to you local/active virtualenv: `pip3 install dist/CPGPython-0.0.1-py3-none-any.whl`


maximiliankaul:
How should we handle `import` and `import from`?
Note: This is scoped in python.

oxisto (approved):
It would be good to extract more of the if/elif constructs into individual `handle_XXX` functions, but let's keep that for the second round of refactoring. Let's merge it as experimental in order to progress.

PR #521, opened 2021-08-17
Update JavaLanguageFrontend.java

pangeneral:
ctx should be a JavaParser Node instead of CPG node


oxisto:
Thanks for catching this, this was a mixup of imports. I have fixed this in https://github.com/Fraunhofer-AISEC/cpg/pull/525 and would propose using #525 over this. It fixes the problem my importing `Node` from the java parser and using the FQN of the CPG node. That is the way we do it in the other handlers as well, the language specific nodes should be prioritised in the language frontends.

PR #523, opened 2021-08-18
Refactor the `dummy` field into proper usage of `implicit` and `inferred`

konradweiss:
Renames the dummy property to inferred as a node that is somewhere present in compilable code but may be missing and is thus inferred. This is in contrast to implicit nodes that are not directly present (not visible) but exists as target(impl- this) or exists as its behavior is executed (implicit return.)


oxisto (approved):
LGTM

PR #575, opened 2021-10-03
Added `--enable-experimental-typescript` to enable TypeScript in `cpg-neo4j`

### LINKED ISSUE #574: using typescript in cpg-neo4j
hi all. may i know if it possible to add support for TypeScript (and its variants *.js, *.jsx etc.) in cpg-neo4j? appreciate if it can be something like `--enable-experimental-typescript` , similar to both Python and Go.

oxisto:
Fixes #574


oxisto:
Can you take a look at this @anisyusof-sc?

PR #589, opened 2021-10-16
Fixes to `VariableUsageResolver`

oxisto:
This fixes a bug in the `VariableUsageResolver`, in which the resolver was trying to resolve a variable, that was already resolved. This can lead to weird behaviours. We always had the stance, that IF the language frontend sets the `refersTo` property, the pass should respect this and not re-set it. There are some particular cases in which the language frontend just knows better.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -33,7 +33,7 @@ plugins {
     signing
 
     id("org.sonarqube") version "3.3"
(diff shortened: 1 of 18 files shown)


oxisto:
@immqu Could you check, if this fixes the issue with functional components in the CloudPG?

PR #604, opened 2021-10-30
New (experimental) C/C++ language frontend based on `tree-sitter`

oxisto:
As discussed in internal rounds, we are on a long-term plan to replace Eclipse CDT (for various reasons). This PR contains the first draft of a language frontend based on https://tree-sitter.github.io.

It uses https://github.com/oxisto/kotlintree to bridge tree-sitter into Java/Kotlin


konradweiss:
@peckto We should consider if we can make something out of this PR or close it.

oxisto:
I think we can close it for now

PR #610, opened 2021-11-17
Adding parsing of comments to golang 

konradweiss:
It uses `ast.CommentMap` to access an associated map of comments to their `ast.Node` and sets the comment using `SetComment`.


cpg-library/src/main/java/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontend.kt
```diff
@@ -54,6 +58,98 @@ class GoLanguageFrontend(config: TranslationConfiguration, scopeManager: ScopeMa
         return parseInternal(file.readText(Charsets.UTF_8), file.path, config.topLevel.absolutePath)
     }
 
+    /**
+     * This function matches a comment to the closest node according to a heuristic: Comments are
+     * assigned to the closest successor node on their ast hierarchy level. Only Exception, if they
+     * don't have a successor starting in the same line but they have a predecessor in the same
+     * line, the comment is matched to that closest predecessor.
+     */
+    fun matchCommentToNode(comment: String, location: Region, tu: TranslationUnitDeclaration) {
```
oxisto, line 67: As discussed, for me this looks very similar to what https://pkg.go.dev/go/ast#NewCommentMap is doing. We should double-check whether we could use this instead of our own implementation.

PR #616, opened 2021-11-28
Parse comments and locations feature to the typescript frontend

konradweiss:
Adds Comment parsing to the Typescript language frontend.

The position of comments are not provided by the node parser itself. It was therefore necessary to read them out manually, here I had to resort to a search by regex. Parsing comments with a regex is an imperfect solution as a context sensitive parser would be necessary. However it should be sufficient to add this feature without switching out the current parser. 
This PR:

- Computes and adds regions of the form line:column to line:column to every node, conforming to other frontends.
- Corrects a positioning error the parser makes when computing the starting position
- Parses comments from the file through a regular expression
- matches comments

### DIFF
--- a/cpg-library/src/main/golang/lib/cpg/main.go
+++ b/cpg-library/src/main/golang/lib/cpg/main.go
@@ -89,8 +89,7 @@ func Java_de_fraunhofer_aisec_cpg_frontends_golang_GoLanguageFrontend_parseInter
 	}
 
 	goFrontend.CommentMap = ast.NewCommentMap(fset, file, file.Comments)
-
-	goFrontend.LogDebug("%+v", goFrontend.CommentMap)
+(??)
 
 	_, err = goFrontend.ParseModule(string(topLevel.([]byte)))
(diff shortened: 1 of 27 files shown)


cpg-library/src/main/golang/lib/cpg/main.go
```diff
@@ -89,6 +89,7 @@ func Java_de_fraunhofer_aisec_cpg_frontends_golang_GoLanguageFrontend_parseInter
 	}
 
 	goFrontend.CommentMap = ast.NewCommentMap(fset, file, file.Comments)
+(??)
```
oxisto, line 92: ?

oxisto (commented):
Looks good except some weird issues in a go file.

cpg-library/src/main/golang/lib/cpg/main.go
konradweiss: Apparently a mistake during solving mergconflicts

PR #618, opened 2021-12-01
Simpler openssl example

oxisto:
For now, this uses C strings instead of `std::string` for an easier data flow analysis.


KuechA (approved):
The changes in `client.cpp` look good to me.

PR #619, opened 2021-12-02
Using DFG in Value Evaluator

oxisto:
This PR adds support for data flow resolution to the `ValueEvaluator`. Instead of directly going to the declaration of an reference, we rather use the data flow edges to find its "last assignment".


cpg-library/src/main/java/de/fraunhofer/aisec/cpg/analysis/ValueEvaluator.kt
```diff
@@ -71,18 +82,36 @@ class ValueEvaluator(
...
             is Literal<*> -> {
                 return expr.value
             }
-            is DeclaredReferenceExpression -> return resolveDeclaration(expr.refersTo)
+            // For a reference, we are interested into its last assignment into the reference
+            // denoted by the previous DFG edge
+            is DeclaredReferenceExpression -> {
+                val prevDFG = expr.prevDFG
+
+                // We could potentially have more incoming DFG nodes, so we want to filter for
```
maximiliankaul, line 98: Did we not say to abort in this case?

maximiliankaul (approved):

cpg-library/src/main/java/de/fraunhofer/aisec/cpg/analysis/ValueEvaluator.kt
oxisto: We could this as well. At some point we should probably have some kind of error indication, e.g. `TooManySolutions`

oxisto: As per discussion, we abort in this case now.

PR #627, opened 2021-12-08
Revert "Parse comments and locations feature to the typescript frontend"

oxisto:
Unfortunately we need to revert this PR, at least for the next CPG release as there is a DRASTIC slowdown of parsing C++ files. For an example file (https://github.com/openssl/openssl/blob/master/crypto/ec/curve448/curve448_tables.c) with a lot of nodes, there was an increase in overall translation time from 19s to over 70s. 

Reverts Fraunhofer-AISEC/cpg#616


oxisto:
As an alternative, I have reverted only the C++ parts in https://github.com/Fraunhofer-AISEC/cpg/pull/626

PR #632, opened 2021-12-10
Extracting LLVM into separate module

### LINKED ISSUE #633: Make LLVM language optional
Can we make the LLVM language feature optional?

I've just build Codyze locally. The new LLVM language feature with its dependency `org.bytedeco:llvm-platform` added around 1.4 GB of additional libraries. Among others, it downloaded multiple platforms and CPU architectures that might not all be required.

@oxisto Do we really want to create Codyze releases of around 1.6 GB?

oxisto:
This PR attempts to extract LLVM into its own module, that we will also publish independently on Maven Central. In the long run we might want to do this for all / most of the languages and possibly have a structure like:

- cpg (depends on all of the sub packages, similar to the `de.fraunhofer.aisec.cpg:cpg` artefact we have now)
- cpg-core
- cpg-analysis
- cpg-llvm
- cpg-cpp
- ...

This is done rather hastily, so we can probably improve this in the future.

Closes #633


oxisto:
Before we publish this on maven central, we should have a consistent naming scheme. Do we want `cpg-llvm`, `cog-python`, etc. oder `cpg-language-llvm`, `cpg-language-python` and then possibly something like `cpg-analysis`?

cc @fwendland @maximiliankaul @KuechA @konradweiss

KuechA:
Personally, I prefer the naming scheme `cpg-language-*`.

fwendland:
For the LLVM part, would it be possible to somehow configure the platform and the cpu architecture?

It might already help if one does not have to download all available architectures. Can we expose the architecture configuration from [here](https://github.com/bytedeco/javacpp-presets/wiki/Reducing-the-Number-of-Dependencies#including-binaries-only-for-mobile-platforms)?

PR #663, opened 2022-01-12
Added support for a compilation database

xaviergeorge:
With this merge, the user will be able to evaluate the JSON compilation database of a project(CXX projects) by just providing the path of the compile_commands.json file. This change is made in the cpg-neo4j module. The compile_commands.json file can be given by adding --json-compilation-database in the program arguments followed by the path of the compile_commands.json file.


cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -61,6 +66,14 @@ private const val DEFAULT_USER_NAME = "neo4j"
 private const val DEFAULT_PASSWORD = "password"
 private const val DEFAULT_SAVE_DEPTH = -1
 
+data class CompilationDbStructure(
```
oxisto, line 69: I think this should probably be move to cpg-core and named something like `CompliationDatabaseEntry` or rather something related to "command objects" as they are called in the original spec.

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -70,14 +83,25 @@ private const val DEFAULT_SAVE_DEPTH = -1
...
-    @CommandLine.Parameters(
-        arity = "1..*",
-        description =
-            [
-                "The paths to analyze. If module support is enabled, the paths will be looked at if they contain modules"]
-    )
-    private var files: Array<String> = arrayOf(".")
+        @CommandLine.Option(
+            names = ["--json-compilation-database"],
+            description = ["Give the json compilation database file path "]
```
oxisto, line 101:
```suggestion
            description = ["The path to an optional a JSON compilation database"]
```

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -226,12 +250,13 @@ class Application : Callable<Int> {
      */
     @OptIn(ExperimentalPython::class, ExperimentalGolang::class, ExperimentalTypeScript::class)
     private fun setupTranslationConfiguration(): TranslationConfiguration {
-        assert(files.isNotEmpty())
-        val filePaths = arrayOfNulls<File>(files.size)
+        assert(exclusive.files.isNotEmpty())
```
oxisto, line 253: The variable name `exclusive` sounds misleading.

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -243,16 +268,60 @@ class Application : Callable<Int> {
             }
             filePaths[index] = file
         }
+        if (exclusive.jsonCompilationDatabase == "" && exclusive.files.isEmpty()) {
+            throw Error(
+                "Files list is empty or jsonCompilationDatabase is also empty. Please provide --files to evaluate or --json-compilation-database for CXX files"
```
oxisto, line 273:
```suggestion
                "Please provide either a list of files to evaluate or a compilation database using the --json-compilation-database flag"
```

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -70,14 +83,25 @@ private const val DEFAULT_SAVE_DEPTH = -1
...
-        description =
-            [
-                "The paths to analyze. If module support is enabled, the paths will be looked at if they contain modules"]
-    )
-    private var files: Array<String> = arrayOf(".")
+        @CommandLine.Option(
+            names = ["--json-compilation-database"],
+            description = ["Give the json compilation database file path "]
+        )
+        var jsonCompilationDatabase: String = ""
```
oxisto, line 103: Use `String? = null` instead of assigning an empty string, this is cleaner.

oxisto: Actually, you can just use `File? = null` as far as I know, pico will convert a string into a file

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CompilationDB.kt
```diff
@@ -0,0 +1,98 @@
...
+    //    return includeFiles;
+    //  }
+    /**
+     * Gets the include directories the from the string value provided. Example for a compile
+     * commdand is : "/usr/local/bin/g++-7 -I/Users/me/prj/Calendar/calendars -g -std=c++11 -o
+     * calendar_run.dir/main.cpp.o -c /Users/me/prj/Calendar/main.cpp"
+     *
+     * This method returns the include-paths in the above the command.
+     */
+    fun getIncludeDirectories(stringVal: String): List<String>? {
```
oxisto, line 60: this function should probably be something like `parseIncludeDirectories`

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -276,6 +286,11 @@ public Builder sourceLocations(List<File> sourceLocations) {
       return this;
     }
 
+    public Builder loadCompilationDatabase(Map<File, List<String>> compilationDatabase) {
```
oxisto, line 289:
```suggestion
    public Builder useCompilationDatabase(Map<File, List<String>> compilationDatabase) {
```

oxisto: Wouldn't it make more sense to provide the File to the compilation database here and extract the necessary entries in the CXX language frontend?

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -243,16 +268,60 @@ class Application : Callable<Int> {
...
+                .topLevel(topLevel)
                 .defaultPasses()
                 .defaultLanguages()
                 .loadIncludes(loadIncludes)
                 .debugParser(DEBUG_PARSER)
 
+        if (exclusive.jsonCompilationDatabase != "") {
+            val jsonStringFile = File(exclusive.jsonCompilationDatabase).readText()
+            val mapper = ObjectMapper().registerKotlinModule()
+            val obj: List<CompilationDbStructure> = mapper.readValue(jsonStringFile)
```
oxisto, line 289: I would move this functionality into the CXX language frontend, so that we can use it independently of cpg-neo4j

oxisto (changes requested):

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
xaviergeorge: What do you think about `mutuallyExclusiveParameters` instead of `exclusive`?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -140,6 +140,9 @@
    */
   final boolean typeSystemActiveInFrontend;
 
+  /** This is the data structure for storing the compilation database. */
+  final Map<File, List<String>> compilationDatabase;
```
konradweiss, line 144: Some more specific documentation on what is stored in file and its relation to the list of strings that explains the format, e.g. stores a mapping from files that have to be compiled to their file specific compilation parameters/commands

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
konradweiss: If we assume that there are more languages with this format of compilation databases of multiple implementations of a frontend for the same language I would not put it into the specific implementation. What do you guys think?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CompilationDB.kt
```diff
@@ -0,0 +1,98 @@
...
+ */
+package de.fraunhofer.aisec.cpg.frontends.cpp
+
+import org.slf4j.LoggerFactory
+import java.util.*
+
+/** This class is used to extract the details inside a compilation database. */
+object CompilationDB {
+    private val log = LoggerFactory.getLogger(CompilationDB::class.java)
+    // This method is used if you want to get all the files inside a folder
```
konradweiss, line 34: I would remove commented out code, if we ever decide to need this feature in the future we will have to implement it again

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
konradweiss: I can see that, are we also using the data class in cpg-core code? if not we may consider doing so and move it there

konradweiss (changes requested):
overall fine, remove commented out code and add some documentation in form of comments as well as take care of the other reviewers considerations.

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
xaviergeorge: Renamed it to CompliationDatabaseEntry and moved it to `cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CompilationDB.kt`.

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
xaviergeorge: Made the type as File? and added it to the latest commit.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
xaviergeorge: For now, `loadCompilationDatabase(Map<File, List<String>> compilationDatabase)` is  changed to `useCompilationDatabase(File compilationDatabaseFile)` and the function that returns the map(`Map<File, List<String>>`) of compilation database is placed inside `cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CompilationDB.kt`. 
@konradweiss   maybe we can have different implementations for getting a compilation database  for different languages in their respective `/frontends/<language>/CompilationDB.kt`? What do you think about this?

cpg-neo4j/src/main/java/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
xaviergeorge: Added this

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -286,8 +291,14 @@ public Builder sourceLocations(List<File> sourceLocations) {
       return this;
     }
 
-    public Builder loadCompilationDatabase(Map<File, List<String>> compilationDatabase) {
-      this.compilationDatabase = compilationDatabase;
+    public Builder useCompilationDatabase(File compilationDatabaseFile) {
+      Map<File, List<String>> compilationDatabaseForTheFile =
+          CompilationDB.Companion.getCompilationDatabaseFromTheFile(compilationDatabaseFile);
+      if (compilationDatabaseForTheFile != null) {
+        this.compilationDatabase = compilationDatabaseForTheFile;
```
oxisto, line 298: This should be done in the CXX language frontend. The builder should only store the file name of the compilation database in the config

xaviergeorge: @oxisto  This is done so that the list of files that need to be analysed by the cpg can be obtained.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
konradweiss: Yes that option would definitely speak for the approach to move it to the frontend itself like @oxisto proposed

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
xaviergeorge: Added documentation for this in the latest commit.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
oxisto: I know, but the configuration builder should be without any side-effects. It should purely store the necessary information in the configuration (the file to the compilation data base). The language frontend or the translation manager should then use this information to read the compilation database (which has side effects) and set the necessary paths for parsing the files.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
xaviergeorge: @konradweiss is the documentation enough?

oxisto (approved):
I did some minor corrections, e.g. I moved the map of files and include paths into the `CompilationDatabase` class. This way we can use the class in the translation config and later on extend on that.

konradweiss (approved):
Looks good to me, IMO we can merge.

Issue #648, opened 2022-01-04
java.lang.ClassCastException when running a specific file from the abyss project


xaviergeorge:
When cpg runs the file https://github.com/bcgsc/abyss/blob/master/DataBase/DB.h , it gives an error which is 

`java.util.concurrent.ExecutionException: java.lang.ClassCastException: class de.fraunhofer.aisec.cpg.passes.scopes.RecordScope cannot be cast to class de.fraunhofer.aisec.cpg.passes.scopes.FunctionScope (de.fraunhofer.aisec.cpg.passes.scopes.RecordScope and de.fraunhofer.aisec.cpg.passes.scopes.FunctionScope are in unnamed module of loader 'app')
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2063)
	at de.fraunhofer.aisec.cpg_vis_neo4j.Application.call(Application.kt:322)
	at de.fraunhofer.aisec.cpg_vis_neo4j.Application.call(Application.kt:70)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
	at picocli.CommandLine.access$1300(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2346)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2311)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
	at picocli.CommandLine.execute(CommandLine.java:2078)
	at de.fraunhofer.aisec.cpg_vis_neo4j.ApplicationKt.main(Application.kt:349)
Caused by: java.lang.ClassCastException: class de.fraunhofer.aisec.cpg.passes.scopes.RecordScope cannot be cast to class de.fraunhofer.aisec.cpg.passes.scopes.FunctionScope (de.fraunhofer.aisec.cpg.passes.scopes.RecordScope and de.fraunhofer.aisec.cpg.passes.scopes.FunctionScope are in unnamed module of loader 'app')
	at de.fraunhofer.aisec.cpg.passes.EvaluationOrderGraphPass.handleFunctionDeclaration(EvaluationOrderGraphPass.java:366)
[...]

PR #664, opened 2022-01-12
Trying to mitigate a CXX ambiguity in CDT that leads to a crash

oxisto:
This PR tries to mitigate #648 by making the EOG pass more robust as well as trying to recognize a certain CXX ambiguity (a call expression mistaken as a function declaration) by returning a problem declaration instead. This should at least not destroy the remaining AST scopes as it did before.


oxisto:
@xaviergeorge can you have a look at this?

xaviergeorge:
@oxisto  I found that this error was not reported(before your commit) when  
```
   // analyze the body
    if (funcDecl.hasBody()) {
      createEOG(funcDecl.getBody());
    }

    var currentScope = lang.getScopeManager().getCurrentScope();  
```
in the file `cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java` inside the function `handleFunctionDeclaration` was changed to 
```
var currentScope = lang.getScopeManager().getCurrentScope();
// analyze the body
    if (funcDecl.hasBody()) {
      createEOG(funcDecl.getBody());
    }
    

```
The value of the `currentScope` was changed when `funcDecl.hasBody()` was true. In our case, when `srand(time(null))` was processed. Do you think it should be declared earlier or is it another bug?

oxisto:
> @oxisto I found that this error was not reported(before your commit) when
> 
> ```
>    // analyze the body
>     if (funcDecl.hasBody()) {
>       createEOG(funcDecl.getBody());
>     }
> 
>     var currentScope = lang.getScopeManager().getCurrentScope();  
> ```
> 
> in the file `cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java` inside the function `handleFunctionDeclaration` was changed to
> 
> ```
> var currentScope = lang.getScopeManager().getCurrentScope();
> // analyze the body
>     if (funcDecl.hasBody()) {
>       createEOG(funcDecl.getBody());
>     }
>     
> ```
> 
> The value of the `currentScope` was changed when `funcDecl.hasBody()` was true. In our case, when `srand(time(null))` was processed. Do you think it should be declared earlier or is it another bug?

Yes, the value of `currentScope` changed after the body was parsed because our CPP parser thought that `srand(time(null))` was a function declaration (inside a class, making it a method declaration). Therefore, the scope was changed to the current record by parsing the function body of `getRand`. This PR mitigates this problem by detecting this ambiguity and changing it to a `ProblemDeclaration`. This way the scope should stay the same.

xaviergeorge:
> > @oxisto I found that this error was not reported(before your commit) when
> > ```
> >    // analyze the body
> >     if (funcDecl.hasBody()) {
> >       createEOG(funcDecl.getBody());
> >     }
> > 
> >     var currentScope = lang.getScopeManager().getCurrentScope();  
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in the file `cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.java` inside the function `handleFunctionDeclaration` was changed to
> > ```
> > var currentScope = lang.getScopeManager().getCurrentScope();
> > // analyze the body
> >     if (funcDecl.hasBody()) {
> >       createEOG(funcDecl.getBody());
> >     }
> >     
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The value of the `currentScope` was changed when `funcDecl.hasBody()` was true. In our case, when `srand(time(null))` was processed. Do you think it should be declared earlier or is it another bug?
> 
> The value of `currentScope` changed after the body was parsed because our CPP parser thought that `srand(time(null))` was a function declaration (inside a class, making it a method declaration). Therefore, the scope was changed to the current record by parsing the function body of `getRand`. This PR mitigates this problem by detecting this ambiguity and changing it to a `ProblemDeclaration`. This way the scope should stay the same.
> 
> I would intentionally check the current scope after the EOG body parsing to detect similar problems in the future. I have added another check, if the scope is really a `FunctionScope`. The crash before resulted in a non-checked cast of the scope.

Okay. The branch works perfect with the file from the abyss project.

PR #677, opened 2022-01-16
More optimisation by including more properties in `hashCode`

oxisto:
This PR includes more fields in `hashCode`, such as `location`. Therefore, we need to consistently set `location` as soon as possible, e.g. in the `NodeBuilder`.


konradweiss:
Does setting them work for all nodes? Previously we had this handy workflow that the handler adds them in the background based on the node that goes into the handling function, adding it to the node that gets returned. IMO this avoid problems when implementors build their nodes without the builder, e.g. in dependent projects that add new Nodes.

oxisto:
I have included the necessary location properties in the appropriate node builders. Should I include it for all of them, @konradweiss?

konradweiss (approved):
Can be merged IMO @oxisto As I added some changes I will leave it to you to either merge or request other changes

oxisto:
> Can be merged IMO @oxisto As I added some changes I will leave it to you to either merge or request other changes

I did some final touches to de-duplicate some code. Will merge after that.

PR #680, opened 2022-01-20
Added pretty printing of benchmarks

oxisto:
This PR adds a little function that can be used to pretty print benchmark results, e.g. to post them in GitHub issues.


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/Benchmark.kt
```diff
@@ -25,15 +25,102 @@
...
+            }
+        }
+    }
+
+    // table header
+    val dash = lengths.joinToString(" | ", "| ", " |") { ("-".repeat(it)) }
+    var i = 0
+    val header = headers.joinToString(" | ", "| ", " |") { it.padEnd(lengths[i++]) }
+
+    // println(dash)
```
konradweiss, line 91: I would remove this commented out code, if we want this dashes optional I would do so as a configuration. I don't know if the purpose was to keep this as an alternative or was just forgotten

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/Benchmark.kt
```diff
@@ -25,15 +25,102 @@
...
+    println(header)
+    println(dash)
+
+    for (row in table) {
+        var i = 0
+        val line = row.joinToString(" | ", "| ", " |") { it.toString().padEnd(lengths[i++]) }
+        println(line)
+    }
+
+    // println(dash)
```
konradweiss, line 102: Here Too

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/Benchmark.kt
```diff
@@ -25,15 +25,102 @@
...
+    println()
+}
+
+fun relativeOrAbsolute(path: Path, topLevel: Path): Path {
```
konradweiss, line 106: I would comment this as it took me a second to figure out the purpose and I am not sure if I got it right

konradweiss (changes requested):
Some minor remarks but in general happy to merge that

konradweiss (approved):

PR #682, opened 2022-01-21
Add test cases for CXX ambiguities

peckto:
In CXX there is an ambiguity with the statement `(A.B)(C)`.
`B` can either be a method or a function pointer.

Example:
```cpp
struct A {
        void foo(int i) {
        }
};

struct B {
        void (*bar)(int);
};

void bar(int i) {
}

int main() {
        A a;
        a.j = 1;
        B b;
        b.bar = &bar;

        // foo is a method
        (a.foo)(1);     // Crash
        a.foo(2);       // OK

        // bar is a function pointer
        (b.bar)(3);     // Crash
        (*b.bar)(4);    // MemberExpression is missing

        return 0;
}
```

Loading the example code into the cpg will result in a crash for the `(A.B)(C)` statements:
```
java.util.concurrent.ExecutionException: java.lang.ClassCastException: class org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTUnaryExpression cannot be cast to class org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTFieldReference (org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTUnaryExpression and org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTFieldReference are in unnamed module of loader 'app')
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at de.fraunhofer.aisec.cpg_vis_neo4j.Application.call(Application.kt:345)
	at de.fraunhofer.aisec.cpg_vis_neo4j.Application.call(Application.kt:72)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
	at picocli.CommandLine.access$1300(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2346)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2311)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
	at picocli.CommandLine.execute(CommandLine.java:2078)
	at de.fraunhofer.aisec.cpg_vis_neo4j.ApplicationKt.main(Application.kt:372)
Caused by: java.lang.ClassCastException: class org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTUnaryExpression cannot be cast to class org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTFieldReference (org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTUnaryExpression and org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTFieldReference are in unnamed module of loader 'app')
	at de.fraunhofer.aisec.cpg.frontends.cpp.ExpressionHandler.handleFunctionCallExpression(ExpressionHandler.kt:488)
	at de.fraunhofer.aisec.cpg.frontends.cpp.ExpressionHandler._init_$lambda-7(ExpressionHandler.kt:91)
	at de.fraunhofer.aisec.cpg.frontends.Handler.handle(Handler.java:111)
	at de.fraunhofer.aisec.cpg.frontends.cpp.StatementHandler.handleExpressionStatement(StatementHandler.kt:255)
	at de.fraunhofer.aisec.cpg.frontends.cpp.StatementHandler._init_$lambda-4(StatementHandler.kt:56)
	at de.fraunhofer.aisec.cpg.frontends.Handler.handle(Handler.java:111)
	at de.fraunhofer.aisec.cpg.frontends.cpp.StatementHandler.handleCompoundStatement(StatementHandler.kt:289)
	at de.fraunhofer.aisec.cpg.frontends.cpp.StatementHandler._init_$lambda-1(StatementHandler.kt:50)
	at de.fraunhofer.aisec.cpg.frontends.Handler.handle(Handler.java:111)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationHandler.handleFunctionDefinition(DeclarationHandler.kt:162)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationHandler._init_$lambda-3(DeclarationHandler.kt:60)
	at de.fraunhofer.aisec.cpg.frontends.Handler.handle(Handler.java:111)
	at de.fraunhofer.aisec.cpg.frontends.cpp.DeclarationHandler.handleTranslationUnit(DeclarationHandler.kt:529)
	at de.fraunhofer.aisec.cpg.frontends.cpp.CXXLanguageFrontend.parse(CXXLanguageFrontend.kt:226)
	at de.fraunhofer.aisec.cpg.TranslationManager.parse(TranslationManager.kt:340)
	at de.fraunhofer.aisec.cpg.TranslationManager.parseSequentially(TranslationManager.kt:284)
	at de.fraunhofer.aisec.cpg.TranslationManager.runFrontends(TranslationManager.kt:204)
	at de.fraunhofer.aisec.cpg.TranslationManager.analyze$lambda-2(TranslationManager.kt:86)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1700)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1692)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
```

I wrote a hot fix to prevent the cpg from crashing, but the resulting graph does not reconstruct the `MemberExpression` and does not distinguish the ambiguity:
cpg-member-function

I see two main questions:
1) How (and at which point) can we distinguish the ambiguity?
2) How should the graph look like for `(A.B)(C)` when `B` is a function pointer?

Clang AST looks for the above example as follows:
```
    |-CXXMemberCallExpr 0x55dd05798630 <line:19:9, col:18> 'void'
    | |-ParenExpr 0x55dd057985f0 <col:9, col:15> '<bound member function type>'
    | | `-MemberExpr 0x55dd057985c0 <col:10, col:12> '<bound member function type>' .foo 0x55dd05769fd8
    | |   `-DeclRefExpr 0x55dd057985a0 <col:10> 'A' lvalue Var 0x55dd0576a708 'a' 'A'
    | `-IntegerLiteral 0x55dd05798610 <col:17> 'int' 1
    |-CXXMemberCallExpr 0x55dd057986c8 <line:20:9, col:16> 'void'
    | |-MemberExpr 0x55dd05798678 <col:9, col:11> '<bound member function type>' .foo 0x55dd05769fd8
    | | `-DeclRefExpr 0x55dd05798658 <col:9> 'A' lvalue Var 0x55dd0576a708 'a' 'A'
    | `-IntegerLiteral 0x55dd057986a8 <col:15> 'int' 2
    |-CallExpr 0x55dd05798798 <line:22:9, col:18> 'void'
    | |-ImplicitCastExpr 0x55dd05798780 <col:9, col:15> 'void (*)(int)' <LValueToRValue>
    | | `-ParenExpr 0x55dd05798740 <col:9, col:15> 'void (*)(int)' lvalue
    | |   `-MemberExpr 0x55dd05798710 <col:10, col:12> 'void (*)(int)' lvalue .bar 0x55dd0576a3d0
    | |     `-DeclRefExpr 0x55dd057986f0 <col:10> 'B' lvalue Var 0x55dd05797f08 'b' 'B'
    | `-IntegerLiteral 0x55dd05798760 <col:17> 'int' 3
    |-CallExpr 0x55dd05798898 <line:23:9, col:19> 'void'
    | |-ImplicitCastExpr 0x55dd05798880 <col:9, col:16> 'void (*)(int)' <FunctionToPointerDecay>
    | | `-ParenExpr 0x55dd05798840 <col:9, col:16> 'void (int)':'void (int)' lvalue
    | |   `-UnaryOperator 0x55dd05798828 <col:10, col:13> 'void (int)':'void (int)' lvalue prefix '*' cannot overflow
    | |     `-ImplicitCastExpr 0x55dd05798810 <col:11, col:13> 'void (*)(int)' <LValueToRValue>
    | |       `-MemberExpr 0x55dd057987e0 <col:11, col:13> 'void (*)(int)' lvalue .bar 0x55dd0576a3d0
    | |         `-DeclRefExpr 0x55dd057987c0 <col:11> 'B' lvalue Var 0x55dd05797f08 'b' 'B'
    | `-IntegerLiteral 0x55dd05798860 <col:18> 'int' 4
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ The following authors have contributed to this project (in alphabetical order):
 * [maximiliankaul](https://github.com/maximiliankaul)
 * [obraunsdorf](https://github.com/obraunsdorf)
 * [oxisto](https://github.com/oxisto)
+* [peckto](https://github.com/peckto)
 * [titze](https://github.com/titze)
 * [vfsrfs](https://github.com/vfsrfs)
 
--- a/cpg-core/src/test/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXAmbiguitiesTest.kt
+++ b/cpg-core/src/test/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXAmbiguitiesTest.kt
@@ -35,6 +35,7 @@ import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration
 import de.fraunhofer.aisec.cpg.graph.statements.DeclarationStatement
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.CallExpression
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.CastExpression
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.MemberCallExpression
 import java.io.File
 import kotlin.test.assertContains
 import kotlin.test.assertEquals
@@ -111,5 +112,29 @@ class CXXAmbiguitiesTest {
 
         val s4 = mainFunc.getBodyStatementAs(4, CastExpression::class.java)
         assertNotNull(s4)
+    }
+
+    /**
+     * In CXX there is an ambiguity with the statement: "(A.B)(C);" 1) If B is a method, this is a
+     * [MemberCallExpression] 2) if B is a function pointer, this is a [CallExpression].
+     *
+     * Function pointer as a struct member are currently not supported in the cpg. This test case
+     * will just ensure that there will be no crash when parsing such a statement. When adding this
+     * functionality in the cpg, this test case must be adapted accordingly.
+     */
+    @Test
+    fun testMethodOrFunction() {
+        val file = File("src/test/resources/method_or_function_call.cpp")
+        val tu = TestUtils.analyzeAndGetFirstTU(listOf(file), file.parentFile.toPath(), true)
+        assertNotNull(tu)
+
+        val mainFunc = tu.byNameOrNull<FunctionDeclaration>("main")
+        assertNotNull(mainFunc)
+
+        val classA = tu.byNameOrNull<RecordDeclaration>("A")
+        assertNotNull(classA)
+
+        val structB = tu.byNameOrNull<RecordDeclaration>("B")
+        assertNotNull(structB)
     }
 }
--- a/cpg-core/src/test/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXAmbiguitiesTest.kt
+++ b/cpg-core/src/test/java/de/fraunhofer/aisec/cpg/frontends/cpp/CXXAmbiguitiesTest.kt
@@ -28,10 +28,13 @@ package de.fraunhofer.aisec.cpg.frontends.cpp
 import de.fraunhofer.aisec.cpg.TestUtils
 import de.fraunhofer.aisec.cpg.graph.bodyOrNull
 import de.fraunhofer.aisec.cpg.graph.byNameOrNull
+import de.fraunhofer.aisec.cpg.graph.declarations.FunctionDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.MethodDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.ProblemDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration
 import de.fraunhofer.aisec.cpg.graph.statements.DeclarationStatement
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.CallExpression
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.CastExpression
 import java.io.File
 import kotlin.test.assertContains
 import kotlin.test.assertEquals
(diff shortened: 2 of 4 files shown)


oxisto:
I think we cannot correctly distinguish between those two because both of them are regular call expressions and the only difference is the way the call "target" is accessed ("directly" vs. a member). The Clang AST allows a call on basically any *expression* (in the second case being a member expression). This is the `callee` property in the Clang AST and it is part of the AST. It then has a `calleeDecl` property (to the underlying *declaration* that is called), which is probably then filled later by the linker, etc.

Where as our graph only has an `INVOKES` edge (which is NOT an AST edge) that goes directly to the *declaration* and is equivalent to the `calleeDecl` property in clang. We do have the `base` property, but that is actually only used in member calls. This has been a mess for quite some time, leading to different handling of calls vs. member calls.

To mitigate this, we could introduce a `callee` property (and potentially then get rid of the `base` property in the future). For normal calls the `callee` would then be `DeclaredReferenceExpression` and for member calls `callee` would be a `MemberExpression`.

peckto:
Did I understand you correctly, that you would suggest making no distinction between a `MemberCallExpressoin` and a `CallExpression` and instead introduce a `CALLEE` edge, which could be a `MemberExpression` or a `DeclaredReferenceExpression`?

This could look like the following:
callee-property

oxisto:
> Did I understand you correctly, that you would suggest making no distinction between a `MemberCallExpressoin` and a `CallExpression` and instead introduce a `CALLEE` edge, which could be a `MemberExpression` or a `DeclaredReferenceExpression`?
> 
> This could look like the following: callee-property

We could have both. The `callee` property would be on a `CallExpression` base class, regardless of which expression is called (reference or member access). If we want to quickly identify member calls, we can still have the derived `MemberCallExpression` (as clang does) which can be used to quickly identify the associated record declaration with a property (which is already there at the moment - and not an AST node).

peckto:
Ok, so we could have a `MemberExpression` as `CALLEE` for a `MemberCallExpression` as well as for a regular `CallExpression`. I think this would be a good solution.

I just noticed, that you fixed the crash in #687.
Introducing the `CALLEE` edge seems like a larger effort.
Should we create a new issue for that?

For me, it would be important to merge the test-cases from this PR, to ensure that we will not break it accidentally.

oxisto:
> Ok, so we could have a `MemberExpression` as `CALLEE` for a `MemberCallExpression` as well as for a regular `CallExpression`. I think this would be a good solution.
> 
> I just noticed, that you fixed the crash in #687. Introducing the `CALLEE` edge seems like a larger effort. Should we create a new issue for that?
> 
> For me, it would be important to merge the test-cases from this PR, to ensure that we will not break it accidentally.

Can you fix the merge-conflicts? Then we can merge the test-cases. You can create a new issue with the callee property.

oxisto (approved):

PR #690, opened 2022-01-31
Adapting visibility modifiers in passes to allow extendabillity

konradweiss:
This PR modifies many of the private functions in the default passes to open them up for modification through subclassing. Members and member functions of the passes are now `protected` instead of `private`. Some functions were `public` and had no purpose in being, therefore, I made them protected as well.


konradweiss:
Already on the Java-doc errors

oxisto (approved):

PR #693, opened 2022-02-02
Extracting python to cpg-language-python module

maximiliankaul:
Open points / questions:
- I removed the whole `experimental-python` stuff (to be in line with the `cpg-language-llvm` module. Correct?
- The `cpg-console` module will fail, if either of the llvm or python modules is missing. Is this intended?
- How to access `PY_EXTENSIONS` field in the `cpg-neo4j` module? Currently, this is hardcoded to `.py`


oxisto:
> Open points / questions:
> 
> * I removed the whole `experimental-python` stuff (to be in line with the `cpg-language-llvm` module. Correct?

I think we can get rid of the experimental tags for the unit tests. However, I would still keep the `@ExperimentalPython` annotation on the language frontend at the moment.

> * The `cpg-console` module will fail, if either of the llvm or python modules is missing. Is this intended?

Depends on your definition of "fail". The cpg-console should depend on all languages, so they need to compile in order for the cog-console to work.

> * How to access `PY_EXTENSIONS` field in the `cpg-neo4j` module? Currently, this is hardcoded to `.py`

You can directly depend on the cpg-language-python module so you can access the class directly. Same as cpg-console.

oxisto (approved):

PR #694, opened 2022-02-02
Extracting Go language frontend into own module

oxisto:
Extracting Go language frontend into own module. This also cleans up the gradle config of different modules a little bit.


oxisto:
Ok I am at a lost why this fails :(

konradweiss (approved):
Looks fine to me

PR #696, opened 2022-02-02
Fixed several Go parsing problems

oxisto:


konradweiss (approved):
Change contain mostly addition of new parsing that looks good to me.

PR #724, opened 2022-02-19
Add option to export benchmark result as json

peckto:
This PR adds an option to export the benchmark result as json.

I'm not happy with the almost identical functions `getBenchmarkResult` and `printBenchmark`.
The only difference is the format of `Translation config`.
Do you have an idea how to unify this?

Benchmark json example:
```js
{
  "Translation config": {
    "inferenceConfiguration": {
      "guessCastExpressions": false,
      "inferRecords": false
    },
    "frontends": {
      "de.fraunhofer.aisec.cpg.frontends.java.JavaLanguageFrontend": [
        ".java"
      ],
      "de.fraunhofer.aisec.cpg.frontends.cpp.CXXLanguageFrontend": [
        ".c",
        ".cpp",
        ".cc",
        ".h",
        ".hpp"
      ],
      "de.fraunhofer.aisec.cpg.frontends.llvm.LLVMIRLanguageFrontend": [
        ".ll"
      ]
    },
    "processAnnotations": false,
    "includePaths": [],
    "codeInNodes": true,
    "failOnError": false,
    "disableCleanup": false,
    "symbols": {},
    "debugParser": true,
    "passes": [
      "de.fraunhofer.aisec.cpg.passes.CompressLLVMPass@3780a43",
      "de.fraunhofer.aisec.cpg.passes.TypeHierarchyResolver@11bdee44",
      "de.fraunhofer.aisec.cpg.passes.JavaExternalTypeHierarchyResolver@3e17b4a7",
      "de.fraunhofer.aisec.cpg.passes.ImportResolver@a1c8540",
      "de.fraunhofer.aisec.cpg.passes.VariableUsageResolver@781e47b2",
      "de.fraunhofer.aisec.cpg.passes.CallResolver@1d6e74c8",
      "de.fraunhofer.aisec.cpg.passes.EvaluationOrderGraphPass@a3ed67d",
      "de.fraunhofer.aisec.cpg.passes.TypeResolver@333122c2",
      "de.fraunhofer.aisec.cpg.passes.ControlFlowSensitiveDFGPass@46c79d59",
      "de.fraunhofer.aisec.cpg.passes.FilenameMapper@c4ee32f"
    ],
    "useUnityBuild": false,
    "topLevel": "/code",
    "loadIncludes": false,
    "includeBlacklist": [],
    "includeWhitelist": [],
    "sourceLocations": [
      "/code/main.c"
    ],
    "typeSystemActiveInFrontend": true,
    "useParallelFrontends": false
  },
  "Number of files translated": 1,
  "Translated file(s)": [
    "file:///code/main.c"
  ],
  "TranslationManager: Executing Language Frontend": "683 ms",
  "CompressLLVMPass: Executing Pass": "0 ms",
  "TypeHierarchyResolver: Executing Pass": "42 ms",
  "JavaExternalTypeHierarchyResolver: Executing Pass": "0 ms",
  "ImportResolver: Executing Pass": "2 ms",
  "VariableUsageResolver: Executing Pass": "26 ms",
  "CallResolver: Executing Pass": "16 ms",
  "EvaluationOrderGraphPass: Executing Pass": "6 ms",
  "TypeResolver: Executing Pass": "6 ms",
  "ControlFlowSensitiveDFGPass: Executing Pass": "10 ms",
  "FilenameMapper: Executing Pass": "2 ms",
  "TranslationManager: Translation into full graph": "808 ms"
}
```

### DIFF
--- a/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
+++ b/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
@@ -557,25 +557,35 @@ public class TranslationConfiguration {
 
   @Override
   public String toString() {
-    return new ToStringBuilder(this, ToStringStyle.JSON_STYLE)
-        .append("debugParser", debugParser)
-        .append("loadIncludes", loadIncludes)
-        .append("includePaths", includePaths)
-        .append("includeWhitelist", includeWhitelist)
-        .append("includeBlacklist", includeBlacklist)
-        .append("frontends", frontends)
-        .append("disableCleanup", disableCleanup)
-        .append("codeInNodes", codeInNodes)
-        .append("processAnnotations", processAnnotations)
-        .append("failOnError", failOnError)
-        .append("symbols", symbols)
-        .append("sourceLocations", sourceLocations)
-        .append("topLevel", topLevel)
-        .append("useUnityBuild", useUnityBuild)
-        .append("useParallelFrontends", useParallelFrontends)
-        .append("typeSystemActiveInFrontend", typeSystemActiveInFrontend)
-        .append("passes", passes)
-        .append("inferenceConfiguration", inferenceConfiguration)
-        .toString();
+    return ToStringBuilder.reflectionToString(this, ToStringStyle.JSON_STYLE);
+  }
+
+  public Map<String, Object> getConfigAsHashMap() {
+    // We first need to convert the passes to string, otherwise the serialisation will fail
+    List<String> passes_names = new ArrayList<String>();
+    for (Pass p : passes) {
+      passes_names.add(p.toString());
+    }
+
+    Map<String, Object> conf = new HashMap<String, Object>();
+    conf.put("debugParser", debugParser);
+    conf.put("loadIncludes", loadIncludes);
+    conf.put("includePaths", includePaths);
+    conf.put("includeWhitelist", includeWhitelist);
+    conf.put("includeBlacklist", includeBlacklist);
+    conf.put("frontends", frontends);
+    conf.put("disableCleanup", disableCleanup);
+    conf.put("codeInNodes", codeInNodes);
+    conf.put("processAnnotations", processAnnotations);
+    conf.put("failOnError", failOnError);
+    conf.put("symbols", symbols);
+    conf.put("sourceLocations", sourceLocations);
+    conf.put("topLevel", topLevel);
+    conf.put("useUnityBuild", useUnityBuild);
+    conf.put("useParallelFrontends", useParallelFrontends);
+    conf.put("typeSystemActiveInFrontend", typeSystemActiveInFrontend);
+    conf.put("passes", passes_names);
+    conf.put("inferenceConfiguration", inferenceConfiguration);
+    return conf;
   }
 }
--- a/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
+++ b/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
@@ -28,6 +28,9 @@ package de.fraunhofer.aisec.cpg;
 import static de.fraunhofer.aisec.cpg.frontends.cpp.CXXLanguageFrontend.CXX_EXTENSIONS;
 import static de.fraunhofer.aisec.cpg.frontends.cpp.CXXLanguageFrontend.CXX_HEADER_EXTENSIONS;
 
+import com.fasterxml.jackson.annotation.JsonIdentityInfo;
+import com.fasterxml.jackson.annotation.JsonIdentityReference;
+import com.fasterxml.jackson.annotation.ObjectIdGenerators;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import de.fraunhofer.aisec.cpg.frontends.CompilationDatabase;
@@ -222,6 +225,8 @@ public class TranslationConfiguration {
     return topLevel;
   }
 
+  @JsonIdentityInfo(generator = ObjectIdGenerators.PropertyGenerator.class, property = "name")
+  @JsonIdentityReference(alwaysAsId = true)
   public List<Pass> getRegisteredPasses() {
     return this.passes;
   }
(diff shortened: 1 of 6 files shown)


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -557,25 +559,47 @@ public TranslationConfiguration build() {
 
   @Override
   public String toString() {
-    return new ToStringBuilder(this, ToStringStyle.JSON_STYLE)
```
oxisto, line 560: I would not do this in `toString`. ToString should be fast and "lean" and definitly not invoke an extra object mapper, etc. I would propose to extract this into a separate function or invoke the jackson functionality in the Benchmark class

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/Benchmark.kt
```diff
@@ -41,6 +41,19 @@ interface StatisticsHolder {
...
+                "Translated file(s)",
+                translatedFiles.map { relativeOrAbsolute(Path.of(it), config.topLevel.toPath()) }
+            ),
+            *benchmarks
+                .map { listOf("${it.caller}: ${it.message}", "${it.duration} ms") }
+                .toTypedArray()
+        )
+    }
     /** Pretty-prints benchmark results for easy copying to GitHub issues. */
     fun printBenchmark() {
```
oxisto, line 58: Ideally, i would suspect that printBenchmark is just a printer and getBenchMarkResults does the actual retrieval of the different metrics. Can you rework it so that printBenchmarks calls getBenchmarkResult and prints the individual items in the list?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
peckto: Apparently we do not even need our own `toString` function.
Or do we need it somewhere else?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/Benchmark.kt
peckto: I refactored it. My only concern was the special handling of the "Translation config". I moved that code part to `printMarkdown`. 
What was the reason for having this special format?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
oxisto: Please keep it :) It is used automatically in a lot of places, for example when using the debugger, etc.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/Benchmark.kt
oxisto: The `toString` method is already printing the translation unit as a JSON. The idea was to have the TU as a json inside the markdown table.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
oxisto: it was also implicitly called by `printBenchmark` to print out the translation config in a more readable format inside the markdown table.

peckto: Sorry, I'm not so familiar to the java world.
If I understood you correctly, it's not recommended to use an object mapper for serialization in the context of the `toString` method.
What would you then suggest to use as `toString` for the `TranslationConfiguration`?

oxisto: The suggested `toString` would be the one that was already implemented :)

peckto: I'm trying to reuse the `getConfigAsJson` inside `toString` to reduce code redundancy.
I tired:
```java
return new ToStringBuilder(this, ToStringStyle.JSON_STYLE).append(getConfigAsJson()).toString();
```
But this fails. 
The following works, but adds the config inside a key:
```java
return new ToStringBuilder(this, ToStringStyle.JSON_STYLE).append("conf", getConfigAsJson()).toString();
```

Can I append a whole Map to the builder? 
Or do I need to iterate all elements and add them separately?

peckto: I think, I finally got it :)

```java
public String toString() {
    return ToStringBuilder.reflectionToString(this, ToStringStyle.JSON_STYLE);
}
```

Is that what you had in mind?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationConfiguration.java
```diff
@@ -557,25 +557,35 @@ public TranslationConfiguration build() {
...
-        .append("typeSystemActiveInFrontend", typeSystemActiveInFrontend)
-        .append("passes", passes)
-        .append("inferenceConfiguration", inferenceConfiguration)
-        .toString();
+    return ToStringBuilder.reflectionToString(this, ToStringStyle.JSON_STYLE);
+  }
+
+  public Map<String, Object> getConfigAsHashMap() {
+    // We first need to convert the passes to string, otherwise the serialisation will fail
+    List<String> passes_names = new ArrayList<String>();
```
oxisto, line 565: Minor thing. This should be `passesNames` or just `names`. No underscores.

oxisto (commented):
Looks good except one minor thing

oxisto:
@peckto I tried to clean up the `StatisticsHolder` interfaces a little bit and tried to introduce a new `BenchmarksResult` class that takes care of the JSON stuff and printing. Please have a look

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/TranslationResult.java
```diff
@@ -135,4 +132,9 @@ public void printBenchmark() {
   public TranslationConfiguration getConfig() {
     return translationManager.getConfig();
   }
+
+  @NotNull
+  public BenchmarkResults getBenchmarkResult() {
```
oxisto, line 137: Inconstistency in naming here. Result or Results?

peckto: Results I would say

peckto:
Thanks for the cleanup.
I just noticed, that the json benchmark results are now represented as tuples and not as a map:
```js
  [
    "Number of files translated",
    1
  ],
  [
    "Translated file(s)",
    [
      "file:///home/tobias/Documents/SourceCodeAudit/code/test2/main.c"
    ]
  ],
  [
    "TranslationManager: Executing Language Frontend",
    "707 ms"
  ],
  [
    "CompressLLVMPass: Executing Pass",
    "0 ms"
  ],
  [
    "TypeHierarchyResolver: Executing Pass",
    "53 ms"
  ],
```
What this change on peruse?
That's also way the tests fail.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/Benchmark.kt
```diff
@@ -94,7 +105,13 @@ fun printMarkdown(table: List<List<Any>>, headers: List<String>) {
 
     for (row in table) {
         var rowIndex = 0
-        val line = row.joinToString(" | ", "| ", " |") { it.toString().padEnd(lengths[rowIndex++]) }
+        val line =
+            row.joinToString(" | ", "| ", " |") {
+                val str =
+                    if (it is Map<*, *>) "`${it.toString().replace(",", ", ").replace(":", ": ")}`"
```
peckto, line 111: This does not work anymore, as we pass the `TranslationConfiguration` as object and not as HashMap. 
Do we need/want this formatting rule? I just copied it from the original code.

It should do something like this:

| key | value
| - | - |
| conf | {"key":"value","foo":"bar"} |

-> 

| key | value
| - | - |
| conf | `{"key": "value", "foo": "bar"}` |

oxisto: Yes, I know. For me this is ok. I would need to implement some logic in the markdown printer that check for a hash map and puts them in backticks. not sure if it is worth the time. Feel free to do so if you want :)

peckto: Ok, then we go without the pretty printing of objects for now.
I converted this into a TODO.

oxisto (approved):

PR #725, opened 2022-02-20
Add support for -isystem and -D

peckto:
This PR extends the `CompilationDatabase` by the following compiler options:
* `-isystem` (system include path)
* `-D` (define symbols)

Additionally, test-cases for the current feature set of the  `CompilationDatabase` are added.

### DIFF
--- a/cpg-core/src/testFixtures/java/de/fraunhofer/aisec/cpg/TestUtils.kt
+++ b/cpg-core/src/testFixtures/java/de/fraunhofer/aisec/cpg/TestUtils.kt
@@ -25,6 +25,7 @@
  */
 package de.fraunhofer.aisec.cpg
 
+import de.fraunhofer.aisec.cpg.frontends.CompilationDatabase
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.TypeManager
 import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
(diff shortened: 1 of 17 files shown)


cpg-core/src/testFixtures/java/de/fraunhofer/aisec/cpg/TestUtils.kt
```diff
@@ -117,9 +117,17 @@ object TestUtils {
         files: List<File>?,
         topLevel: Path,
         usePasses: Boolean,
-        configModifier: Consumer<TranslationConfiguration.Builder>? = null
+        configModifier: Consumer<TranslationConfiguration.Builder>? = null,
+        jsonCompilationDatabase: File? = null
```
oxisto, line 121: I would avoid adding a separate parameter for the jsonCompilationDatabase here, because we will end up with one function for each config options. You should use the `configModifier` for that instead. That allows you to access the config builder in a lambda expression. 

`analyze(files, topLevel, usePasses) { it.useCompilationDatabase(this, that) }`

oxisto (commented):
This somehow seems to brake the go frontend :D Not sure why

cpg-core/src/testFixtures/java/de/fraunhofer/aisec/cpg/TestUtils.kt
peckto: Thanks for the hint!
I fixed it.

peckto:
Tests are fixed now

oxisto (approved):

oxisto:
Does this already include support for system paths? I couldnt find this in the code?

peckto:
With system include paths I refer to the compiler option `-isystem`.
As far as I know, there is no difference between `-I` and `-isystem`.
In this PR, I add both to the list of includes.

PR #726, opened 2022-02-21
Improvements of translating and analyzing LLVM code

KuechA:
* Extend the handling of LLVM code
* Adapt the analysis module to target LLVM samples


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/ValueEvaluator.kt
```diff
@@ -72,237 +70,336 @@ class ValueEvaluator(
...
-                return evaluate(decl.initializer)
-            }
-        }
-
-        return cannotEvaluate(decl, this)
-    }
-
     /** Tries to evaluate this expression. Anything can happen. */
-    fun evaluate(expr: Expression?): Any? {
+    fun evaluate(expr: Node?): Any? {
```
oxisto, line 74: The variable should probably not be named `expr` then.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/ValueEvaluator.kt
```diff
@@ -72,237 +70,336 @@ class ValueEvaluator(
...
-            is VariableDeclaration -> return evaluate(decl.initializer)
-            is FieldDeclaration -> {
-                return evaluate(decl.initializer)
-            }
-        }
-
-        return cannotEvaluate(decl, this)
-    }
-
     /** Tries to evaluate this expression. Anything can happen. */
```
oxisto, line 73:
```suggestion
    /** Tries to evaluate this node. Anything can happen. */
```

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/ValueEvaluator.kt
```diff
@@ -72,237 +70,336 @@ class ValueEvaluator(
...
+        } else if (lhsValue is Double && rhsValue is Number) {
+            lhsValue == rhsValue.toDouble()
+        } else if (lhsValue is Float && rhsValue is Number) {
+            lhsValue == rhsValue.toFloat()
+        } else {
+            cannotEvaluate(expr, this)
+        }
+    }
+
+    /** We handle some basic unary operators. These also affect pointers and dereferences for C. */
```
oxisto, line 296:
```suggestion
    /** We handle some basic unary operators. These also affect pointers and dereferences for languages that support them. */
```

oxisto (commented):
Minor java doc and naming issues

oxisto:
> Looks fine to me, i like the improvement through modularization in the ValueEvaluator. The if-else branches still look like big code blocks that could be simplified but i did not come up with a better solution on the spot.

I have replaced them with `when` statements, which is slightly better.

PR #728, opened 2022-02-22
Add this field to cpp struct as they are classes

peckto:
This PR adds the `this` field to `RecordDeclarations` with type struct, as in cpp structs are classes.
Because we currently make no difference between a c struct and a cpp struct/class, this does also apply to c structs.
This is nothing new, as we create an implicit constructor for structs already.

Additionally, test-cases are added.

### DIFF
--- a/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.kt
+++ b/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.kt
@@ -368,7 +368,8 @@ class DeclaratorHandler(lang: CXXLanguageFrontend) :
                 lang.scopeManager.currentNamePrefixWithDelimiter + ctx.name.toString(),
                 kind,
                 ctx.rawSignature,
-                true
+                true,
+                lang
             )
         recordDeclaration.superClasses =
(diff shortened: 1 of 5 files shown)


oxisto:
To be honest, I would avoid doing this. Adding the `this` field to a record declaration is a major hack, which in my personal opinion should be removed and not extended.

peckto:
The reason why I added the feature is, because we currently have a problem with the following c++ code:
```c++
struct A {
	int a = 1;
	int foo() {
		return this->a;
	}
};
```

This fails when resolving `this->a` inside `handleFieldReference`  due to the missing `this` field.

Do you have an idea how to solve this in a more general way?
If this is a larger rework, I would prefer to apply this PR as a hot-fix.

oxisto:
> The reason why I added the feature is, because we currently have a problem with the following c++ code:
> 
> ```c++
> struct A {
> 	int a = 1;
> 	int foo() {
> 		return this->a;
> 	}
> };
> ```
> 
> This fails when resolving `this->a` inside `handleFieldReference` due to the missing `this` field.
> 
> Do you have an idea how to solve this in a more general way? If this is a larger rework, I would prefer to apply this PR as a hot-fix.

Ok. then let's hot fix it then.

oxisto (approved):
Reluctantly agreeing to it ;)

PR #741, opened 2022-03-09
CPG-Analysis: Fix evaluation of values with different types

### LINKED ISSUE #729: Incorrect type handling in `ValueEvaluator`
The `ValueEvaluator` assumes that the rhs of a binary operator has the same type as the lhs and thus may round the value incorrectly. This can result in wrong evaluations, e.g., `3 < 3.1` is evaluated to false even though it is true.

KuechA:
Fixes #729


oxisto (approved):
I have extracted the number magic into an extension function, otherwise LGTM

Issue #742, opened 2022-03-10
StackOverflow for some PointerTypes


KuechA:
Some PointerTypes seem to lead to stack overflows.

Stacktrace:

```
java.lang.StackOverflowError
	at java.base/java.util.Arrays.hashCode(Arrays.java:4685)
	at java.base/java.util.Objects.hash(Objects.java:146)
	at de.fraunhofer.aisec.cpg.graph.types.Type$Qualifier.hashCode(Type.java:230)
	at java.base/java.util.Arrays.hashCode(Arrays.java:4685)
	at java.base/java.util.Objects.hash(Objects.java:146)
	at de.fraunhofer.aisec.cpg.graph.types.Type.hashCode(Type.java:357)
	at de.fraunhofer.aisec.cpg.graph.types.PointerType.hashCode(PointerType.java:159)
	at java.base/java.util.Arrays.hashCode(Arrays.java:4685)
	at java.base/java.util.Objects.hash(Objects.java:146)
	at de.fraunhofer.aisec.cpg.graph.types.PointerType.hashCode(PointerType.java:159)
	at java.base/java.util.Arrays.hashCode(Arrays.java:4685)
	at java.base/java.util.Objects.hash(Objects.java:146)
	at de.fraunhofer.aisec.cpg.graph.types.PointerType.hashCode(PointerType.java:159)
	at java.base/java.util.Arrays.hashCode(Arrays.java:4685)
	at java.base/java.util.Objects.hash(Objects.java:146)
	at de.fraunhofer.aisec.cpg.graph.types.PointerType.hashCode(PointerType.java:159)
	at java.base/java.util.Arrays.hashCode(Arrays.java:4685)
	at java.base/java.util.Objects.hash(Objects.java:146)
	at de.fraunhofer.aisec.cpg.graph.types.PointerType.hashCode(PointerType.java:159)
	at java.base/java.util.Arrays.hashCode(Arrays.java:4685)
	at java.base/java.util.Objects.hash(Objects.java:146)
	at de.fraunhofer.aisec.cpg.graph.types.PointerType.hashCode(PointerType.java:159)
	at java.base/java.util.Arrays.hashCode(Arrays.java:4685)
	at java.base/java.util.Objects.hash(Objects.java:146)
	at de.fraunhofer.aisec.cpg.graph.types.PointerType.hashCode(PointerType.java:159)
	at java.base/java.util.Arrays.hashCode(Arrays.java:4685)
[...]

PR #752, opened 2022-04-04
Fixed stack overflow errors in type propagation

KuechA:
This PR tries to optimise the (confusing) way we do type propagation by improving the loop detection.


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/HasType.java
```diff
@@ -55,27 +56,27 @@ default void setType(Type type) {
    * Set the node's type. This may start a chain of type listener notifications
    *
    * @param type new type
-   * @param root The node that initiated the type change chain. When a node receives a type setting
-   *     command where root == this, we know that we have a type listener circle and can abort. If
-   *     root == null, the type change is seen as an externally triggered event and subsequent type
-   *     listeners receive the current node as their root.
+   * @param root The nodes which we have seen in the type change chain. When a node receives a type
+   *     setting command where root.contains(this), we know that we have a type listener circle and
+   *     can abort. If root == null, the type change is seen as an externally triggered event and
```
oxisto, line 61: Is `root == null` still correct or is it rather an empty list?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/HasType.java
```diff
@@ -55,27 +56,27 @@ default void setType(Type type) {
...
+   *     subsequent type listeners receive the current node as their root.
    */
-  void setType(Type type, HasType root);
+  void setType(Type type, Collection<HasType> root);
 
   Set<Type> getPossibleSubTypes();
 
   default void setPossibleSubTypes(Set<Type> possibleSubTypes) {
-    setPossibleSubTypes(possibleSubTypes, null);
+    setPossibleSubTypes(possibleSubTypes, new ArrayList<>());
```
oxisto, line 69: aren't you checking for null in the override functions anyway, so maybe this new list is unnecessary. Otherwise, it would be good to a a non null annotation to `root`.

oxisto (commented):
Some minor things.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/HasType.java
KuechA: Added `@NonNull`

KuechA:
Closes #742

PR #758, opened 2022-04-07
Add intermediate node to represent a software component

### LINKED ISSUE #341: Implement a way to provide an application boundary between translation units
Provide a node for a collection of translation units. It would be nice to have a concept like a "module" or "application" or "program", i.e. a list of translation units that belong together.

KuechA:
Introduce a new node "SoftwareComponent" which represents a piece of software. The CPG should be configurable in a way to support more than one such component and potentially having relationships between these components which are not simple method calls. Discussion of features etc. ongoing in #753.

Closes #341


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/SoftwareComponent.kt
```diff
@@ -0,0 +1,45 @@
...
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph
+
+import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
+import java.util.*
+
+class SoftwareComponent : Node() {
```
oxisto, line 31: Missing javadoc for class

oxisto (commented):
Looks good. Seems to be mostly backwards compatible. We should probably release a new minor version after this.

oxisto:
Quick idea: Maybe just `Component` instead of `SoftwareComponent`? I think its pretty clear that it is software. Might be a little bit shorter.

oxisto:
Quick idea: Maybe just `Component` instead of `SoftwareComponent`? I think its pretty clear that it is software. Might be a little bit shorter.

PR #761, opened 2022-04-13
Reduce size of `processedMap` in LanguageFrontend and reuse it to resolve LLVM goto statements.

### LINKED ISSUE #740: Only store "interesting" nodes in the processedMapping property of a language fronten
It only seems to be used to resolve labels of a Goto statement. A pass or a language frontend should somehow state which node types are "interesting" only those node types should be stored in this processedMapping map.

Furthermore, we have a similar functionality in the LLVM language frontend and we should check, whether this could also use the existing processedMapping.

KuechA:
The Map `processedMap` in the LanguageFrontend is (currently?) only used to resolve targets of goto statements but stores all nodes. To keep it somewhat smaller, we only put GotoStatements and LabelStatements to the map.

We reuse this map to resolve goto targets in the LLVM frontend instead of keeping the LabelStatements in an own map.

Closes #740


oxisto (approved):
Good idea! Looks good.

PR #762, opened 2022-04-19
Extracting Java language frontend

oxisto:
This PR tries to extract the Java language frontend into a seperate module, similar to the LLVM, Python and Go frontends. Several things need to be re-worked, since the Java frontend was the very first and is extremly tied into the cpg-core package.

- [x] Provide a possibility for language frontend to register additional default passes (e.g. hierarchy resolver pass)
- [ ] Provide a way for langauge frontends to register themselves as default langauges
- [ ] Provide a better way to "load" language frontends and list existing ones, especially for the type manager, e.g. using reflection at startup
- [ ] Remove Java-specific functionality in the variable usage resolver and the call resolver or better provide a way for language frontends to "hook" into the resolution
- [ ] Rework DFG, EOG and possibly other tests, that "assume" that the Java frontend is there. Either move the Java-specific part to the Java module or rewrite the tests to work on pre-defined anguage-agnostic CPG nodes to test the actual behaviour
- [ ] Convert the rest of the Java code to Kotlin
- [ ] #763, because of some issues that arose when extracting

Probably lots more to do


oxisto:
Will try again in the future, too many things have changed in the mean time for this PR to work.

Issue #704, opened 2022-02-10
Next-Generation Call Resolving


oxisto:
## Concepts

We will introduce some new concepts in call resolving

### `CastResult` and `Language.tryCast`

`CastResult` will be the result of the function `Language.tryCast` and describes whether a cast of one type into another is successful according to the logic of the `Language`. A default implementation is provided. It also has a property that describes the "depth" of the cast needed, which is the distance in the ancestry between base and derived class.

It is a sealed class that has the following sub-classes/objects:
* `CastNotPossible`, which says that the two types are incompatible and cannot be cast into each other. It has a depth of -1.
* `DirectMatchNoCastNeeded`, which says that the types directly match with a depth of 0. This is mostly for languages that have no implicit casts and directly match types to interfaces (such as Go). Also for types that are basically the same or just an alias of each other.
* `ImplicitCast` (as a companion object) with a depth of 1 specifies that one (implicit) cast is needed to convert between the types. This is used in most languages to cast between numeric types.
* `ImplicitCast ` (as a class) with a configurable depth means that an (implicit) cast is needed between object types and depth specifies the distance between the base and derived type. We probably should rename this to something else.

### `SignatureResult` and `FunctionDeclaration.matchesSignature`

`SignatureResult` will be the result of the function `FunctionDeclaration.matchesSignature` which calculates whether the provided `CallExpression` will match the signature of the current `FunctionDeclaration`. It can be configured to take into account variadic and default parameters (depending on the `Language`).

It is a sealed class that has the following sub-classes/objects:
* `IncompatibleSignature` which specifies that the signatures do not match even after implicit casts
* `SignatureMatches` specifies that the signature matches, although implicit casts might be necessary, it also has a `casts` parameter which contains a list of `CastResult`, one for each parameter. This can then later be used to compare different signature match results in order to decide which function is suitable best.

### `CallResolutionResult` and `ScopeManager.resolveCall`

`CallResolutionResult` is the result of `ScopeManager.resolveCall`. It holds all necessary intermediate results as well as the final result of the call resolution. Among the results we want to store:
* a reference to the `CallExpression` we are resolving
* a list of `candidateFunctions`, which are the starting point of the resolution and contain all symbols with the desired name of the call in scope
* a set of `viableFunctions` that restrict the candidate functions to functions whose signature match
* the `signatureResults`, which is a map of function declarations and their `SignatureResult`
[...]

PR #763, opened 2022-04-19
Improving CXX method resolution

### LINKED ISSUE #650: CallExpression is connected to wrong MethodDeclaration
In case the implementation of a method is not given, but a method with same name and same signature (but from a different class) exists, the `CallExpression` is connected to the wrong `MethodDeclaration` (via `INVOKES` edge).

Example:
```c++
class A {
public:
        void foo(int a);
};

class B {
public:
        void foo(int a);
};

void B::foo(int a) {
}

int main() {
        A a;
        B b;

        a.foo(1);
        b.foo(2);

        foo();

        C c;
        c.foo(3);
}
```

When loading the example c++ code into the cpg, all three calls of a method `foo` (`a.foo`, `b.foo`, `c.foo`) are connected to the `MethodDeclaration` of `B::foo`.

cpg_bug_CallExpression

oxisto:
This PR is a little bit of a hack that improves CXX method resolution for now, until we redesign the call resolver (see #704)

Fixes #650


peckto:
Thanks a lot for this fix!
It works quite well. I just noticed a problem when resolving a method call, from an unknown class (eg. `c.foo(4)`).
In case a function with same name does exist, it is connected to this function, else no `INVOKES` edge is generated at all.

I added a test-case for the method resolution.

oxisto:
> Thanks a lot for this fix! It works quite well. I just noticed a problem when resolving a method call, from an unknown class (eg. `c.foo(4)`). In case a function with same name does exist, it is connected to this function, else no `INVOKES` edge is generated at all.
> 
> I added a test-case for the method resolution.

I actually already had a test ready and just forgot to git add it. I have done this now, sorry for the override ;)

With regards to your "problem". This is intended. For your use case to produce an INVOKES edge to an inferred function of an inferred class, you need to modify the `InferenceConfiguration`. The default does not infer record declarations. The test shows how to enable this.

peckto:
> With regards to your "problem". This is intended. For your use case to produce an INVOKES edge to an inferred function of an inferred class, you need to modify the `InferenceConfiguration`. The default does not infer record declarations. The test shows how to enable this.

Ok, that I didn't know.
But I'm still not sure about the case, when there is a `FunctionDeclaration` for `foo(int)` but not for `c.foo(int)`.
Regardless of the `inferenceConfiguration`, `c.foo(int)` is connected to `foo(int)`:
```cpp
void foo (int a) { }

int main() {
        foo(3);

        C c;
        c.foo(4);
}
```

image

oxisto:
> > With regards to your "problem". This is intended. For your use case to produce an INVOKES edge to an inferred function of an inferred class, you need to modify the `InferenceConfiguration`. The default does not infer record declarations. The test shows how to enable this.
> 
> Ok, that I didn't know. But I'm still not sure about the case, when there is a `FunctionDeclaration` for `foo(int)` but not for `c.foo(int)`. Regardless of the `inferenceConfiguration`, `c.foo(int)` is connected to `foo(int)`:
> 
> ```c++
> void foo (int a) { }
> 
> int main() {
>         foo(3);
> 
>         C c;
>         c.foo(4);
> }
> ```
> 
> image

Ah, that should not be, I will check that.

peckto:
Thanks for the fix!
Solves the issue for now.

PR #765, opened 2022-04-20
First attempt to implement bug reasoning logic for Query API

KuechA:
We probably want to simplify the identification of bugs by providing an API to query the graph for vulnerabilities.

Tasks:
* [x] Evaluate the different components of a query (probably something like first order logic)
* [x] Parse the queries
* [ ] Add required analyses to the `cpg-analysis` module

See WIP wiki page concerning the [Query API](https://github.com/Fraunhofer-AISEC/cpg/wiki/Query-API).


oxisto:
Just a suggestion: Wouldn't it be easier to use Kotlin itself as the script language rather than try to parse specific string-based fields and identifiers? This way we could probably re-use all the existing node types and their property and not build an additional property layer on top of it.

For example https://medium.com/adobetech/building-elegant-dsls-with-kotlin-707726c5ed21

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/statements/expressions/BinaryOperator.java
```diff
@@ -245,4 +242,27 @@ public boolean equals(Object o) {
...
+  @Nullable
+  @Override
+  public Expression getValue() {
+    return isAssignment() ? rhs : null;
+  }
+
+  public boolean isAssignment() {
+    // TODO(oxisto): We need to discuss, if the other operators are also assignments and if we
+    // really want them
+    return this.operatorCode.equals(
```
KuechA, line 264: Why is `==` an assignment? Isn't that pretty much the only operator which is not an assignment? I guess you meant `=` or maybe wanted to negate the condition?

oxisto: Typo, should be `= `

cpg-console/src/main/kotlin/de/fraunhofer/aisec/cpg/query/Query.kt
```diff
@@ -0,0 +1,349 @@
+/*
```
oxisto, line 1: Shouldn't we actually move this to `cpg-analysis`?

cpg-console/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
```diff
@@ -0,0 +1,298 @@
...
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.query
+
+import de.fraunhofer.aisec.cpg.graph.compareTo
+
+/**
+ * Holds the [value] to which the statements have been evaluated. The [children] define previous
```
oxisto, line 31: a short explainatation what this actually is would be good/helpful.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -136,6 +136,104 @@ class StatementNotFound : Exception()
 
 class DeclarationNotFound(message: String) : Exception(message)
 
+fun Node.followPrevDFGEdgesUntilHit(predicate: (Node) -> Boolean): List<Node> {
```
oxisto, line 139: missing doc

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -136,6 +136,104 @@ class StatementNotFound : Exception()
...
+    return result
+}
+
+/**
+ * Returns a list of nodes which are evaluation paths between the starting node [this] and the end
+ * node fulfilling [predicate]. Paths which do not end at such a node are not included in the
+ * result. Hence, if the return value is a non-empty list, a path from [this] to such a node is
+ * **possible but not mandatory**.
+ */
+fun Node.followNextEOGEdgesUntilHit(predicate: (Node) -> Boolean): List<List<Node>> {
```
oxisto, line 205: What's the particular difference between the "new" (until hit) and "old" functions without the suffix?

cpg-console/src/main/kotlin/de/fraunhofer/aisec/cpg/query/Query.kt
```diff
@@ -0,0 +1,349 @@
...
+ * The optional argument [sel] can be used to filter nodes for which the condition has to be
+ * fulfilled. This filter should be rather simple in most cases since its evaluation is not part of
+ * the resulting reasoning chain.
+ *
+ * This method can be used similar to the logical implication to test "sel => mustSatisfy".
+ */
+@ExperimentalGraph
+@OptIn(ExperimentalTypeInference::class)
+@OverloadResolutionByLambdaReturnType
+inline fun <reified T> Node.all(
```
oxisto, line 50: I did some additional research The problem is that this function actually has two lambda functions and only the second one is the discerning factor in chosen the final override. This prompts the very ugly warning in the IDE/compiler. My suggestion therefore would be to actually split this into two functions `all` which takes the `Boolean` as the "short" version and `allExtended` (or something better named) which takes the query tree variant.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: The main difference is that the "new" functions return all paths satisfying `predicate` whereas the "old" functions return only one of them.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/Query.kt
```diff
@@ -0,0 +1,423 @@
...
+ * Evaluates if the conditions specified in [mustSatisfy] hold for all nodes in the graph.
+ *
+ * The optional argument [sel] can be used to filter nodes for which the condition has to be
+ * fulfilled. This filter should be rather simple in most cases since its evaluation is not part of
+ * the resulting reasoning chain.
+ *
+ * This method can be used similar to the logical implication to test "sel => mustSatisfy".
+ */
+@ExperimentalGraph
+inline fun <reified T> Node.forallExtended(
```
oxisto, line 50: Since you renamed the "old" all, I think we can use `all` for this instead of `forall`.

oxisto (approved):
I propose merging an initial version of this so we can tweak it further in smaller PRs

PR #772, opened 2022-04-29
Fluent Node DSL

oxisto:
This PR adds a fluent node DSL to quickly build up a tree of CPG nodes. This can be used for unit tests or any occasion you quickly want to spin up a tree of CPG nodes that are independent of the underlying programming language. This is useful if you want to test the behaviour of some passes purely on the CPG tree, rather than a concrete program. It purely creates an AST using the CPG nodes, and does *not* fill any of the properties that are populated by passes, such as `INVOKES` or `REFERS_TO`, basically emulating what a `LanguageFrontend` does.

It also introduces a new interface called `ArgumentHolder`, similar to the existing `StatementHolder` and `DeclarationHolder` concepts.

A simple example can be found in the `DslTest` class but looks similar to the following:
```Kotlin
with(TestLanguageFrontend()) {
    translationUnit("file.cpp") {
        function("main") {
            body {
                declare { variable("a") { literal(1) } }
                returnStmt { ref("a") + literal(2) }
            }
        }
    }
}
```

### DIFF
--- a/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/ArgumentHolder.kt
+++ b/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/ArgumentHolder.kt
@@ -25,69 +25,13 @@
  */
 package de.fraunhofer.aisec.cpg.graph
 
-import de.fraunhofer.aisec.cpg.graph.statements.expressions.BinaryOperator
-import de.fraunhofer.aisec.cpg.graph.statements.expressions.DeclaredReferenceExpression
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
-import de.fraunhofer.aisec.cpg.graph.statements.expressions.Literal
 
 interface ArgumentHolder {
-    /**
-     * Creates a new binary operator with operator code + out of two expressions and assigns them to
-     * this expression as an argument.
-     */
-    operator fun Expression.plus(rhs: Expression): BinaryOperator {
-        return binOp("+") {
-            this.lhs = this@plus
-            this.rhs = rhs
-        }
-    }
-
-    /**
-     * Creates a new binary operator with operator code - out of two expressions and assigns them to
-     * this expression as an argument.
-     */
-    operator fun Expression.minus(rhs: Expression): BinaryOperator {
-        return binOp("-") {
-            this.lhs = this@minus
-            this.rhs = rhs
-        }
-    }
 
     fun addArgument(expression: Expression)
-}
 
-/** Creates a new reference and assigns it to the [ArgumentHolder]. */
-fun ArgumentHolder.ref(
-    to: String,
-    init: (DeclaredReferenceExpression.() -> Unit)? = null
-): DeclaredReferenceExpression {
-    val node = new(to, parent = this as? Node, init = init)
-
-    // this.addArgument(node)
-
-    return node
-}
-
-/** Creates a new literal and assigns it to the [ArgumentHolder]. */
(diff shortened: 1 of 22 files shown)


oxisto:
Ready for initial review. I would add more nodes as we need them or are we missing essentials for the test in core, @KuechA?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/ArgumentHolder.kt
```diff
@@ -0,0 +1,37 @@
...
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph
+
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
+
+interface ArgumentHolder : Holder<Expression> {
```
KuechA, line 30: Why do we need this interface? => Documentation

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/Holder.kt
```diff
@@ -0,0 +1,30 @@
...
+ *                   $$ |      $$$$$$$  |$$ |$$$$\
+ *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph
+
+interface Holder<N : Node> {
```
KuechA, line 28: doc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
```diff
@@ -0,0 +1,385 @@
...
+}
+
+/**
+ * Creates a new [CompoundStatement] in the Fluent Node DSL and sets it to the
+ * [IfStatement.elseStatement] of the nearest enclosing [IfStatement]. The [init] block can be used
+ * to create further sub-nodes as well as configuring the created node itself.
+ */
+context(IfStatement)
+
+fun LanguageFrontend.els(
```
KuechA, line 292: The name `els` is kind of weird but I'm not sure what would be better. `elseStmt` is more long but would probably be less confusing

KuechA (approved):
looks good to me. It seems to cover the most relevant concepts already. But thinking about the tests in core, I guess there are quite some tests looking at symbol resolution, so having a way to specify `RecordDeclaration`s and fields and implementing a basic type hierarchy there will probably be necessary soon

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
oxisto: Should I also rename `then` to `thenStmt`?

KuechA: For consistency reasons it could make sense but to me, it would be ok like this as well

oxisto: Done

PR #775, opened 2022-05-11
Performance tests and improvements: LLVM and Type updates

KuechA:
* Test the performance of larger LLVM files
* Improve performance when handling phi: Use SubgraphWalker less frequently
* Improve performance of type updates: Use lists instead of sets


oxisto (approved):

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
```diff
@@ -336,13 +336,16 @@ public static void setTypeSystemActive(boolean active) {
...
-  public Map<HasType, Set<Type>> getTypeCache() {
+  public Map<HasType, List<Type>> getTypeCache() {
     return typeCache;
   }
 
   public synchronized void cacheType(HasType node, Type type) {
     if (!isUnknown(type)) {
-      typeCache.computeIfAbsent(node, n -> new HashSet<>()).add(type);
+      List<Type> types = typeCache.computeIfAbsent(node, n -> new ArrayList<>());
+      if (!types.contains(type)) {
```
oxisto, line 346: Isn't this essentially the same as a set does? Not sure this is really faster / better here.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/declarations/ValueDeclaration.java
```diff
@@ -83,7 +83,7 @@ public Type getPropagationType() {
   }
 
   @Override
-  public void setType(Type type, Collection<HasType> root) {
+  public void setType(Type type, List<HasType> root) {
```
oxisto, line 86: Same here. Keep signature and only change the implememation?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/declarations/FieldDeclaration.java
```diff
@@ -165,12 +165,11 @@ public void typeChanged(HasType src, Collection<HasType> root, Type oldType) {
   }
 
   @Override
-  public void possibleSubTypesChanged(
-      HasType src, Collection<HasType> root, Set<Type> oldSubTypes) {
+  public void possibleSubTypesChanged(HasType src, List<HasType> root) {
```
oxisto, line 168: Here as well I guess

PR #780, opened 2022-05-19
Parses single line comments

konradweiss:
This PR:

- Extracts single line comments with the python tokenizer
- Matches them to the closest node according to a heuristic 
- Fixes Region computation to match sarif region formats
- Adds region computation to implicitly added `CompoundStatements` and `ParamVariableDeclaration`


maximiliankaul:
Hi @konradweiss
I tried your PR with the following code snippet and the results did not look very convincing to me:

```python
# A comment

i = 42  # a number

# a function
def foo(): # foo
    pass


def bar(i,  # a parameter
        j  # another parameter
        ):
    pass


a = ('start'  # comment start
     'end')  # comment end

b = {
    "a": 1,  # a entry
    "b": 2  # b entry
}
```
I can only find a single comment (that happens to be the entire file) when exploring the above code with the neo4j query `MATCH (n) WHERE n.comment IS NOT NULL RETURN n`.

I would also like to keep this feature optional, as it currently requires to parse the code multiple times (however, I do not see an easy solution with Python's `ast` module for adding this functionality).

Currently, I would prefer to not merge this PR. I think this functionality should be in an optional `Pass` or something similar. There is also a discussion why `ast` is not carrying comments here: https://bugs.python.org/issue24119

maximiliankaul:
String comments are currently available in the Python front end (as string literals). Would it not be cleaner to make use of them in a `Pass` and attach them with some heuristic to a node?
```python
""" a comment """
def foo():
  """ more text """
  pass
```
(This would work for @immqu)

konradweiss:
@maximiliankaul Finished all the requested changes including separating the comment matching code from the frontend, and the comment extraction into a new file. The feature is also optional, default off, and configurable over the translation configuration. If its OK I would merge this now

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/CommentMatcher.kt
```diff
@@ -0,0 +1,134 @@
...
+ *
+ */
+package de.fraunhofer.aisec.cpg.helpers
+
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.declarations.NamespaceDeclaration
+import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
+import de.fraunhofer.aisec.cpg.sarif.Region
+
+class CommentMatcher {
```
oxisto, line 33: Missing doc for class

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/CommentMatcher.kt
```diff
@@ -0,0 +1,134 @@
...
+    fun getEnclosingChild(node: Node, location: Region): Node {
```
oxisto, line 34: Missing doc for function

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/CommentMatcher.kt
```diff
@@ -0,0 +1,134 @@
...
+        var children = SubgraphWalker.getAstChildren(node)
+        children.addAll(
+            children.filterIsInstance<NamespaceDeclaration>().flatMap {
```
oxisto, line 37: Please provide a short explanation what the reason for these 3 lines are, as they are not self-explanatory.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
```diff
@@ -105,6 +105,13 @@ class PythonLanguageFrontend(config: TranslationConfiguration, scopeManager: Sco
 
             // run python function parse_code()
             tu = interp.invoke("parse_code", code, path, this) as TranslationUnitDeclaration
+
+            if (config.matchCommentsToNodes) {
+                // Parse comments and attach to nodes
+                val pyOptional = javaClass.getResource("/CPGPython/_comment_parsing.py")
+                interp.exec(pyOptional?.readText())
```
oxisto, line 112: shouldn't the whole exec be wrap in a `?.let` or `if != null`, otherwise, you will execute an empty string

oxisto:
> @maximiliankaul Finished all the requested changes including separating the comment matching code from the frontend, and the comment extraction into a new file. The feature is also optional, default off, and configurable over the translation configuration. If its OK I would merge this now

An acutual unit test for the comment matched would be nice as well.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/CommentMatcher.kt
konradweiss: Added small explanation on the purpose of the class

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/CommentMatcher.kt
konradweiss: Added requested doc

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/CommentMatcher.kt
konradweiss: Added explanation

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/CommentMatcher.kt
```diff
@@ -0,0 +1,134 @@
...
+                        { it.location?.region?.startLine ?: 0 },
+                        { it.location?.region?.startColumn ?: 0 }
+                    )
+                )
+                .firstOrNull()
+        val closestLine = closest?.location?.region?.startLine ?: location.endLine + 1
+
+        // If the closest successor is not in the same line there may be a more adequate predecessor
+        // to associated the
+        // comment to (Has to be in the same line)
```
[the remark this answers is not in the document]
konradweiss: Formatting corrected

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
konradweiss: I made this conditional, the resource should not be missing, however, now it is safer. If the resource is missing comments are not extracted nor matched

konradweiss:
@oxisto @maximiliankaul I made the requested changes including adding a Unit test, using the error uncovering sample provided by @maximiliankaul.

maximiliankaul (approved):
Looks good to me :)

PR #784, opened 2022-05-25
Fixed missing DFG edges in Go frontend

oxisto:
Currently, the declaration of a variable and the expression that represents an object literal, e.g., `d := &Data{}` were not connected with a DFG edge. This was due to the fact that there is not real object literal expression in the CPG and we need to "emulate" it with a combination of a `ConstructExpression` and a `InitializerListExpression`.

Furthermore, a DFG edge was missing in `InitializerListExpression::addInitializer`, it was only set in `setInitializers`.


immqu:
> @immqu please have a look at this.

I've tested it, looks good!

PR #785, opened 2022-05-27
Making `SubgraphWalker::flattenAST` insanely fast

oxisto:
This PR adds improvements to the SubgraphWalker, cutting its execution time by more than 95 % (on my machine). Executing the `WalkerTest`, which traverses 82619 nodes, takes around 8-9s on `master`,  and now takes around 0.4s on my MacBook Pro M1. On GitHub the improvement is roughly 20s instead of 50s.

It does so by introducing an `IdentitySet`, which is a special implementation of a set, that uses reference-equality instead of object-equality, meaning that only nodes that are not the *same* node are stored in the set, instead of checking the node via `hashCode` or `equals`. This should be enough to avoid potential loops in the CPG that could be introduced by fault language frontends, which accidentally set references to nodes wrong.

There are potentially a lot of other places where we currently use a `HashSet` to keep a work-list that could benefit from this list.


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/SubgraphWalker.java
```diff
@@ -177,9 +177,7 @@ private static void flattenASTInternal(@NonNull Set<Node> list, @NonNull Node n)
     list.add(n);
 
     for (Node child : SubgraphWalker.getAstChildren(n)) {
-      if (!list.contains(child)) {
-        flattenASTInternal(list, child);
-      }
+      flattenASTInternal(list, child);
```
KuechA, line 180: The `list` has been used to detect loops in the AST which could originate from errors during translation or similar. With those changes, the argument `list` is no longer used for this purpose and I'm wondering if this could be an issue. Instead of the contains, we could also return if the `add` operation returns `false`.

Also, `list` is not a good name for the argument because it's actually a set, probably `result` or similar makes more sense in the current usage.

KuechA (changes requested):

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/SubgraphWalker.java
KuechA: The check `if (!list.contains(child))` was meant to avoid endless recursion. I don't see how this is ensured now. Of course, the set only contains each element once but we may try to add it multiple times, or not? I'd suggest to fix this by changing line 177 from `list.add(n)` to `if(!list.add(n)) return`. Or maybe I got something wrong?

oxisto: Ah yes, you are right. I was only adding the loop detection to the AST visitor (see VisitorTest), but not to the flattenAST.

KuechA (approved):

PR #835, opened 2022-06-15
Removing `this` field and moving all languages into the `receiver` logic

oxisto:
This PR changes all language frontends into the `receiver` logic, instead of creating a `this` field. The `receiver` field of a `MethodDeclaration` contains a more detailed explanation, especially for frontend developers how to create it.

One notable exception is the usage of inner classes in Java. In this case, a `this` referring to the outer class is created, e.g. `OuterClass.this`. This is similar to what the Java compiler does in the byte code.


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclaratorHandler.kt
```diff
@@ -166,13 +168,16 @@ class DeclaratorHandler(lang: CXXLanguageFrontend) :
...
         ctx: IASTNode,
     ): MethodDeclaration {
-        // check, if its a constructor
-        return if (name == recordDeclaration?.name) {
-            NodeBuilder.newConstructorDeclaration(name, null, recordDeclaration, lang, ctx)
-        } else NodeBuilder.newMethodDeclaration(name, null, false, recordDeclaration, lang, ctx)
+        // Check, if it's a constructor
+        val method =
+            if (name == recordDeclaration?.name) {
+                NodeBuilder.newConstructorDeclaration(name, null, recordDeclaration, lang, ctx)
```
KuechA, line 174: Remove `NodeBuilder` consistently (as in line 105) and use `{}` for the else branch too?

oxisto: Done

PR #869, opened 2022-08-10
Implement ordering for passes

maximiliankaul:
This PR adds the following annotations for passes/frontends:
- `ExecuteFirst` a pass marked with this annotation has to be executed as the very first pass.
- `ExecuteLast` a pass marked with this annotation has to be executed as the very last pass.
- `RegisterDependency` the other pass will be executed before the current pass. The optional parameter `softDependency` controls whether a missing dependency will be added to the list of active passes if it is missing.
- `RegisterExtraPass` can be added to frontends to register frontend specific passes.
- `RequiredFrontend` A pass can deactivate itself if the current frontend does not match the annotated frontend.

Open Questions:
- [x] What dependencies do we have? We should be able to shuffle the list of passes (search for shuffle in TranslationConfiguration.java) and pass all out tests, but we do not...
- [x] Register passes based on the current frontend: the frontend is unknown at the time of ordering the passes. Any ideas?
- [ ] Tests?
- [x] Random failures on local tests (but only one random test fails)
- [x] `PassRegisterExtraPass` uses the currently available frontends. However, this list contains the Java and c++ frontend even when parsing Java / LLVM...


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/PassIsFirstPass.kt
```diff
@@ -0,0 +1,35 @@
...
+package de.fraunhofer.aisec.cpg.passes
+
+/**
+ * Indicates whether this pass should be executed as the first pass. Note: setting this flag for
+ * more than one active pass will yield an error. Note: setting this flag will not activate the
+ * pass. You must register the pass manually.
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.CLASS)
+annotation class PassIsFirstPass
```
oxisto, line 35: I would suggest removing the `Pass` from all of those annotations, since they are already in the `passes´ package. `FirstPass` would be enough I would say.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/PassIsLastPass.kt
```diff
@@ -0,0 +1,35 @@
...
+package de.fraunhofer.aisec.cpg.passes
+
+/**
+ * Indicates whether this pass should be executed as the last pass. Note: setting this flag for more
+ * than one active pass will yield an error. Note: setting this flag will not activate the pass. You
+ * must register the pass manually.
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.CLASS)
+annotation class PassIsLastPass
```
oxisto, line 35: `LastPass`

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/PassOrderingPassWithDependencies.kt
```diff
@@ -0,0 +1,32 @@
...
+ *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.passes
+
+/** A simple helper class to match a pass with dependencies. */
+data class PassOrderingPassWithDependencies(
```
oxisto, line 29: Confusing class name.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/PassRegisterHardDependency.kt
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Register a hard dependency for the annotated pass. This ensures that:
+ * - the annotated pass is executed after its dependency
+ * - the dependency is added to the list of active passes even if not manually specified by the user
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.CLASS)
+@Repeatable
+annotation class PassRegisterHardDependency(val value: KClass<out Pass>)
```
oxisto, line 38: Can we merge this with `SoftDependency` and make it a value (maybe defaulting to "soft"? I would then suggest a simple `DependsOn` as annotation name.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/PassRegisterExtraPass.kt
```diff
@@ -0,0 +1,34 @@
...
+ */
+package de.fraunhofer.aisec.cpg.passes
+
+import kotlin.reflect.KClass
+
+/** Register a new pass required by a fronted. */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.CLASS)
+@Repeatable
+annotation class PassRegisterExtraPass(val value: KClass<out Pass>)
```
oxisto, line 34: `RegisterExtraPass`

oxisto (commented):
Looks nice from the functionality point of view. I have several suggestions for a cleaner "interface", e.g. the naming of the annotations.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/PassIsLastPass.kt
oxisto: or better `ExecuteLast`

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/PassIsFirstPass.kt
oxisto, line 35: or better `ExecuteFirst`

oxisto:
Some more dependencies:

- JavaExternalTypeHierarchyResolver depends on TypeHierarchyResolver
- ImportResolver depends on (TypeHierarchyResolver), possibly the JavaExternalTypeHierarchyResolver for Java at least
- TypeResolver depends on the call resolver (currently, in the future, it will be the other way around)

maximiliankaul:
@oxisto I updated this PR with your suggestions.

cpg-language-llvm/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/CompressLLVMPass.kt
```diff
@@ -35,13 +35,9 @@ import de.fraunhofer.aisec.cpg.graph.types.UnknownType
 import de.fraunhofer.aisec.cpg.helpers.SubgraphWalker
 import java.util.*
 
+@RequiredFrontend(LLVMIRLanguageFrontend::class)
 class CompressLLVMPass : Pass() {
```
KuechA, line 39: The CompressLLVMPass should be executed as the first pass (if possible)

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/RegisterDependency.kt
```diff
@@ -0,0 +1,44 @@
...
+ * - the annotated pass is executed after its dependency when `softDependency` is `false`:
+ * - the dependency is added to the list of active passes even if not manually specified by the user
+ * when `softDependency`is true:
+ * - the dependency is not added to the list of active passes - the order is only enforced if the
+ * user manually adds the pass
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.CLASS)
+@Repeatable
+annotation class RegisterDependency(
```
oxisto, line 41: My suggestion was just a simple `DependsOn` or do you prefer the `RegisterDependency` name? I thin annotations should probably describe a "state" rather than an "action".

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/Pass.java
```diff
@@ -86,4 +118,50 @@ public void setLang(@Nullable LanguageFrontend lang) {
...
+    return hardDependencies;
+  }
+
+  /**
+   * Check whether the current language matches the language required by [RequiredLanguage]
+   *
+   * @return true, if the pass does not require a specific language frontend or if it matches the
+   *     [RequiredLanguage]
+   */
+  public boolean runWithCurrentFrontend() {
```
oxisto, line 154: `runWithCurrentFrontend` sounds like "execute this pass with current frontend" to me. Maybe `run*s*WithCurrentFrontend` or something that is more fitting to a boolean value?

oxisto (commented):
Almost done :) Just some minor naming semantics.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/FirstPass.kt
```diff
@@ -0,0 +1,33 @@
...
+ *
+ */
+package de.fraunhofer.aisec.cpg.passes
+
+/**
+ * Indicates whether this pass should be executed as the first pass. Note: setting this flag for
+ * more than one active pass will yield an error. Note: setting this flag will not activate the
+ * pass. You must register the pass manually.
+ */
+@Retention(AnnotationRetention.RUNTIME) @Target(AnnotationTarget.CLASS) annotation class FirstPass
```
oxisto, line 33: One last change request: `ExecuteFirst` similar to the already existing `ExecuteLast`.

oxisto (commented):
99 % done :)

PR #904, opened 2022-09-09
Additional cleanup of test functions

oxisto:


oxisto:
We will re-visit this in the future

PR #915, opened 2022-09-19
Introduction of `callee` in `CallExpression`

oxisto:
This PR introduces an additional field `callee` in a call expression. The idea behind this field is to have an `Expression` that is being "called" by the call expression. For most cases, this will be a `DeclaredReferenceExpression` pointing to the function declaration. However, in some languages, other things might be "callable".

Note, that this is currently not yet used in the `CallResolver`, but only introduces the field in preparation of #773


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/statements/expressions/CallExpression.kt
```diff
@@ -97,6 +98,32 @@ open class CallExpression : Expression(), HasType.TypeListener, HasBase, Seconda
             value?.registerTypeListener(this)
         }
 
+    /**
+     * The expression that is being "called". This is currently not yet used in the [CallResolver]
+     * but will be in the future.
+     */
+    @field:SubGraph("AST")
+    var callee: Expression? = null
```
KuechA, line 106: Maybe, you could add some more details in the comment (e.g. as in the description of the PR).

For a `DeclaredReferenceExpression`, will the `refersTo` have to point to all the FunctionDeclarations in the `invokes` list (currently, `refersTo` can only have one element, not a list)? Does that make the `invokes` list superfluous? It would be helpful to specify a bit more the intended behavior

PR #923, opened 2022-09-22
refactor: setting base address

Paulo-Lopes-Estevao:
in many places in the code the address is repeated and noted that to a base.


oxisto:
Superseded by #968

PR #937, opened 2022-10-10
Improve FSM implementation by implementing both a NFA and a DFA

maximilian-galanis:
- Implements both a DFA and a NFA with methods to convert a NFA to a DFA and equality checks based on their accepted language
- Removes the IdentifierPass


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/IdentifierPass.kt
```diff
@@ -1,55 +0,0 @@
-/*
```
KuechA, line 1: Even if we don't need this pass any more for the order evaluation, I'd keep it just in case other projects make use of it. We should set deprecate it though

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/State.kt
```diff
@@ -32,64 +32,113 @@ package de.fraunhofer.aisec.cpg.analysis.fsm
  * - [isAcceptingState] indicates if this State accepts the FSM (in our case, this means that the
  * order of statements was correct).
  */
-class State(val name: String, val isStart: Boolean = false, var isAcceptingState: Boolean = false) :
-    Cloneable {
+sealed class State(name: Int, isStart: Boolean = false, isAcceptingState: Boolean = false) {
```
KuechA, line 35: Do we really want to restrict a user to an integer as a name? A String would be more readable...

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -170,9 +168,6 @@ open class Node : IVisitable<Node>, Persistable, LanguageProvider, ScopeProvider
      */
     var isImplicit = false
 
-    /** Required field for object graph mapping. It contains the node id. */
-    @field:Id @field:GeneratedValue var id: Long? = null
```
KuechA, line 174: Isn't this required by neo4j?

cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/DFAEqualityTest.kt
```diff
@@ -0,0 +1,250 @@
...
+import kotlin.test.*
+
+/**
+ * Tests whether the [DFA.equals] method works as expected and correctly identifies whether two DFAs
+ * accept the same language. If this test fails, the [NfaDfaConstructionTest] suite will also fail.
+ */
+class DFAEqualityTest {
+    @Test
+    /** Tests an empty DFA */
+    fun `test empty DFA`() {
```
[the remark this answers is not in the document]
maximilian-galanis: I think that it makes the test output more readable. This feature is mentioned in [Kotlin's coding conventions ](https://kotlinlang.org/docs/coding-conventions.html#names-for-test-methods)
If we don't want such method names in the CPG at all, I'm happy to use 'normal' names instead.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/IdentifierPass.kt
maximilian-galanis: I've re-added the pass

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/DFA.kt
```diff
@@ -27,145 +27,86 @@ package de.fraunhofer.aisec.cpg.analysis.fsm
...
-class DFA : Cloneable {
-    companion object {
-        @JvmStatic val EPSILON: String = "ε"
-    }
-
-    var states = mutableSetOf<State>()
-    private var stateCounter = 1
-    var currentState: State? = null
-    var executionTrace = mutableListOf<Triple<State, Node, BaseOpEdge>>()
+data class Trace(val state: State, val cpgNode: Node, val edge: Edge)
```
oxisto, line 30: a small kdoc would be nice.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/FSM.kt
```diff
@@ -0,0 +1,193 @@
...
+ *                   $$ |      $$$$$$$  |$$ |$$$$\
+ *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.analysis.fsm
+
+sealed class FSM(states: Set<State>) {
```
oxisto, line 28: kdoc

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/IdentifierPass.kt
```diff
@@ -30,6 +30,7 @@ import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.helpers.SubgraphWalker
 
 /** This [Pass] adds a unique ID to each [Node] in the CPG. */
+@Deprecated("Use a reference to a Node instead")
 class IdentifierPass : Pass() {
```
oxisto, line 34: I think we can remove it - or do we have any more dependencies to this pass? I think the ordering was the only use-case for it.

cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/DFAEqualityTest.kt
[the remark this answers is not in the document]
oxisto: I would also prefer non-space function names for now.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/IdentifierPass.kt
maximilian-galanis: We weren't sure whether applications using the CPG as a library still use it (i.e., the Clouditor). But if that's not the case, I'll just remove it.

oxisto (approved):

PR #963, opened 2022-11-07
Adding proper naming support to the CPG

oxisto:
This PR contains a major re-write of the naming system in the CPG. A node now only contains a full name (of the `Name` class), which is - depending on the node type - automatically populated by the current name space. The magic is handled by a new `MetadataProvider`, the `NamespaceProvider`.

PR #968, opened 2022-11-10
Using node builder functions for Go language frontend

oxisto:
- [x] Declaration builder
- [x] Expression builder
- [x] Statement builder


KuechA (commented):
I would have preferred it without the changes to the Builders to avoid conflicts with #963 but since it seems to affect only few locations, it should be ok.

KuechA (approved):

PR #998, opened 2022-11-17
Add missing DFG Edge for `NewExpression`s

KuechA:


oxisto (approved):
I remember that we had to do some ugly hacks for the type listener between new expression and initialiser but I do not see any reason now against a DFG edge.

konradweiss (approved):
Well concise and what expected from the PR title so ready to be merged

PR #1010, opened 2022-12-12
Add location to reference of member call

### LINKED ISSUE #1009: Graph nodes having NULL as the code location
I am trying to find nodes from code by matching the location of the code with the location information that is contained in the nodes. 
For that i need the startLine (and startColumn). 
As i understand it every "code" node should hold the information about the file it is in (artifact), its startLine and its startColumn
My problem is that most of the nodes don't have this information.

## Example / Steps to reproduce

I have this main.cpp:

```cpp
#include <array>

struct Buffer {
    std::array<int, 10> buf;

    void set(int idx, int value) {
        buf[idx] = value;
    }
};

static void foo() {
    Buffer b_good;
    b_good.set(9, 0);
}

static void bar() {
    Buffer b_bad;
    b_bad.set(10, 0);
}

int main(){
    foo();
    bar();
}
```

I have created a compile json containing only the main.cpp file:

```json
[
    {
        "file": "/home/andreas/masterthesis/sca-training/src/challenges/buffer/main.cpp"
    }
]

```

I created the graph with neo4j with this command:

```bash
Java_OPTS="Xss1024m" cpg-neo4j --host neo4j --json-compilation-database test_buffer.json 
```

I can see the graph in my browser and issue the following query:

```sql
MATCH(n:Node) WHERE  n.startLine is null  RETURN n.name, n.artifact, n.startLine, n.startColumn, n.isImplicit
```

and get the following result:

nullLocation

oxisto:
Fixes #1009

### DIFF
--- a/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.kt
+++ b/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.kt
@@ -428,7 +428,7 @@ class ExpressionHandler(lang: CXXLanguageFrontend) :
                     UnknownType.getUnknownType(language),
                     reference.name
                 )
-            member.location = frontend.getLocationFromRawNode<Expression>(reference)
+            member.location = frontend.getLocationFromRawNode(ctx.functionNameExpression)
             callExpression =
                 newMemberCallExpression(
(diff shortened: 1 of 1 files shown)


konradweiss (changes requested):
There is an empty file added to the test resources that I would remove from the PR, the rest is good.

PR #1011, opened 2022-12-12
Added DFG edges for `ForEachStatement`

### LINKED ISSUE #1006: `variable` and `iterable` from `ForEachStatement` are not connected in DFG
# Unexpected behavior

When analyzing code with a for each loop, there is no dataflow from the `iterable` to the `variable` of the `ForEachStatement`. The `variable` also has `UninitializedValue` as `initializer` which might be related to the missing dataflow.

# Example

```
class For {
    public void forIterator() {
        int[] a = {1,2,3,4,5};
        for(int j : a) {
            System.out.println(j);
        }
    }
}
```

```
            .defaultPasses()
            .defaultLanguages()
            .codeInNodes(true)
```

oxisto:
This adds DFG edges for the `ForEachStatement`. The data flows basically from the iterable to the variable.

Fixes #1006


konradweiss (approved):
I am not so sure if we want to remove the explicit representation of an uninitialized variable. On the other hand, I can imagine that even in the QL we can deduce that a DFG path to a `VariableDeclaration` implies that something might not be initialized. So I would say its fine if you don't have concerns and worst case we have to reintroduce the feature

oxisto:
> I am not so sure if we want to remove the explicit representation of an uninitialized variable. On the other hand, I can imagine that even in the QL we can deduce that a DFG path to a `VariableDeclaration` implies that something might not be initialized. So I would say its fine if you don't have concerns and worst case we have to reintroduce the feature

This class was only ever used on the java frontend and to be honest I never even knew of its existence. It's probably from a time where we did not allow `null` on an initializer.

PR #1013, opened 2022-12-14
Add getter for `conditions` in `LoopScope`

seelchen:
This PR adds a custom getter for `conditions` in `LoopScope` to get the condition of the corresponding loop statement.

### DIFF
--- a/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/LoopScope.kt
+++ b/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/LoopScope.kt
@@ -35,7 +35,37 @@ class LoopScope(loopStatement: Statement) :
     /**
      * Statements that constitute the start of the Loop depending on the used pass, mostly of size 1
      */
-    var starts: List<Node> = ArrayList()
+    val starts: List<Node>
+        get() =
+            when (val loopStatement = astNode) {
+                is WhileStatement -> {
+                    if (loopStatement.conditionDeclaration != null)
+                        SubgraphWalker.getEOGPathEdges(loopStatement.conditionDeclaration).entries
+                    else if (loopStatement.condition != null)
+                        SubgraphWalker.getEOGPathEdges(loopStatement.condition).entries
+                    else SubgraphWalker.getEOGPathEdges(loopStatement.statement).entries
+                }
+                is ForStatement -> {
+                    if (loopStatement.conditionDeclaration != null)
+                        SubgraphWalker.getEOGPathEdges(loopStatement.conditionDeclaration).entries
+                    else if (loopStatement.condition != null)
+                        SubgraphWalker.getEOGPathEdges(loopStatement.condition).entries
+                    else SubgraphWalker.getEOGPathEdges(loopStatement.statement).entries
+                }
+                is ForEachStatement -> {
+                    SubgraphWalker.getEOGPathEdges(loopStatement).entries
+                }
+                is DoStatement -> {
+                    SubgraphWalker.getEOGPathEdges(loopStatement.statement).entries
+                }
+                else -> {
+                    LOGGER.error(
+                        "Currently the component {} is not supported as loop scope.",
+                        astNode!!.javaClass
+                    )
+                    ArrayList()
+                }
+            }
 
     /** Statements that constitute the start of the Loop condition evaluation, mostly of size 1 */
     val conditions: List<Node>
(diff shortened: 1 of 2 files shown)


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/LoopScope.kt
```diff
@@ -38,7 +38,26 @@ class LoopScope(loopStatement: Statement) :
     var starts: List<Node> = ArrayList()
 
     /** Statements that constitute the start of the Loop condition evaluation, mostly of size 1 */
-    var conditions: List<Node> = ArrayList()
+    val conditions: List<Node>
```
oxisto, line 41: Do you know if/where the other variables in this scope class are populated? Is this done in a pass or are they also empty?

konradweiss: Looking at the code in the IDE the field is not used, however there is a function `starts()` that is used to return this list of nodes: [Loop starts](https://github.com/Fraunhofer-AISEC/cpg/blob/2666ba530f940d566e9506ea0f6cab96a9f398df/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/LoopScope.kt#L58-L91)
maybe we can just use that function as the getter. I can't tell you why there is a function and a field. Maybe something went wrong when we did the Java to Kotlin conversion or I refactored once without deleting the field.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/LoopScope.kt
```diff
@@ -38,7 +38,26 @@ class LoopScope(loopStatement: Statement) :
...
+        get() =
+            when (val node = astNode) {
+                is WhileStatement ->
+                    mutableListOf(node.condition, node.conditionDeclaration).filterNotNull()
+                is ForStatement -> mutableListOf(node.condition).filterNotNull()
+                is ForEachStatement -> mutableListOf(node.variable).filterNotNull()
```
konradweiss, line 47: This should be the iterator, however, the start of the loop is defined by the EOGPass and retrieved from the EOG path edges.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/LoopScope.kt
konradweiss: ah no MB, this is about the `conditions` field, I will leave the comment above in case its useful later, I will take a quick look myself, however, it seems we can delete starts.

konradweiss: So this was once used here: [Connect Continues](https://github.com/Fraunhofer-AISEC/cpg/blob/c4239ae4fda4cdbe103695f1848499f30b7f09d7/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt#L750-L759) and was later replaced with inplace switching, however as you can see it has a different `condition` for the `ForEachStatement`. Yours is probably better/ more intuitive. If you could also change that function to use your `conditions` it would be nice.

seelchen: I changed it as you suggested, see 533c68955d5f4eadf655e23664070977af9bad61. I also added the `AssertStatement` case from [Connect Continues](https://github.com/Fraunhofer-AISEC/cpg/blob/c4239ae4fda4cdbe103695f1848499f30b7f09d7/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt#L750-L759) to the getter.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/LoopScope.kt
```diff
@@ -38,7 +38,27 @@ class LoopScope(loopStatement: Statement) :
     var starts: List<Node> = ArrayList()
```
oxisto, line 38: While you are at it: Could you remove the `starts()` function and move the logic into the getter only property `starts`, similar to `conditions`?

seelchen: Done in cd9ea0b140253e541c004c082bb9407671c9d8a2. I also changed the `when` expression slightly to enable smart casting and removed all explicit casts.

oxisto (approved):

Issue #204, opened 2020-08-20
Heuristic for subtypes


Masrepus:
Consider the following situation:
```java
A a = new B();  // B extends A
```
Here, we currently get `type = A` and `possibleSubTypes = {A, B}`, when actually, we would need to get `type = B` and `possibleSubTypes = {B}`, as we only have a single possibility for a's type here.

So my first idea was to check during `setType` whether `possibleSubTypes` only contains the common type (here: A) and one descendant (here: B). If this is the case, we know that we can reduce it to only B. Unfortunately, we can also have this:
```java
A a = someFlag ? new A() : new B();
```
Here, we explicitly can't know for sure whether we have type A or B, so indeed `possibleSubTypes = {A, B}` would be correct. In the solution discussed above, this would again be reduced to `{B}`, this time incorrectly. There are many more ways how this could happen, e.g. depending on previous branches being taken.

Now I'm not sure whether we want to
1. Accept the current situation, where we might have both A and B as subtypes in cases where we could actually be more precise
2. Try to find any heuristic on when to reduce the set. This e.g. affects call resolution, where two `invokes` edges will be created in some situations, one to the original method in A and one to the actually overriding method in B.

PR #1025, opened 2023-01-09
Add types to the result of basic binary operations

KuechA:
Currently, the results of binary operations receive a type only in one of these two cases:
1) it's a simple assignment with the `=` operator, or
2) it's an append (`+`) and one of the values lhs/rhs is of type `java.lang.String`.

I think we can do better at least for arithmetic types (+, -, *, /, etc.). However, as this is quite specific to languages, I added the code to the `Language` class so that we can control the types which should be considered and the exact operators. Also, the exact semantics could differ between the programming languages.


KuechA:
Blocked by #1033

KuechA:
Fixes #1071

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/BinaryOperator.kt
```diff
@@ -117,20 +119,39 @@ class BinaryOperator : Expression(), HasType.TypeListener, Assignment, HasBase,
         }
         val previous = type
         if (operatorCode == "=") {
+            if (
+                src == rhs &&
```
KuechA, line 123: This is a quick fix for #1071 but only holds for numbers. For other types (e.g. classes), we probably have to check what we want to propagate and is somewhat related to #204.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -111,6 +119,51 @@ abstract class Language<T : LanguageFrontend> : Node() {
...
     }
+
+    /**
+     * Determines how to propagate types across binary operations since these may differ among the
+     * programming languages.
+     */
+    open fun propagateTypeOfBinaryOperation(operation: BinaryOperator): Type {
+        if (operation.operatorCode == "==" || operation.operatorCode == "===") {
+            // A comparison, so we return the type "boolean"
+            return TypeParser.createFrom("boolean", this)
```
oxisto, line 130: We might need to return a specific "BoolType" depending on the language here.

oxisto: A BooleanType should exist from #1104

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -111,6 +119,53 @@ abstract class Language<T : LanguageFrontend> : Node() {
...
+
+    /**
+     * Determines how to propagate types across binary operations since these may differ among the
+     * programming languages.
+     */
+    open fun propagateTypeOfBinaryOperation(operation: BinaryOperator): Type {
+        if (operation.operatorCode == "==" || operation.operatorCode == "===") {
+            // A comparison, so we return the type "boolean"
+            return this.simpleTypes.values.firstOrNull { it is BooleanType }
+                ?: this.simpleTypes.values.firstOrNull { it.name.localName.startsWith("bool") }
```
oxisto, line 131: Probably just return UnknownType in this case, everything else is a little bit of a weird guess work I would say.

KuechA: Also in the second case looking for a name starting with `"bool"`? The BooleanType exists but it is almost never used. So, if we want to get rid of this check, we have use that type in all languages and not only in go

oxisto: > Also in the second case looking for a name starting with `"bool"`? The BooleanType exists but it is almost never used. So, if we want to get rid of this check, we have use that type in all languages and not only in go

Well, we should ;)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
```diff
@@ -609,12 +650,47 @@ infix fun Expression.eq(rhs: Expression): BinaryOperator {
...
+    node.lhs = this
+    node.rhs = init(node)
+
+    (this@StatementHolder) += node
+
+    return node
+}
+
+/**
+ * >>>>>>> main Creates a new [BinaryOperator] with a `=` [BinaryOperator.operatorCode] in the
```
oxisto, line 688: same here

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
oxisto: I can quickly whip up a PR that does that for the supported languages.

oxisto: See. https://github.com/Fraunhofer-AISEC/cpg/pull/1113.

After some consideration we can probably leave the look for `boolXXX` part in, because not all languages necessarily have a "real" boolean type. Although its mainly C and python and I guess that we need to implement this particular function for those languages anyway since they are crazy?

konradweiss (approved):

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
```diff
@@ -98,9 +98,9 @@ class PythonLanguage : Language<PythonLanguageFrontend>(), HasShortCircuitOperat
             ) {
                 // In Python, the // operation keeps the type as an int if both inputs are integers
                 // or casts it to a float otherwise.
-                return simpleTypes["int"]!!
+                return builtInTypes["int"]!!
```
oxisto, line 101: There is an (undocumented) function `getSimpleTypeOf` in `Language` that you used in other occasions for this.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
```diff
@@ -86,7 +86,7 @@ class PythonLanguage : Language<PythonLanguageFrontend>(), HasShortCircuitOperat
                 operation.rhs.propagationType is NumericType
         ) {
             // In Python, the / operation automatically casts the result to a float
-            return simpleTypes["float"]!!
+            return builtInTypes["float"]!!
```
oxisto, line 89: same here

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
oxisto: There is also this `newPrimitiveType` extension function which confuses me even more because it seems to return a type but then modifier its modifier?!

KuechA: Yes, the `newPrimitiveType` function is quite strange. It was used only in 3 places, so I just removed it.

I replaced the usage of `builtInTypes` with `getSimpleTypeOf`

PR #1028, opened 2023-01-11
Fixed C++ FQN for methods with unknown classes

oxisto:
The FQN of method decorations for unknown classes were wrong. Additionally, there was too much boilerplate code in the C++ method name handling (there still probably is). This PR relieves some of this pain.


konradweiss (approved):
Looks good to me

peckto (approved):

PR #1030, opened 2023-01-11
Improve order evaluation

maximilian-galanis:
This pull request improves the DFA order evaluation in two aspects: 
- it allows a CPG node to correspond to multiple names allowing following syntax in Codyze/Coko:
```Kotlin
order(testObj::constructor) {
            +testObj::start.use { call(1) }
            +testObj::start
            +testObj::finish
        }
```
In this case, `testObj::start.use { call(1) }` corresponds to the set of `start` nodes that have `1` as an argument compared to `testObj::start`, which returns all nodes of `start`. The CPG nodes contained in both sets therefore need the ability to map to multiple names.
- Implement the `consideredResetNodes` feature, which allows the `DFAOrderEvaluator` to reset the order evaluation on a base when a reset node is encountered. This allows the correct evaluation of cases like the following:
```Java
Botan p7 = new Botan(2);
p7.start(iv);
p7.finish(buf);

p7 = new Botan(3); // Botan(3) would be in the consideredResetNodes and thus reset the order evaluation
p7.start(iv);
p7.finish(buf);
```
Before this change, the `DFAOrderEvaluator` would have triggered a false order evaluation error not realizing that `p7 = new Botan(3);` must finalize & reset the ongoing order evaluation.

Additionally, it moves `Node.followNextEOG` to the correct `Extensions.kt` file.

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/DFAOrderEvaluator.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/DFAOrderEvaluator.kt
@@ -49,6 +49,18 @@ import org.slf4j.LoggerFactory
  * - [nodeToRelevantMethod]: A mapping between CPG nodes and their operators used by the respective
  * edges in the [dfa]. Currently, we only consider [CallExpression]s. If a node is not contained in
  * this list, it is not considered by the evaluation as we assume that the method is not relevant.
+ * - [consideredResetNodes]: These nodes reset the order evaluation such that e.g., a reassignment of a variable with a
+ * new object is handled correctly. In this case, the constructor node must be part of the [consideredResetNodes].
+ * This allows the [DFAOrderEvaluator] to detect that in such a case,
+ * ```
+ *        (1) Botan p7 = new Botan(2);
+ *        (2) p7.start(iv);
+ *        (3) p7.finish(buf);
+ *
+ *        (4) p7 = new Botan(3);
+ *        (5) p7.start(iv);
+ * ```
+ * (4) should actually start a new order evaluation.
  * - [thisPositionOfNode]: If a non-object oriented language was used, this is a map from CPG nodes
  * (i.e., the [CallExpression]) to the argument position serving as base of the operation.
  *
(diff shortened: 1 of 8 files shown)


oxisto (approved):
Looks good to me, as far as I can judge. Some description on the PR itself would be nice though.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/DFAOrderEvaluator.kt
```diff
@@ -151,10 +167,39 @@ open class DFAOrderEvaluator(
...
+                // Handle 'reset nodes'.
+                if (node in consideredResetNodes) {
+                    val baseAndOp =
+                        getBaseAndOpOfNode(node as CallExpression, eogPath, interproceduralFlows)
+                    if (baseAndOp != null) {
+                        val (base, _) = baseAndOp
+                        // get the DFA associated with this base
+                        val dfa = baseToFSM.computeIfAbsent(base) { dfa.deepCopy() }
+
+                        if (dfa.isAccepted) {
```
KuechA, line 179: A small comment would be good here so that we remember what we did ;)

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/DFAOrderEvaluator.kt
```diff
@@ -151,10 +167,39 @@ open class DFAOrderEvaluator(
...
+                            actionAcceptingTermination(
+                                base,
+                                dfa,
+                                interproceduralFlows[base] == true
+                            )
+                        } else if (dfa.currentState?.isStart == true) {
+                            // nothing to do here, we just want to continue with the next nodes in
+                            // the EOG
+                        } else {
+                            actionNonAcceptingTermination(
```
KuechA, line 189: A comment would be nice

KuechA (commented):
Some comments could be nice but overall it looks good

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/fsm/DFAOrderEvaluator.kt
maximilian-galanis: added a comment explaining that this evaluates the DFA when a reset node is encountered

PR #1031, opened 2023-01-12
Restructuring `CallExpression` and beginning to use `callee` in `CallResolver`

### LINKED ISSUE #298: Type of "member" field of MemberCallExpression
I am not sure, whether we discussed this before, but why is the `member` field of `MemberCallExpression` a generic `Node`? It seems mostly `DeclaredReferenceExpression`s are assigned to it. But wouldn't it even make more sense to restrict it to the type `MemberExpression` (which also inherits from `DeclaredReferenceExpression`)?

### LINKED ISSUE #495: Implement TypeScript FQN

### LINKED ISSUE #689: Introduce `callee` property on a `CallExpression`
In languages with function points, we could have functions as part of a struct, which will verry mutch look like a `MemberCallExpression` but which is still a `CallExpression`

Example:
```c++
struct A {
        void foo(int i) {
        }
};

struct B {
        void (*bar)(int);
};

void bar(int i) {
}

int main() {
        A a;
        a.j = 1;
        B b;
        b.bar = &bar;

        // foo is a method
        (a.foo)(1); 
        a.foo(2); 

        // bar is a function pointer
        (b.bar)(3);
        (*b.bar)(4);

        return 0;
}
```

Because the `CallExpression` has no `member` property, we cannot reflect this expression in the current graph layout.
As discussed in #682, a clean solution would be to add a `callee` property to the `CallExpression` and maybe get rid of the `base` property of the `MemberCallExpression`.

The new property could look like the following:
callee-property

In this issue we can discuss this change in the graph layout and possible implications.

### LINKED ISSUE #1026: memberCallExpression is  mistakenly parsed as staticCallExpression due to the package path of import
# Unexpected behavior
The node type of `test.doSomething` on line 6 is affected by package path on first line. If penultimate package name which ends with `.*` is the same as base name of `memberCallExpression`, then the `memberCallExpression` will be mistakenly parsed as `staticCallExpression`, e.g. node on line 6. Modify package name "test" on first line to others, the node type is parsed as `memberCallExpression` correctly.

# Code of interest
package structure
```
|--com
   |--test
      |--Test.java
|--Sample.java
```
Sample.java
```java
import com.test.*;  // line 1

public class Sample {
    public void func() {
        Test test = new Test();
        test.doSomething();  // line 6
    }
}
```

# Configuration
v5.1.0
```java
new TranslationConfiguration.Builder()
                .sourceLocations(new File(filename))
                .defaultPasses()
                .defaultLanguages()
                .build();
```

### LINKED ISSUE #1038: A lot of warnings for no code for language node
We are getting a lot of warnings like this recently:

```
WARN  LanguageFrontend Unexpected: No code for node CLanguage[location=<null>]
```

oxisto:
This PR proposes a major restructuring of `CallExpression` and `MemberCallExpression` nodes. The idea to have a more coherent way of dealing with calls, as well as enabling some new features to resolve calls to any kind of expressions (already modelled in the `callee` property).  It also removes a lot of converting between `MemberExpression` nodes and `DeclaredReferenceExpression` nodes, since now you can just use the `MemberExpression` in the `callee` field.

I also started to refactor the `CallResolver` to take the information in the `callee` property, but it still needs a major refactor (or possible rewrite). However, this might push us into the right direction.

Changes:
- [x] Removes `base` from the `CallExpression`
- [x] Only a `MemberCallExpression` now has a `base`
- [x] Using `callee` in the call resolving (sort of)
- [x] Using `BinaryOperator` with `.*` operator for function pointer calls to class methods
- [x] Removing `StaticCallExpression` in favor of an `isStatic` variable in `MemberCallExpression`

Fixes #495
Fixes #689
Fixes #298
Fixes #1026
Fixes #1038


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/ExpressionBuilder.kt
```diff
@@ -329,36 +321,14 @@ fun MetadataProvider.newMemberExpression(
...
 }
 
-/**
- * Creates a new [StaticCallExpression]. The [MetadataProvider] receiver will be used to fill
- * different meta-data using [Node.applyMetadata]. Calling this extension function outside of Kotlin
- * requires an appropriate [MetadataProvider], such as a [LanguageFrontend] as an additional
- * prepended argument.
- */
-@JvmOverloads
-fun MetadataProvider.newStaticCallExpression(
```
KuechA, line 347: I think this means that we should delete the class `StaticCallExpression` and everything is now a `MemberCallExpression`. Do we actually need this subclass of `CallExpression`?

gradle.properties
```diff
@@ -1,6 +1,6 @@
 org.gradle.jvmargs=-Xmx4096m -XX:MaxMetaspaceSize=1024m -Dkotlin.daemon.jvm.options=-Xmx4g
 
-enableGoFrontend=false
-enablePythonFrontend=false
-enableLLVMFrontend=false
-enableTypeScriptFrontend=false
+enableGoFrontend=true
```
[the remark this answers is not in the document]
oxisto: Not intentionally, the old git problem

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/ExpressionBuilder.kt
oxisto: The class should have been deleted. I will double-check

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.kt
```diff
@@ -454,18 +445,14 @@ class ExpressionHandler(lang: CXXLanguageFrontend) :
...
-            callExpression =
-                newMemberCallExpression(
-                    ctx.functionNameExpression.rawSignature,
-                    "",
-                    reference.lhs,
-                    reference.rhs,
-                    reference.operatorCode,
-                    ctx.rawSignature
-                )
+        } else if (
```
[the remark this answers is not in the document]
oxisto: The old code was basically a misunderstanding. We thought that sometimes an expression like `a.b` would occur as a BinaryOperator with `.`. However, the only circumstances is a a function pointer call in the form of `a.*b` and in this case the operator is `.*` and not `.`. We were mistakenly thinking that the `pmdot` operator was a `.`, however it is a "pointer to member dereference" operator (see https://github.com/eclipse-cdt/cdt/blob/7c6bd5bdcb4cf8c31050aeeb102250a27a157728/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/core/dom/ast/IASTBinaryExpression.java#L203) and should be parsed as `.*` (see https://github.com/eclipse-cdt/cdt/blob/7c6bd5bdcb4cf8c31050aeeb102250a27a157728/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/rewrite/astwriter/ExpressionWriter.java#L90)

KuechA (approved):

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CPPLanguage.kt
```diff
@@ -147,14 +150,15 @@ class CPPLanguage :
...
+            // resolver, we resolve the call. Otherwise, there won't be an inferred template, we
+            // will do an
+            // inferred FunctionDeclaration instead.
             call.templateParametersEdges = mutableListOf()
-            if (handleTemplateFunctionCalls(null, call, false, scopeManager, currentTU)) {
-                return
+            val (ok, candidates) =
+                handleTemplateFunctionCalls(null, call, false, scopeManager, currentTU)
+            if (ok) {
+                return candidates
```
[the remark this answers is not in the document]
oxisto: Not quite sure how else to do it? in case the template function resolution is "ok", we need to immediately return, otherwise weird things happen.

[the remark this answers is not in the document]
konradweiss: Ah maybe I was to unclear, the Issue is the else block after a then block with a return. The recommended way to write it is to avoid the `else {...}` and move the contained code after the If.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/statements/expressions/ArraySubscriptionExpression.java
```diff
@@ -119,4 +120,10 @@ public int hashCode() {
   public Expression getBase() {
     return this.arrayExpression;
   }
+
+  @Nullable
+  @Override
+  public String getOperatorCode() {
```
konradweiss, line 126: Why is this necessary or was there a specific design idea behind this change. Also, why does it have a base, I would only expect these to exist if they inherited from BinaryOperator or MemberExpression.

konradweiss: Ah no, we made it an implementation of `HasBase`, not so sure about the `getOperatorCode` though

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/CPPLanguage.kt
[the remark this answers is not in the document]
oxisto: Ah now I get it.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/statements/expressions/ArraySubscriptionExpression.java
oxisto: the array subscription expression always implemented `HasBase`, which makes sense, since it operates on a specific base (which can be check for null for example). I decided to extend `HasBase` with the `operatorCode`, which was already part of the BinaryOperator and the MemberExpression, to differentiate *how* one access the base.

konradweiss: Ok, I like the thought behind it. It only extends the things we can do with the node type so its a good change.

konradweiss (approved):
Mergeable :) The only Issue remaining is of style and readability which I would like to see changed but is up to the the author.

PR #1033, opened 2023-01-13
More fine-grained and language specific interpretation of primitive types

KuechA:
The interpretation of primitive or at least very basic types (e.g. integers, floating point values, Strings) is different for each language. Currently, they are all contained in the `ObjectType` class and are generated in a generic way. This representation is not very accurate for many cases and, instead, it would be nice to have at least types for integers, floats and strings which can be modified based to the programming language where it's used. This should simplify and generalize subsequent analyses.

Tasks:
* [x] Add more specific subtypes of `ObjectType`
* [x] Move some of the type parsing code to the `Language` class and make it extensible
* [x] Implement the functions for each supported language
* [ ] Refactoring of the `TypeParser`


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -58,12 +65,84 @@ abstract class Language<T : LanguageFrontend> : Node() {
...
+    open val accessModifiers: Set<String>
+        get() = setOf("public", "protected", "private")
+
     /** Creates a new [LanguageFrontend] object to parse the language. */
     abstract fun newFrontend(
         config: TranslationConfiguration,
         scopeManager: ScopeManager = ScopeManager()
     ): T
 
+    open fun getTypeOf(typeString: String, modifier: ObjectType.Modifier): Type? {
```
oxisto, line 78: the naming should be something related to "simple" types I guess.

I also wonder, whether it would make more sense to move this to a `LanguageFrontend`? I would rather keep "logic" out of the language class.

Alternatively, we could provide a `Map` that maps between a name/string and a type? That would suit the configuration style of the language class better I would say.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -58,12 +65,84 @@ abstract class Language<T : LanguageFrontend> : Node() {
...
+    private val RESTRICT_QUALIFIER = "restrict"
+    private val ATOMIC_QUALIFIER = "atomic"
+
+    open fun updateQualifier(qualifierString: String, old: Qualifier): Boolean {
+        if (this !is HasQualifier || qualifierString !in qualifiers) {
+            return false
+        }
+        val qualifier = old
+        when (qualifierString) {
+            FINAL_QUALIFIER,
```
oxisto, line 106: Do we even use these qualifiers?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -58,12 +65,84 @@ abstract class Language<T : LanguageFrontend> : Node() {
...
+            }
+        }
+
+        return false
+    }
+
+    open fun asStorageSpecifier(specifier: String): Storage? {
+        // TODO: Which of these affect which language? Probably, we should separate it more clearly.
+        // I'm also wondering why we actually need this information.
+        if (specifier.uppercase() == "STATIC") {
```
oxisto, line 133: Same here? Is that even used somewhere?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/FloatingPointType.kt
```diff
@@ -0,0 +1,62 @@
...
+
+import de.fraunhofer.aisec.cpg.frontends.Language
+import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend
+import de.fraunhofer.aisec.cpg.graph.parseName
+
+class FloatingPointType : ObjectType {
+    val bitwidth: Int
+
+    constructor() : super() {
+        bitwidth = -1
```
oxisto, line 36: `Int?` instead? We probably forget to check for `-1` at some point in a calculation and I can only imagine the face palming moments ;)

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/IntegerType.kt
```diff
@@ -0,0 +1,62 @@
...
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.types
+
+import de.fraunhofer.aisec.cpg.frontends.Language
+import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend
+import de.fraunhofer.aisec.cpg.graph.parseName
+
+class IntegerType : ObjectType {
+    val bitwidth: Int
```
oxisto, line 33: `bitWidth` or `bitSize`?

oxisto (commented):
In general I think its a good idea to enable some better analyses in the future and avoid weird comparing with type == "java.lang.String".

I do have some concerns about "polluting" the `Language` class though

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/Language.kt
KuechA: What do you mean with "use"? We don't know what the end-user does with this information. But I agree, it seems to be information which is not too interesting

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/Language.kt
KuechA: I would not want to move it to the `LanguageFrontend` because we don't know when this method or the `TypeParser` will be used. I.e., if it is in a Pass, the frontend won't be available. But the idea with the `Map` sounds good (and I also thought of that already)

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/IntegerType.kt
KuechA: Changed it to `bitWidth`

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/Language.kt
KuechA: I'm wondering why this belongs to the `Type`. I think that this should be a property of the variable

oxisto: Yeah, this does not belong to the type. I think we can remove it and maybe later re-introduce it in a declaration if needed.

cpg-language-go/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontendTest.kt
```diff
@@ -167,7 +167,7 @@ class GoLanguageFrontendTest : BaseTest() {
         make = decl.initializer
         assertNotNull(make)
         assertTrue(make is ConstructExpression)
-        assertEquals(TypeParser.createFrom("map<string,string>", GoLanguage()), make.type)
+        assertEquals(TypeParser.createFrom("map[string,string]", GoLanguage()), make.type)
```
oxisto, line 170: This is actually not really a generic type

oxisto:
Note to myself: Move `TypeCache` to `Component`

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/IntegerType.kt
```diff
@@ -0,0 +1,59 @@
...
+
+import de.fraunhofer.aisec.cpg.frontends.Language
+import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend
+import de.fraunhofer.aisec.cpg.graph.Name
+
+/** Instances of this class represent integer types. */
+class IntegerType : NumericType {
+    constructor() : super()
+
+    constructor(
```
oxisto, line 36: Can't we merge all of these constructors together into one with default arguments?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/NumericType.kt
```diff
@@ -0,0 +1,92 @@
...
+import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend
+import de.fraunhofer.aisec.cpg.graph.Name
+import java.util.*
+
+/** This type collects all kind of numeric types. */
+open class NumericType : ObjectType {
+    val bitWidth: Int?
+    val modifier: Modifier
+
+    constructor() : super() {
```
oxisto, line 38: Same as above

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/StringType.kt
```diff
@@ -23,27 +23,33 @@
...
 
-  @Override
-  public Type.Qualifier toEntityAttribute(Map<String, ?> value) {
-    Map<String, Boolean> val = (Map<String, Boolean>) value;
-    return new Type.Qualifier(
-        val.get("isConst"), val.get("isVolatile"), val.get("isRestrict"), val.get("isAtomic"));
-  }
+    constructor() : super()
+
+    constructor(
```
oxisto, line 36: Soo many constructors

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/LegacyTypeManager.java
```diff
@@ -52,14 +52,14 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class TypeManager {
+public class LegacyTypeManager {
```
oxisto, line 55: Should I revert this renaming? Then less files would be changed

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -116,44 +116,51 @@ fun Node.applyMetadata(
...
+            }
+        this.name = this.newName(name, localNameOnly, namespace)
     }
 
     if (codeOverride != null) {
         this.code = codeOverride
     }
 }
 
+fun LanguageProvider.newName(
```
oxisto, line 134: doc missing. Should we set this to private?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/TypeBuilder.kt
```diff
@@ -25,32 +25,29 @@
  */
 package de.fraunhofer.aisec.cpg.graph
 
+import de.fraunhofer.aisec.cpg.graph.types.NumericType
+import de.fraunhofer.aisec.cpg.graph.types.NumericType.Modifier
 import de.fraunhofer.aisec.cpg.graph.types.ObjectType
-import de.fraunhofer.aisec.cpg.graph.types.ObjectType.Modifier
-import de.fraunhofer.aisec.cpg.graph.types.Type
 
 fun MetadataProvider.newPrimitiveType(
```
oxisto, line 32: Extend it from LangaugProvider instead of MetadataProvider?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/FloatingPointType.kt
```diff
@@ -0,0 +1,60 @@
...
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.types
+
+import de.fraunhofer.aisec.cpg.frontends.Language
+import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend
+import de.fraunhofer.aisec.cpg.graph.Name
+
+/** Instances of this class represent floating point types. */
+class FloatingPointType : NumericType {
```
maximiliankaul, line 33: Probably a Kotlin thing I'm not aware of: why do we need all these constructors? I'm in favor of keeping it simple and just having one.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/IntegerType.kt
maximiliankaul: +1

maximiliankaul (approved):
LGTM. It would be easier to review changes if PR wouldn't be sooooooo huge in the future.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/IntegerType.kt
KuechA: Yes... That was mostly the result of the "old" system requiring many more fields. I removed all but the default constructor and one with the arguments. The default one is required by ogm as far as I know, right?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/TypeBuilder.kt
KuechA: Yes

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/types/IntegerType.kt
oxisto: As far as I know the neo4j OGM magically injects these default constructors, so there are not needed imho.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
KuechA: I think we can keep it public

PR #1052, opened 2023-01-19
Fix DFG of the variable of a`ForEachStatement`

### LINKED ISSUE #1006: `variable` and `iterable` from `ForEachStatement` are not connected in DFG

KuechA:
The variable declaration in a `ForEachStatement` does not have an initializer and is therefore not really considered by the `ControlFlowSensitiveDFGPass`. We have to handle this case separately.

Closes #1006


oxisto (commented):
Do we need to handle anything else in the normal DFG pass?

KuechA:
No, actually, the respective edges from the `DFGPass` would probably be removed anyway... I tried an alternative approach by setting the initializer of the variable in the normal DFG Pass with the iterable but I didn't like the solution (even if it would have worked).

oxisto:
> No, actually, the respective edges from the `DFGPass` would probably be removed anyway... I tried an alternative approach by setting the initializer of the variable in the normal DFG Pass with the iterable but I didn't like the solution (even if it would have worked).

I just meant, if the user is missing out on this "feature", if he is only using the "normal" DFG pass?

KuechA:
> > No, actually, the respective edges from the `DFGPass` would probably be removed anyway... I tried an alternative approach by setting the initializer of the variable in the normal DFG Pass with the iterable but I didn't like the solution (even if it would have worked).
> 
> I just meant, if the user is missing out on this "feature", if he is only using the "normal" DFG pass?

Good point. Since the variable declaration is wrapped in a DeclarationStatement, the DFG edge was indeed not added only with the normal DFG pass. Btw. the `DeclarationStatement` has a DFG edge to the `ForEachStatement`. Do you have any idea why this is the case? We do not add this edge in our passes

oxisto:
> > > No, actually, the respective edges from the `DFGPass` would probably be removed anyway... I tried an alternative approach by setting the initializer of the variable in the normal DFG Pass with the iterable but I didn't like the solution (even if it would have worked).
> > 
> > 
> > I just meant, if the user is missing out on this "feature", if he is only using the "normal" DFG pass?
> 
> Good point. Since the variable declaration is wrapped in a DeclarationStatement, the DFG edge was indeed not added only with the normal DFG pass. Btw. the `DeclarationStatement` has a DFG edge to the `ForEachStatement`. Do you have any idea why this is the case? We do not add this edge in our passes

uhm no idea :D @konradweiss ?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -206,6 +207,23 @@ open class ControlFlowSensitiveDFGPass : Pass() {
...
+                        as? VariableDeclaration
+
+                writtenDecl?.let { wd ->
+                    iterable?.let { wd.addPrevDFG(it) }
+                    // Add the variable declaration to the list of previous write nodes in this path
+                    previousWrites[wd] = mutableListOf(wd)
+                }
             }
 
             // Check for loops: No loop statement with the same state as before and no write which
```
oxisto, line 229: While you are editing this file, the comment for `handleVariableDeclaration` is wrong (copy/paste) from function declaration.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
```diff
@@ -90,6 +92,15 @@ class DFGPass : Pass() {
         }
     }
 
+    /**
+     * For a [DeclarationStatement], the whole statement flows into its declarations. This is fine
+     * because it's used as a wrapper if a Statement is needed but we only have a Declaration (which
+     * is not a statement).
+     */
+    private fun handleDeclarationStatement(node: DeclarationStatement) {
```
oxisto, line 100: Decision: Not handling this for *all* declaration statements, but only do this for the `ForStatement`.

KuechA: I removed this method and handle the `ForEachStatement` differently now. @maximiliankaul please check if the new handling makes sense to you

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -76,7 +77,7 @@ open class ControlFlowSensitiveDFGPass : Pass() {
      * Removes all the incoming and outgoing DFG edges for each variable declaration in the function
      * [node].
      */
-    private fun clearFlowsOfVariableDeclarations(node: FunctionDeclaration) {
+    private fun clearFlowsOfVariableDeclarations(node: Node) {
```
maximiliankaul, line 80: Why now `Node` instead of the `FunctionDeclaration`?  The doc string reads like this should be reverted.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -93,7 +94,7 @@ open class ControlFlowSensitiveDFGPass : Pass() {
      * - Assignments with an operation e.g. of the form "variable += rhs"
      * - Read operations on a variable
      */
-    private fun handleFunction(node: FunctionDeclaration) {
+    private fun handleFunction(node: Node) {
```
maximiliankaul, line 97: Why does `handleFunction` receive a `Node` instead of a `FunctionDeclaration`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -208,6 +209,59 @@ open class ControlFlowSensitiveDFGPass : Pass() {
...
+            } else if (currentNode is ForEachStatement && currentNode.variable != null) {
+                // The VariableDeclaration in the ForEachStatement doesn't have an initializer, so
+                // the "normal" case won't work. We handle this case separately here...
+                // This is what we write to the declaration
+                val iterable = currentNode.iterable as? Expression
+
+                val writtenTo =
+                    when (currentNode.variable) {
+                        is DeclarationStatement ->
+                            (currentNode.variable as DeclarationStatement).singleDeclaration!!
```
maximiliankaul, line 221: I assume this works "right now", but this is broken once we have support for multiple vars (#1089). Correct?

maximiliankaul (approved):
I'm generally in favor of this PR. The only thing I dislike is the heavy use of `!!` in the pass.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
KuechA: That's correct. Actually, this will probably require a couple of changes in this pass (but this will depend on the implementation of course)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
KuechA: See above

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
KuechA: I updated the text to fit the change. I changed it to `Node` because probably, we also want to handle blocks of code which aren't in any function but (e.g. a script)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
maximiliankaul: Let's worry about this in the future...

oxisto:
> I'm generally in favor of this PR. The only thing I dislike is the heavy use of `!!` in the pass.

Agree. Can you change it to a local variable assignment + smart card instead? We should really avoid !!

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -208,6 +209,59 @@ open class ControlFlowSensitiveDFGPass : Pass() {
...
+                            (currentNode.variable as DeclarationStatement).singleDeclaration
+                        else -> currentNode.variable
+                    }
+
+                // We wrote something to this variable declaration
+                writtenDecl =
+                    when (writtenTo) {
+                        is Declaration -> writtenTo
+                        is DeclaredReferenceExpression -> writtenTo.refersTo
+                        else -> null // TODO: This shouldn't happen
```
konradweiss, line 230: Maybe not a todo, either a normal comment or a log warning IMO, what do you think? I would avoid TODO if there is nothing that we actually want to implement here. Maybe, `NOTE`

konradweiss (changes requested):

PR #1054, opened 2023-01-19
fix configure frontends script

maximiliankaul:
The current version of the configure frontends script fails on my machine. I suppose the following change to make it work.
@maximilian-galanis is this a mac vs linux thing or just a bug?

### DIFF
--- a/configure_frontends.sh
+++ b/configure_frontends.sh
@@ -20,7 +20,6 @@ function setProperty {
     sed -i "s/\(${PROP_KEY}[[:space:]]*=[[:space:]]*\).*\$/\1${PROP_VALUE}/" $GRADLE_PROPERTIES_FILE
   fi
 }
-}
 
 function ask() {
   local _ask_answer
--- a/configure_frontends.sh
+++ b/configure_frontends.sh
@@ -14,7 +14,12 @@ function getProperty {
 function setProperty {
   local PROP_KEY=$1
   local PROP_VALUE=$2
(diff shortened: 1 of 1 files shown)


configure_frontends.sh
```diff
@@ -14,7 +14,7 @@ function getProperty {
 function setProperty {
   local PROP_KEY=$1
   local PROP_VALUE=$2
-  sed -i '' "s/\(${PROP_KEY}[[:space:]]*=[[:space:]]*\).*\$/\1${PROP_VALUE}/" $GRADLE_PROPERTIES_FILE
+  sed -i "s/\(${PROP_KEY}[[:space:]]*=[[:space:]]*\).*\$/\1${PROP_VALUE}/" $GRADLE_PROPERTIES_FILE
```
maximilian-galanis, line 17: unfortunately the '' is needed after the -i flag on macOS.
I had a quick talk with chatGPT with suggested following solution
```suggestion
  if [[ "$(uname)" == "Darwin" ]]; then
    sed -i '' "s/\(${PROP_KEY}[[:space:]]*=[[:space:]]*\).*\$/\1${PROP_VALUE}/" $GRADLE_PROPERTIES_FILE
  else
    sed -i "s/\(${PROP_KEY}[[:space:]]*=[[:space:]]*\).*\$/\1${PROP_VALUE}/" $GRADLE_PROPERTIES_FILE
  fi
}
```

PR #1055, opened 2023-01-20
Use `UNREACHABLE` property in `followNextEOG` methods

### LINKED ISSUE #1049: `UNREACHABLE` property is not used when evaluating `followNextEOGEdgesUntilHit`
# Unexpected behavior

The `UNREACHABLE` property of the PropertyEdges is not used in `followNextEOGEdgesUntilHit` to determine whether a path is failed or fulfilled. I think that this information should be used if it is available since we know for certain that a path with an `UNREACHABLE` property is never executed.

KuechA:
We may want to filter unreachable paths from the results of the `followNextEOG` and `followNextEOGUntilHit` methods. Closes #1049 

I realized that we do not set this property for the `prevEOGEdges` even if it should be a bi-directional property. Unreachable basically means that there's a condition which makes this specific transition `n1 - nextEOG -> n2` impossible. This means that also `n1 <- prevEOG - n2` should be not be relevant for the backwards traversal, right? Should we add this feature as well (@oxisto @konradweiss)?

### DIFF
--- a/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/Extensions.kt
+++ b/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/Extensions.kt
@@ -376,13 +376,20 @@ fun Node.followPrevEOGEdgesUntilHit(predicate: (Node) -> Boolean): FulfilledAndF
         val currentPath = worklist.removeFirst()
         // The last node of the path is where we continue. We get all of its outgoing DFG edges and
         // follow them
-        if (currentPath.last().prevEOG.isEmpty()) {
+        if (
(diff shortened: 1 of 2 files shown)


maximiliankaul:
CPG meeting notes: don't forget about `goto` `label`s.

oxisto:
> CPG meeting notes: don't forget about `goto` `label`s.

But in a separate PR

KuechA:
Double-check the EOG edges. They should already be bi-directional

KuechA:
> Double-check the EOG edges. They should already be bi-directional

Ok, the edges are indeed bi-directional. I also added the code to take into account unreachability when backwards-traversing the EOG. However, as discussed, we stop with the traversal a bit too late but fixing this will subject to another PR

oxisto (approved):

PR #1057, opened 2023-01-20
Fixed annotation parsing in python

oxisto:


oxisto:
@maximiliankaul Do we have some one-liner how to fix these annoying style errors in python?

maximiliankaul:
> @maximiliankaul Do we have some one-liner how to fix these annoying style errors in python?

Unfortunately, I'm not aware of a solution that "just works". You could try `autopep8 --in-place --recursive --aggressive src/main/python` (or maybe even with two `--aggressive`. However, the results are of varying quality :(
I assume that most of the annoyances come from the 80 character limit, right? I am in favor of increasing this to 120 or something similar. Would that be helpful?

oxisto:
I am lost with python :D

maximiliankaul (approved):

PR #1061, opened 2023-01-21
Improvements to the `CXXLanguageFrontend`

### LINKED ISSUE #28: ExpressionHandler#expressionTypeProxy should return cpg type class
`ExpressionHandler#expressionTypeProxy` is currently returning a `IType` (CDT type), then depending on the handling function which calls it, the type is parsed completely different. Sometimes it looks for pointers, sometime it does not. The function should directly return a CPG `Type` to reduce the difference in handling this.

### LINKED ISSUE #189: Avoid using CDT type resolving
TODO

oxisto:
This PR adds several improvements to the C/C++ language frontend. Primarily, we remove reliance to CDT types. This is one more step towards our strategy to use CDT solely as parser, but not do any kind of symbol resolving.

Fixes #28
Fixes #189

### DIFF
--- a/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
+++ b/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
@@ -29,6 +29,7 @@ import static de.fraunhofer.aisec.cpg.graph.DeclarationBuilderKt.newTypedefDecla
 
 import de.fraunhofer.aisec.cpg.frontends.Language;
 import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend;
+import de.fraunhofer.aisec.cpg.frontends.cpp.CLanguage;
 import de.fraunhofer.aisec.cpg.graph.declarations.Declaration;
 import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration;
 import de.fraunhofer.aisec.cpg.graph.declarations.TemplateDeclaration;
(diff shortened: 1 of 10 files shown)


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.kt
```diff
@@ -489,45 +410,14 @@ class ExpressionHandler(lang: CXXLanguageFrontend) :
...
-        } else if (proxy !is TypeOfDependentExpression) {
-            declaredReferenceExpression.type =
-                TypeParser.createFrom(expressionTypeProxy(ctx).toString(), true, frontend)
-        }
-
-        /* this expression could actually be a field / member expression, but somehow CDT only recognizes them as a member expression if it has an explicit 'this'
-         */
+        // this expression could actually be a field / member expression, but somehow CDT only
+        // recognizes them as a member expression if it has an explicit 'this'
         // TODO: handle this? convert the declared reference expression into a member expression?
```
konradweiss, line 415: Do we tackle this. Can this IdExpression even be a member expression?

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
```diff
@@ -546,23 +547,35 @@ private Set<Ancestor> getAncestors(RecordDeclaration recordDeclaration, int dept
     return ancestors;
   }
 
-  public boolean isSupertypeOf(Type superType, Type subType, ScopeProvider provider) {
+  public boolean isSupertypeOf(Type superType, Type subType, MetadataProvider provider) {
```
konradweiss, line 550: Is any MetadataProvider allowed and expected here. I not I would either restrict the type or if there is no specific type then name it to something descriptive like `scopeProvider`

konradweiss (changes requested):

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/TypeManager.java
oxisto: Basically, we need a metadata provider that is ideally a language provider and a scope provider. Unfortunately we cannot scope it like that in Java (I think it may be possible in Kotlin), so I chose the basic metadata provider class.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.kt
oxisto: We sort of handle this later in the call resolver, so maybe we could remove the TODO or refer to the call resolver. The problem is that if you are within a class method, any reference to a name can either be a member expression or a simple reference to something outside of the class. For example

```
class Test {
  void test() {
    anotherTest(); // member (call) expression with implicit this
    yetAnotherTest(); // call expression to function outside of class
  }

  void anotherTest() {}
}

void yetAnotherTest().
```

In both cases we model it as a regular `CallExpression` with a `DeclaredReferenceExpression` and then later in the CXX specific call resolver, we handle it like it would have been a `MemberExpression` if we see that it refers to a method.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/statements/expressions/DeclaredReferenceExpression.java
```diff
@@ -138,13 +139,18 @@ public void possibleSubTypesChanged(HasType src, List<HasType> root) {
     if (!TypeManager.isTypeSystemActive()) {
       return;
     }
+
+    // since we want to update the sub types, we need to exclude ourselves from the root, otherwise
+    // it won't work. What a weird and broken system!
+    root.remove(this);
```
oxisto, line 145: @KuechA 

Just FYI, this was one of those weird cases where I had to remove the node from `root`, but hopefully this will be obsolete in the future anyway.

konradweiss (approved):

PR #1062, opened 2023-01-25
make configure_frontends.sh work on darwin and other os

maximiliankaul:


oxisto (approved):
Works on my machine (Mac)

PR #1063, opened 2023-01-25
Also serializing the full name of a Node as `name` in Neo4J

oxisto:
This adds the `name` property back into Neo4J. It contains the `fullName` (aka `toString()`) representation of a `Name`.


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/NameConverter.kt
```diff
@@ -41,12 +41,14 @@ class NameConverter : CompositeAttributeConverter<Name> {
 
     companion object {
         const val FIELD_FULL_NAME = "fullName"
+        const val FIELD_NAME = "name"
```
konradweiss, line 44: Can we have a comment here saying that name is an alias for fullName? Or where is the best place to mention that in code, the member where the name is stored?

konradweiss (approved):
I would consider where we add the aliasing information `name` `fullName`, the rest is fine IMO

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/helpers/NameConverter.kt
oxisto: I added some comments to the persistence function

PR #1068, opened 2023-01-26
Remove the mark directory

KuechA:


oxisto:
You also need to remove the following lines:

https://github.com/Fraunhofer-AISEC/cpg/blob/cf4223e03cb7024c4abeb917dc9f7f81b176ad05/.github/workflows/build.yml#L14-L21

PR #1072, opened 2023-01-29
Conversion of more classes to Kotlin

oxisto:
This PR converts as much code as possible to Kotlin. Some things cannot be converted because of weird type stuff that only works in Java.

I also added some things to the `CONTRIBUTING.md` for code guidelines, which we might move to the docs. https://github.com/Fraunhofer-AISEC/cpg/blob/e33eb543ead5de1606b2a8aafb904dd0fc99d633/CONTRIBUTING.md#code-guidelines


cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/ScopeManager.kt
```diff
@@ -465,9 +465,9 @@ class ScopeManager : ScopeProvider {
             }
             (scope as Breakable).addBreakStatement(breakStatement)
         } else {
-            val labelStatement = getLabelStatement(breakStatement.label)
+            val labelStatement = getLabelStatement(breakStatement.label!!)
```
oxisto, line 468: @KuechA please avoid !!, we cannot guarantee that all of these nodes are not-null

oxisto:
Discussion with regards to required properties: Since we are only constructing nodes with the node builder, we could also just make them required in a constructor. Would that make more sense? I think so? 

cc @KuechA @konradweiss

KuechA:
Yes, I wanted to propose the same thing. Right now, we have hundreds of properties which are nullable and seem to be optional but in reality, they aren't.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/passes/scopes/ScopeManager.kt
KuechA: I added two more checks about the nullable `subStatement`. We check that the label is null in the if-branch, so it should be fine with `!!` here.

oxisto:
> Yes, I wanted to propose the same thing. Right now, we have hundreds of properties which are nullable and seem to be optional but in reality, they aren't.

Perfect. Any preference in dealing it with a constructor vs. initialising it with ProblemExpressions ? Probably we could sort of both, for example make the expressions nullable in the node builder (in case something does not get parsed) and the node builder can then create a problemnode/expression and hand it to the constructor. In this case we would only really create problem nodes if there is an actual problem.

KuechA:
> Perfect. Any preference in dealing it with a constructor vs. initialising it with ProblemExpressions ? Probably we could sort of both, for example make the expressions nullable in the node builder (in case something does not get parsed) and the node builder can then create a problemnode/expression and hand it to the constructor. In this case we would only really create problem nodes if there is an actual problem.

I'm not sure if it makes a real difference but somehow, I'd prefer to make the decision about a ProblemNode in the NodeBuilder. Also, I wouldn't change the NodeBuilder's interface to keep the impact "small" (lol)

oxisto:
> > Perfect. Any preference in dealing it with a constructor vs. initialising it with ProblemExpressions ? Probably we could sort of both, for example make the expressions nullable in the node builder (in case something does not get parsed) and the node builder can then create a problemnode/expression and hand it to the constructor. In this case we would only really create problem nodes if there is an actual problem.
> 
> I'm not sure if it makes a real difference but somehow, I'd prefer to make the decision about a ProblemNode in the NodeBuilder. Also, I wouldn't change the NodeBuilder's interface to keep the impact "small" (lol)

Sounds good. We just have to see whether Neo4j OGM needs some kind of default empty constructor - but as far as I know this gets magically injected in the newer versions.

konradweiss:
> > > Perfect. Any preference in dealing it with a constructor vs. initialising it with ProblemExpressions ? Probably we could sort of both, for example make the expressions nullable in the node builder (in case something does not get parsed) and the node builder can then create a problemnode/expression and hand it to the constructor. In this case we would only really create problem nodes if there is an actual problem.
> > 
> > 
> > I'm not sure if it makes a real difference but somehow, I'd prefer to make the decision about a ProblemNode in the NodeBuilder. Also, I wouldn't change the NodeBuilder's interface to keep the impact "small" (lol)
> 
> Sounds good. We just have to see whether Neo4j OGM needs some kind of default empty constructor for nodes - but as far as I know this gets magically injected in the newer versions.

I do like the idea, I am sceptical on the amount of properties that may not be set though, however, with the refactoring we also get some clarity on which once are currently not set in specific cases and can make a decision on whether or not we want to change it.

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/declarations/ClassTemplateDeclaration.kt
```diff
@@ -0,0 +1,96 @@
...
+/** Node representing a declaration of a ClassTemplate */
+class ClassTemplateDeclaration : TemplateDeclaration() {
+    /**
+     * Edges pointing to all RecordDeclarations that are realized by the ClassTempalte. Before the
+     * expansion pass there is only a single RecordDeclaration which is instantiated after the
+     * expansion pass for each instantiation of the ClassTemplate there will be a realization
+     */
+    @Relationship(value = "REALIZATION", direction = Relationship.Direction.OUTGOING)
+    @field:SubGraph("AST")
+    private val _realization: MutableList<PropertyEdge<RecordDeclaration>> = ArrayList()
```
oxisto, line 46: Remove the `_´ style accessors

cpg-core/src/main/java/de/fraunhofer/aisec/cpg/graph/declarations/EnumDeclaration.kt
```diff
@@ -35,32 +35,32 @@ import org.neo4j.ogm.annotation.Relationship
...
 
     @Relationship(value = "SUPER_TYPES", direction = Relationship.Direction.OUTGOING)
-    var superTypesPropertyEdge: List<PropertyEdge<Type>> = ArrayList()
+    var superTypeEdges: List<PropertyEdge<Type>> = ArrayList()
         private set
 
     @Relationship var superTypeDeclarations: Set<RecordDeclaration> = HashSet()
 
     @property:SubGraph("AST")
     var entries: List<EnumConstantDeclaration>
```
oxisto, line 49:
```suggestion
    var entries by PropertyEdgeDelegate(EnumDeclaration::entryEdges)
```

then you can get rid of the get/setter

KuechA (approved):

Issue #1065, opened 2023-01-26
Move DFG edges from EOG pass to DFG pass


maximiliankaul:
The EOG pass should not create DFG edges.

oxisto:
Closed by #1074

PR #1074, opened 2023-01-30
Moving remaining DFG edge creation from EOG pass to DFG pass.

konradweiss:
Moving DFG edges that show relation between data and branching nodes from the EOG pass that models branching to the DFG pass as the responsible for DFG edges.

This does not change the graphs structure and only refactors DFG edge creation to reduce the fragmentation of that code semantic.

### DIFF
--- a/cpg-core/specifications/dfg.md
+++ b/cpg-core/specifications/dfg.md
@@ -437,7 +437,7 @@ Scheme:
 ### WhileStatement
 
 Interesting fields:
-* `condition: Statement`: The condition that is evaluated before making the branching decisio
+* `condition: Statement`: The condition that is evaluated before making the branching decision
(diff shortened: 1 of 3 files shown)


cpg-core/specifications/dfg.md
```diff
@@ -414,8 +417,84 @@ Scheme:
...
+  flowchart LR
+    node([DoStatement]) -.- condition(condition)
+    condition -- DFG --> node
+  ```
+  
+
+### WhileStatement
+
+Interesting fields:
+* `condition: Statement`: The condition that is evaluated before making the branching decisio
```
KuechA, line 440:
```suggestion
* `condition: Statement`: The condition that is evaluated before making the branching decision
```

KuechA:
Closes #1065

KuechA (approved):

PR #1078, opened 2023-02-01
Correct parsing of functions in namespaces in C++

### LINKED ISSUE #1070: fullName property is incomplete when using namespaces
# Unexpected behavior

The namespace is missing in the `fullName` property of the type node.
For c++, the default namespace within a function definition is the namespace of function itself.
The cpg does not take this into account and creates a new type, outside the namespace.

image

# Code of interest

```c++
namespace ABC {
        struct A {
        };
        void foo();
        void bar(ABC::A a);
};

void ABC::foo() {
        A a;
        bar(a);
}

void ABC::bar(ABC::A a) {
}

```

None

None

oxisto:
Furthermore, this tries to mitigate the problem that types used within such a function could refer to a namespaced entity but are referred with its local name. Since this requires resolving of symbols in a frontend, we introduce a new annotaiton `@ResolveInFrontend` to warn developers.

Fixes #1070

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.kt
@@ -33,10 +33,10 @@ import de.fraunhofer.aisec.cpg.graph.statements.CompoundStatement
 import de.fraunhofer.aisec.cpg.graph.statements.ReturnStatement
 import de.fraunhofer.aisec.cpg.graph.statements.Statement
 import de.fraunhofer.aisec.cpg.graph.types.*
+import java.util.function.Supplier
 import org.eclipse.cdt.core.dom.ast.*
 import org.eclipse.cdt.core.dom.ast.IASTTranslationUnit.IDependencyTree.IASTInclusionNode
 import org.eclipse.cdt.internal.core.dom.parser.cpp.*
-import java.util.function.Supplier
 
 /**
  * This class is a [CXXHandler] which takes care of translating C/C++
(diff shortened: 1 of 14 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.kt
```diff
@@ -213,11 +215,12 @@ class DeclarationHandler(lang: CXXLanguageFrontend) :
 
         frontend.processAttributes(declaration, ctx)
 
-        frontend.scopeManager.leaveScope(declaration)
-        if (recordDeclaration != null && outsideOfRecord) {
-            frontend.scopeManager.leaveScope(recordDeclaration)
+        if (holder != null && outsideOfScope) {
```
konradweiss, line 218: Why are we leaving the scope of the holder before the scope of the declaration? We enter them in order ``holder -> declaration` so we should leave them in the order `declaration -> holder` right? Maybe I am missing something. In this case it would not make a difference though, I think.

konradweiss (changes requested):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.kt
oxisto: You are correct. This was probably a problem by moving code fragments around.

oxisto: Reverted back to the previous correct way.

oxisto:
@konradweiss Once this is merged, a 6.1 release would be nice

konradweiss (approved):

PR #1080, opened 2023-02-01
Improvements to Neo4J OGM

oxisto:
This PR enables proper serialization of the `Scope` objects. To enable this, we moved the scopes from `.passes.scopes` to `.graph`, which is a better place for them anyway.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/ValueDeclarationScope.kt
```diff
@@ -40,10 +40,10 @@ import org.slf4j.LoggerFactory
  * Works for if, for, and extends to the block scope
  */
 open class ValueDeclarationScope(override var astNode: Node?) : Scope(astNode) {
-    var valueDeclarations = mutableListOf<ValueDeclaration>()
+    @Transient var valueDeclarations = mutableListOf<ValueDeclaration>()
```
konradweiss, line 43: Why are we not persisting these? Just to catch me up on what in the scopes we want to store in a graph and what not

konradweiss (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/ValueDeclarationScope.kt
oxisto: Ok, we could actually persist those, I just didn't want to add too many edges. The other transient properties are Maps, which cannot be persisted without major pain.

Do you want me to remove `@Transient` from the declarations list in `ValueDeclarationScope` and `StructureDeclarationScope`?

konradweiss: No, if the assesment is that thats to many edges to persist into the neo4j I am fine with that. we can change our strategy in the future if we want.

PR #1086, opened 2023-02-03
Consider all children to find the `prevDFG` for `FunctionDeclaration`s

### LINKED ISSUE #1077: only one has the DFG edge to its FunctionDeclaration when several ReturnStatements are in the same function
# Unexpected behavior
If several ReturnStatements in a function, it will be only one ReturnStatement has the DFG edge to its FunctionDeclaration. In the following example, the ReturnStatement which has the DFG edge is a virtual return node. I am not sure if this is a problem.

# Example
```java
public class Test {
    public int func() {
        int a = 1;
        if (a == 1) {
            return 1; // ReturnStatement 1, no DFG edge
        } else {
            return 2; // ReturnStatement 2, no DFG edge
        }
    }
}
```

# Configuration
defaultPass() with v6.0.0

KuechA:
I forgot to traverse the AST of function declarations to identify all possible return statements in the `DFGPass`. However, there is still an implicit return statement in the function even if it's not necessary. (@konradweiss Can we get rid of it?)

Fixes #1077

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/StatementHandler.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/StatementHandler.kt
@@ -97,6 +97,7 @@ class StatementHandler(lang: JavaLanguageFrontend?) :
                     as? de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
         }
         val returnStatement = this.newReturnStatement(returnStmt.toString())
+        returnStatement.isImplicit = !returnStmt.tokenRange.isPresent
 
         // expressionRefersToDeclaration to arguments, if there are any
         expression?.let { returnStatement.returnValue = it }
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/StatementHandler.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/StatementHandler.kt
@@ -97,7 +97,8 @@ class StatementHandler(lang: JavaLanguageFrontend?) :
                     as? de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
         }
         val returnStatement = this.newReturnStatement(returnStmt.toString())
-        // JavaParser seems to add implicit return statements, that are not part of the original source code. We mark it as such
+        // JavaParser seems to add implicit return statements, that are not part of the original
+        // source code. We mark it as such
         returnStatement.isImplicit = !returnStmt.tokenRange.isPresent
 
         // expressionRefersToDeclaration to arguments, if there are any
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/StatementHandler.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/StatementHandler.kt
@@ -97,6 +97,7 @@ class StatementHandler(lang: JavaLanguageFrontend?) :
                     as? de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
         }
         val returnStatement = this.newReturnStatement(returnStmt.toString())
+        // JavaParser seems to add implicit return statements, that are not part of the original source code. We mark it as such
         returnStatement.isImplicit = !returnStmt.tokenRange.isPresent
 
         // expressionRefersToDeclaration to arguments, if there are any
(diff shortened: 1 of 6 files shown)


cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/DFGTest.kt
```diff
@@ -0,0 +1,65 @@
...
+        val result =
+            TestUtils.analyze(
+                listOf(Path.of(topLevel.toString(), "ReturnTest.java").toFile()),
+                topLevel,
+                true
+            )
+        val returnFunction = result.functions["testReturn"]
+        assertNotNull(returnFunction)
+
+        // TODO: For some reason, there's the "implicit" return statement even if the if and the
```
oxisto, line 52: I think the "implicit" return logic is rather stupid and only checks if the *last* statement of the compound statement is a return statement. we should probably filter for all returns first and then check if its necessary to include it.

oxisto: Here is the CPP code:
https://github.com/Fraunhofer-AISEC/cpg/blob/67522e366e549864a60f0c1b841f4c8a97a97bd6/cpg-core/src/main/java/de/fraunhofer/aisec/cpg/frontends/cpp/DeclarationHandler.kt#L199-L212

oxisto: However, I can't see similar code for the java frontend (which you are using).

oxisto: Interesting. It seems that the JavaParser is inserting this `return`?!. If you look at the code of the `BlockStmt` given by JavaParser it is:
```java
{
    int a = 1;
    if (a == 5) {
        return 2;
    } else {
        return a;
    }
    return;
}
```

Even though the return is not part of the original source code.

KuechA: yeah, I was a bit confused because I didn't find the java-related code. But even if we have these statements, we do not have the EOG when inserting the return statement into the graph. So, I'm not sure if there's a good way to avoid adding these statements in the frontends.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -225,6 +229,27 @@ open class ControlFlowSensitiveDFGPass : Pass() {
...
+    private fun removeUnreachableImplicitReturnStatement(
+        node: Node,
+        reachableReturnStatements: MutableSet<ReturnStatement>
+    ) {
+        val lastStatement =
+            ((node as? FunctionDeclaration)?.body as? CompoundStatement)?.statements?.lastOrNull()
+        if (
+            lastStatement is ReturnStatement &&
+                lastStatement.location ==
+                    null && // Location == null means that it's an "implicit" return statement
```
oxisto, line 249: I would prefer setting the isImplicit flag in the java frontend instead.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/StatementHandler.kt
```diff
@@ -97,6 +97,7 @@ class StatementHandler(lang: JavaLanguageFrontend?) :
                     as? de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
         }
         val returnStatement = this.newReturnStatement(returnStmt.toString())
+        returnStatement.isImplicit = !returnStmt.tokenRange.isPresent
```
oxisto, line 100:
```suggestion
        // JavaParser seems to add implicit return statements, that are not part of the original source code. We mark it as such
        returnStatement.isImplicit = !returnStmt.tokenRange.isPresent
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -225,6 +229,26 @@ open class ControlFlowSensitiveDFGPass : Pass() {
                     }
             }
         }
+
+        removeUnreachableImplicitReturnStatement(node, returnStatements)
+    }
+
+    /**
+     * Removes the DFG edges for a potential implicit return statement if it is not
+     * in[reachableReturnStatements].
```
oxisto, line 238:
```suggestion
     * in [reachableReturnStatements].
```

oxisto (approved):

PR #1087, opened 2023-02-03
Extract the Java Frontend to a submodule

KuechA:
To separate the languages supported by the cpg and the actual "core", i.e., the model and the passes, we can also extract the java language frontend into an own module `cpg-language-java`. Unfortunately, many tests rely on java files, so I also moved many of the tests to this module. However, we should check which of the tests can and should be generalized and run independent from a language but on a manually assembled graph.


oxisto:
I tried to build a fluent DSL to construct nodes, e.g. for testing in https://github.com/Fraunhofer-AISEC/cpg/pull/772. I abandoned it because we had other issues, but I can revisit this. I had the same idea that some (all?) of the tests in core should instead be run on a generic set of nodes instead of nodes parsed by a frontend.

konradweiss:
I did not know that we want to extract the frontend, the initial intention was to extract optional languages to keep them separate, but I like that we make this move. Some of the tests should already be language independent: the EOG passes. I am curious if they break too.

konradweiss:
I did not know that we want to extract the frontend, the initial intention was to extract optional languages to keep them separate, but I like that we make this move. Some of the tests should already be language independent: the EOG passes. I am curious if they break too.

KuechA:
> I did not know that we want to extract the frontend, the initial intention was to extract optional languages to keep them separate, but I like that we make this move. Some of the tests should already be language independent: the EOG passes. I am curious if they break too.

Yes, I'd suggest to have all languages and frontends in separate submodules. Currently, a couple of passes still use `CLanguage` or so to check about some properties. By moving all this to separate modules, we can enforce that this will never happen again ;)

Concerning the tests, we will still have to model the respective graph with the DSL but it turns out that it's not that easy in all cases. This will happen in other PRs where we can also figure out the purpose of each of the tests because that's not always clear to me to be honest.

build.gradle.kts
```diff
@@ -65,6 +65,12 @@ sonarqube {
 // Load the properties that define which frontends to include
 //
 // this code block also exists in settings.gradle.kts
+val enableJavaFrontend by extra {
+    val enableJavaFrontend: String by project
+    enableJavaFrontend.toBoolean()
+}
+project.logger.lifecycle("Go frontend is ${if (enableJavaFrontend) "enabled" else "disabled"}")
```
oxisto, line 72:
```suggestion
project.logger.lifecycle("Java frontend is ${if (enableJavaFrontend) "enabled" else "disabled"}")
```

cpg-analysis/build.gradle.kts
```diff
@@ -42,6 +42,7 @@ publishing {
 
 dependencies {
     api(projects.cpgCore)
+    api(projects.cpgLanguageJava)
```
oxisto, line 45: is that intentional? or do you only want to include it for the tests?

cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/SizeEvaluatorTest.kt
```diff
@@ -49,7 +51,10 @@ class SizeEvaluatorTest {
                 listOf(topLevel.resolve("size.java").toFile()),
                 topLevel,
                 true
-            )
+            ) {
+                it.registerLanguage(JavaLanguage())
+                    .registerPass(JavaExternalTypeHierarchyResolver())
```
oxisto, line 56: shouldn't this pass automatically be registered with the language?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageFrontend.kt
```diff
@@ -89,6 +90,14 @@ abstract class LanguageFrontend(
         return init(this)
     }
 
+    /**
+     * Similar to [parse], this function returns a [TranslationResult], but rather than parsing
+     * source code, the function [init] is used to build nodes in the Node Fluent DSL.
+     */
+    fun buildTR(init: LanguageFrontend.() -> TranslationResult): TranslationResult {
```
oxisto, line 97: should we instead return a translation result in `build`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
```diff
@@ -50,6 +67,86 @@ fun LanguageFrontend.translationUnit(
...
+
+    return node
+}
+
+/**
+ * Creates a new [TranslationUnitDeclaration] in the Fluent Node DSL with the given [name]. The
+ * declaration will be set to the [ScopeManager.globalScope]. The [init] block can be used to create
+ * further sub-nodes as well as configuring the created node itself.
+ */
+fun LanguageFrontend.translationUnit(
```
oxisto, line 80: Why do we have this twice now?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
```diff
@@ -222,6 +362,72 @@ fun LanguageFrontend.call(
     return node
 }
 
+/**
```
oxisto, line 365: copy/paste docu?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
```diff
@@ -222,6 +362,72 @@ fun LanguageFrontend.call(
...
+): Expression {
+    val node =
+        if (name.parent != null) {
+            newMemberExpression(name.localName, memberOrRef(name.parent))
+        } else {
+            newDeclaredReferenceExpression(name.localName, parseType(name.localName))
+        }
+    node.type = type
+
+    node.isStaticAccess = name.localName.first().isUpperCase()
```
oxisto, line 427: What's the logic behind that?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ImportResolver.kt
```diff
@@ -38,7 +38,7 @@ import java.util.*
 import java.util.regex.Pattern
 
 @DependsOn(TypeHierarchyResolver::class)
-@DependsOn(value = JavaExternalTypeHierarchyResolver::class, softDependency = true)
+// @DependsOn(value = JavaExternalTypeHierarchyResolver::class, softDependency = true)
```
oxisto, line 41: Remove commented out line

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/GraphExamples.kt
```diff
@@ -0,0 +1,222 @@
...
+import de.fraunhofer.aisec.cpg.frontends.TestLanguage
+import de.fraunhofer.aisec.cpg.frontends.TestLanguageFrontend
+import de.fraunhofer.aisec.cpg.graph.builder.*
+import de.fraunhofer.aisec.cpg.graph.newVariableDeclaration
+import java.io.File
+import java.util.function.Consumer
+
+class GraphExamples {
+    companion object {
+        fun getTRWithConfig(
```
oxisto, line 37: This is extremely weird to use. There is no way to know which paths will work here and the name of this function is weird as well.

cpg-core/src/testFixtures/kotlin/de/fraunhofer/aisec/cpg/frontends/TestLanguage.kt
```diff
@@ -39,11 +39,15 @@ import kotlin.reflect.KClass
  * This is a test language that can be used for unit test, where we need a language but do not have
  * a specific one.
  */
-class TestLanguage : Language<TestLanguageFrontend>() {
+class TestLanguage(namespaceDelimiter: String) : Language<TestLanguageFrontend>() {
```
oxisto, line 42: Can we make `::` the default at least? So we can just use an empty constructor, if wanted.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
KuechA: One of them requires the `TranslationResult` as a context whereas this one doesn't. It's related to `build`vs `buildTR` from above. If we always return the translation result in `build`, we can get rid of this method.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageFrontend.kt
KuechA: I think it would make sense to change `build`

cpg-analysis/build.gradle.kts
KuechA: It should only be required for the tests, so I changed it to `testImplementation()`.

cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/SizeEvaluatorTest.kt
KuechA: No, you have to register it. The annotation `RequiredFrontend` means that the pass won't be executed if the `JavaLanguageFrontend` isn't registered.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
KuechA: `isStaticAccess` is somehow used to avoid resolving a variable for a `DeclaredReferenceExpression`. E.g., the `VariableUsageResolver` adds a field `System` to the enclosing TR if there's a call `System.out.println` (or similar). Setting this variable avoids it and won't resolve the symbol. No idea if this is always intended and if there's a better way to do this but I would be more than happy to have an alternative (and clearly documented) solution

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/GraphExamples.kt
KuechA: In theory, it should be able to get rid of the `analyze` methods but to be honest, I'm not even sure if it works for all cases :D

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
oxisto: Yes but in this case you should just set isStaticAccess in the node's init function (for `System...`) instead of relying on this weird logic here; because for example in Go upper and lowercase variable names do have semantics, but definitely not static/non-static.

cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/SizeEvaluatorTest.kt
oxisto: Ah ok. I thought we added an annotation that supports automatic registering of a "default" pass once a language is registered. If not, we should probably have it :) 

cc @maximiliankaul

maximiliankaul: A frontend can require passes (and passes can require other passes).
The `JavaLanguageFrontend` could be annotated with `RegisterExtraPass` which would make it always execute the `JavaExternalTypeHierarchyResolver`. Is this pass a pass which should always be executed (for Java)?

oxisto: > A frontend can require passes (and passes can require other passes).
> The `JavaLanguageFrontend` could be annotated with `RegisterExtraPass` which would make it always execute the `JavaExternalTypeHierarchyResolver`. Is this pass a pass which should always be executed (for Java)?

Yes, it is.

oxisto (approved):

oxisto:
@konradweiss Can we have another `v6` release with current main changes before we merge this in?

PR #1088, opened 2023-02-08
Even faster Control Flow Sensitive DFG Pass

KuechA:
We remove all declarations from the map which are no longer relevant in the scope of the next hop. This should remove quite some cycles in the pass if variables are declared in an "inner scope" which don't live outside that scope


konradweiss (approved):
LGTM

PR #1095, opened 2023-02-14
Initial support for java and cpp lambdas

### LINKED ISSUE #10: Support anonymous functions in the Graph
Something like Java lambda functions or C++ lambda expressions, such as:
```
SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, [](int preverify_ok, X509_STORE_CTX *x509_ctx) {
    return 1;
  });
```

- [x] Create a class representing the graph node. Extend from FunctionDeclaration? Is it a FunctionDeclaration without name?
- [x] Assign a handler in CDT to parse it
- [x] Assign a handler in JavaParser to parse it
- [ ] Adjust CFG passes to handle it

### LINKED ISSUE #1005: Anonymous classes in Java not translated to RecordDeclaration
# Unexpected behavior

When analyzing C++ code, an anonymous class is translated into a RecordDeclaration.
However, when analyzing Java code with an anonymous class, the anonymous class is only translated into a DeclarationStatement but not further into a RecordDeclaration. This means that implemented or overridden methods in the anonymous class cannot be analyzed.

# Example

```
interface HelloWorld {
	public void greet();
}

class Anonymous {

	public static void main() {
		HelloWorld frenchGreeting = new HelloWorld() {
			public void greet() {
				System.out.println("tout le monde");
			}
		};

		frenchGreeting.greet();
	}
}
```

```
            .defaultPasses()
            .defaultLanguages()
            .codeInNodes(true)
```

KuechA:
We already have a `LambdaExpression` node but neither java nor cpp have handlers to assemble it. This PR adds initial support to handle lambdas.

Tasks:
* [x] Implement Java handler
* [x] Implement CPP handler
* [x] Connect the EOG => Construct the eog inside the lambda's function. The "outer" EOG will just continue normally.
* [x] Cpp Call resolution like any other function pointer call.
* [x] Test variable resolving
* [x] Support anonymous classes too (optional, if easy)

Closes #10 , closes #1005


konradweiss:
Apparently I broke some tests, already on it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/ExpressionHandler.kt
```diff
@@ -72,12 +74,60 @@ class ExpressionHandler(lang: CXXLanguageFrontend) :
...
+        // either make the function "mutable" or pass everything by reference.
+        lambda.areVariablesMutable =
+            (node.declarator as? CPPASTFunctionDeclarator)?.isMutable == true ||
+                node.captureDefault == ICPPASTLambdaExpression.CaptureDefault.BY_REFERENCE
+
+        val anonymousFunction =
+            frontend.declaratorHandler.handle(node.declarator) as? FunctionDeclaration
+                ?: return lambda
+
+        if (node.declarator.trailingReturnType != null) {
```
oxisto, line 117: Note to self: trailingReturnType should be parsed in the declaration handler

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/enhancements/EOGTest.kt
```diff
@@ -1203,6 +1203,78 @@ internal class EOGTest : BaseTest() {
...
+            "std::cout",
+            (lambda.function?.nextEOG?.get(0) as? DeclaredReferenceExpression)?.name.toString()
+        )
+
+        val cout = lambda.function?.nextEOG?.get(0) as? DeclaredReferenceExpression
+        assertEquals(1, cout?.nextEOG?.size)
+        assertEquals("Hello ", (cout?.nextEOG?.get(0) as? Literal<*>)?.value.toString())
+
+        val hello = cout?.nextEOG?.get(0) as? Literal<*>
+        assertEquals(1, hello?.nextEOG?.size)
```
oxisto, line 1246: Inserting a couple of assertNotNull would be good to drastically remove the amount of `?` in the code

konradweiss (approved):

cpg-core/src/test/resources/lambda.cpp
```diff
@@ -0,0 +1,5 @@
+int main() {
```
oxisto, line 1: somehow we have two different lambda and lambdas.cpp and `testLambdas`in CXXLanguagueFrontend and some other test.

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/LambdaTest.kt
```diff
@@ -0,0 +1,447 @@
...
+import de.fraunhofer.aisec.cpg.graph.*
+import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration
+import de.fraunhofer.aisec.cpg.graph.declarations.VariableDeclaration
+import de.fraunhofer.aisec.cpg.graph.statements.CompoundStatement
+import de.fraunhofer.aisec.cpg.graph.statements.ReturnStatement
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.*
+import java.io.File
+import kotlin.test.*
+
+class LambdaTest {
```
oxisto, line 39: This should probably be split into two different test classes for cpp and java, since we want to separate those language frontends. Also this should probably be in the frontend package.

oxisto (changes requested):

KuechA:
@peckto will test this PR with more real-world samples

oxisto (approved):

PR #1098, opened 2023-02-16
Add `ArrayRangeExpressions` to the EOGPass

### LINKED ISSUE #377: EOG should be made aware of ArrayRangeExpression
@oxisto: EOG should be made aware of this expression.

_Originally posted by @maximiliankaul in https://github.com/Fraunhofer-AISEC/cpg/discussions/372#discussioncomment-530081_

KuechA:
Adds handling `ArrayRangeExpression`s to the EOG pass and also adds a `step` property to the class. However, I'm not sure if all languages (i.e., c++) use the expressions in a dynamic way. In that case, I'm not even sure how/if this would be modeled in the EOG.

I'd also suggest to rename the class from `ArrayRangeExpression` to `RangeExpression` since we do not really care about the array here (there's not even a way to specify the array).

Closes #377 (and is related to #372).


konradweiss:
> Nice!

Ah wait its still a Draft, @KuechA Is something missing?

PR #1103, opened 2023-02-22
Resolve relative include path when loading compilation database

peckto:
Interpret the include paths in the compilation database as relative to the specified directory (as it is for the source file already).

Ref: https://clang.llvm.org/docs/JSONCompilationDatabase.html

> directory: The working directory of the compilation. All paths specified in the command or file fields must be either absolute or relative to this directory.

main.c
```c
#include "header.h"

int main() {
    foo();
    return 0;
}
```

compile_commands.json:
```json
[
  {
    "arguments": [
      "/usr/bin/gcc",
      "-c",
      "-Iinc",
      "src/main.c"
    ],
    "directory": "/code/",
    "file": "src/main.c"
  }
]
```

Folder structure:
```
/code/
├── compile_commands.json
├── inc
│   └── header.h
└── src
    └── main.c
```

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/CompilationDatabase.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/CompilationDatabase.kt
@@ -34,6 +34,7 @@ import java.io.File
 import java.nio.file.Paths
 import java.util.*
 import kotlin.collections.ArrayList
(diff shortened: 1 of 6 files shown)


oxisto (commented):
Can you remove the binary main.o? Otherwise its probably ok

peckto:
> Can you remove the binary main.o? Otherwise its probably ok

ups, fixed now

oxisto (approved):

PR #1107, opened 2023-02-24
Merge all include paths from compilation database for unity build mode

peckto:
When using the unity build mode, all source files are included in a temporary main file.
When using this option together with a compilation database, we need to merge the include paths from all source files in the compilation database and assign them to the temporary main file.
This of course losses the possibility of setting include paths for individual translation units.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/CompilationDatabase.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/CompilationDatabase.kt
@@ -62,10 +62,21 @@ class CompilationDatabase : ArrayList<CompilationDatabase.CompilationDatabaseEnt
             return includePaths.keys.toList()
         }
 
+    fun addIncludePath(srcFile: File, paths: List<String>) {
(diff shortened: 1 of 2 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/CompilationDatabase.kt
```diff
@@ -62,10 +62,19 @@ class CompilationDatabase : ArrayList<CompilationDatabase.CompilationDatabaseEnt
...
+        includePaths[srcFile] = paths
+    }
+
     /** Returns the include paths for the specified file. */
     fun getIncludePaths(file: File): List<String>? {
         return includePaths[file]
     }
+
+    /** Returns the include paths for all files in compilation database. */
+    fun getIncludePaths(): List<String> {
```
oxisto, line 75: Can we rename this to `getAllIncludePaths` to not clash with the `includePaths` property. Also this can be converted to a Kotlin property:
```kotlin
val allIncludePaths: List<String> {
get() {
  return includePaths.values.flatten()
}
```

PR #1110, opened 2023-02-28
Fix problem with PropertyEdges in EnumDeclaration

### LINKED ISSUE #1106: EnumDeclaration causes StackOverflowError in neo4j OGM
# Stack trace

```
java.lang.IllegalArgumentException: Infinite recursion (StackOverflowError) (through reference chain: de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration["scope"]->de.fraunhofer.aisec.cpg.graph.scopes.GlobalScope["astNode"]->de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration["scope"]->de.fraunhofer.aisec.cpg.graph.scopes.GlobalScope["astNode"]->de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration["scope"]->de.fraunhofer.aisec.cpg.graph.scopes.GlobalScope["astNode"]->de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration["scope"]->de.fraunhofer.aisec.cpg.graph.scopes.GlobalScope["astNode"]->de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration["scope"]->[...]
```

# Unexpected behavior

When loading any `EnumDeclaration` into the neo4j database, the cpg crashes.

This bug seems to be introduced during converting java classes to kotlin.
Specifically: https://github.com/Fraunhofer-AISEC/cpg/pull/1072/commits/ded84a793e02099996eaacd77bef30ef0de96170
When exchanging the `EnumDeclaration.kt` by `EnumDeclaration.java`, the `EnumDeclaration` can be imported without error.

For me, the java and kotlin code looks identically. 
I have no clue, why this breaks.
Does serialization work different for java then for kotlin classes?
In this case, similar errors could  have been also introduced in other node types which have been converted.

# Code of interest

```c
enum MY_ENUM {
   A,
   B,
   C
};
```

None

None

KuechA:
OGM seems to require `MutableLists` or it will throw a stackoverflow exception (which is very misleading).  I also reviewed the other occurrences of relationships to ensure that we fix this problem everywhere.

In addition, I fixed the inconsistent usage of the `SubGraph` annotation. Annotating properties (or even methods) works but isn't supported by the `SubgraphWalker` (yet).

Closes #1106

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CLanguage.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CLanguage.kt
@@ -31,6 +31,7 @@ import de.fraunhofer.aisec.cpg.TranslationConfiguration
 import de.fraunhofer.aisec.cpg.frontends.*
 import de.fraunhofer.aisec.cpg.graph.types.*
 import kotlin.reflect.KClass
+import org.neo4j.ogm.annotation.Transient
 
 /** The C language. */
 open class CLanguage :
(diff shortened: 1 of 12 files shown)


konradweiss (approved):
We should keep the task in mind to check the other property edge lists for that bug if we don't check for them in this PR.

oxisto:
> We should keep the task in mind to check the other property edge lists for that bug if we don't check for them in this PR.

I just did a quick "find in files" for a non-mutable list and could not find any more, so seems to be ok for now.

PR #1112, opened 2023-03-02
Convert types and utilities to Kotlin

KuechA:
After this PR, almost no java files should be left. Only obsolete code which will be removed very soon should be left.

Files left:
* [x] package `de.fraunhofer.aisec.cpg.processing`
* [x] `TriConsumer`


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/FunctionPointerType.kt
```diff
@@ -0,0 +1,120 @@
...
+import java.util.*
+import org.neo4j.ogm.annotation.Relationship
+
+/**
+ * FunctionPointerType represents FunctionPointers in CPP containing a list of parameters and a
+ * return type.
+ */
+class FunctionPointerType : Type {
+    @Relationship(value = "PARAMETERS", direction = Relationship.Direction.OUTGOING)
+    var parametersPropertyEdge: MutableList<PropertyEdge<Type>> = mutableListOf()
```
oxisto, line 45: While we are at it: Do we really need property edges here? Personally, I would try to reduce the number of property edges to cases where they really make sense and not just introduce them for the sake of having them. Here, I do not see a use case for them.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/ObjectType.kt
```diff
@@ -0,0 +1,162 @@
...
+import de.fraunhofer.aisec.cpg.graph.types.UnknownType.Companion.getUnknownType
+import java.util.*
+import org.neo4j.ogm.annotation.Relationship
+
+/**
+ * This is the main type in the Type system. ObjectTypes describe objects, as instances of a class.
+ * This also includes primitive data types.
+ */
+open class ObjectType : Type, SecondaryTypeEdge {
+    override fun updateType(typeState: Collection<Type>) {
```
oxisto, line 47: functions should be moved below variables

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/ObjectType.kt
```diff
@@ -0,0 +1,162 @@
...
+                propertyEdge.end = newType
+            }
+        }
+    }
+
+    // Reference from the ObjectType to its class (RecordDeclaration) only if the class is available
+    var recordDeclaration: RecordDeclaration? = null
+
+    @Relationship(value = "GENERICS", direction = Relationship.Direction.OUTGOING)
+    var genericsPropertyEdges: MutableList<PropertyEdge<Type>> = mutableListOf()
```
oxisto, line 70: also here, I do not see any good reason to have a property edge list

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/PointerType.kt
```diff
@@ -0,0 +1,138 @@
...
+        if (pointerOrigin == PointerOrigin.ARRAY) {
+            name = elementType.name.append("[]")
+        } else {
+            name = elementType.name.append("*")
+        }
+        this.pointerOrigin = pointerOrigin
+        this.elementType = elementType
+    }
+
+    constructor(type: Type?, elementType: Type, pointerOrigin: PointerOrigin?) : super(type!!) {
```
oxisto, line 60: probably needs a default constructor or one with default arguments here, for neo4j

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/ReferenceType.kt
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * ReferenceTypes describe CPP References (int&amp;), which represent an alternative name for a
+ * variable. It is necessary to make this distinction, and not just rely on the original type as it
+ * is required for matching parameters in function arguments to discover which implementation is
+ * called.
+ */
+class ReferenceType : Type, SecondOrderType {
+    private var reference: Type? = null
+
+    private constructor()
```
oxisto, line 40: is that enough for neo4j?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/ReferenceType.kt
```diff
@@ -0,0 +1,107 @@
...
+        if (other !is ReferenceType) return false
+        return if (!super.equals(other)) false else reference == other.reference
+    }
+
+    override fun hashCode(): Int {
+        return Objects.hash(super.hashCode(), reference)
+    }
+
+    override fun toString(): String {
+        return ("ReferenceType{" +
```
oxisto, line 97: any particular reason not to use our string builder?

cpg-language-llvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/llvm/ExpressionHandler.kt
```diff
@@ -531,7 +531,7 @@ class ExpressionHandler(lang: LLVMIRLanguageFrontend) :
                     newMemberExpression(
                         fieldName,
                         base,
-                        field?.type ?: UnknownType.getUnknownType(),
+                        field?.type ?: UnknownType.unknownType,
```
oxisto, line 534: include language in unknown type?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/FunctionPointerType.kt
konradweiss: Here I have to opposit opinion so I am not sure what to do, as far as I know the save the order of the parameter s which is an information that would get lost in Neo4j as we don't have the list index there. Would it be acceptable to keep them?

I am afraid we cannot solve that with a converter, we tried that before and there was no way to get the list and add properties for the single elements.

KuechA: In general, we should look at all the usages to identify their usages and eliminate unnecessary ones and implement the respective use-case for the edges we keep in the same PR. Right now, we have them almost everywhere and never make use of the properties which is quite useless. So let's say this is a problem of "very near future us"

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/ObjectType.kt
oxisto: Got it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/FunctionPointerType.kt
oxisto: Ah I forgot about that :( Ok in this case we need to keep it like that. Maybe I'll have another look if this is somehow possible without introducing property edges (for the use case where its just the index).

oxisto (approved):

oxisto:
I have added a `newUnknownType` function, the rest looks good. Merging now.

PR #1113, opened 2023-03-03
Improvement of built-in types for all supported languages

oxisto:
This PR is a cleanup and improvement for the built-in types for all supported languages

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
@@ -67,7 +67,7 @@ abstract class Language<T : LanguageFrontend> : Node() {
     @Transient
     open val simpleTypes: Map<String, Type> =
         mapOf(
-            "boolean" to IntegerType("boolean", 1, this, NumericType.Modifier.SIGNED),
+            "boolean" to BooleanType("boolean", 1, this),
             "char" to IntegerType("char", 8, this, NumericType.Modifier.NOT_APPLICABLE),
             "byte" to IntegerType("byte", 8, this, NumericType.Modifier.SIGNED),
             "short" to IntegerType("short", 16, this, NumericType.Modifier.SIGNED),
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
@@ -36,7 +36,6 @@ import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.types.*
 import java.io.File
 import kotlin.reflect.KClass
-import org.neo4j.ogm.annotation.Transient
 
 /**
  * Represents a programming language. When creating new languages in the CPG, one must derive custom
@@ -57,25 +56,13 @@ abstract class Language<T : LanguageFrontend> : Node() {
     /** The class of the frontend which is used to parse files of this language. */
     abstract val frontend: KClass<out T>
 
-    /** The primitive types of this language. */
+    /** The primitive type names of this language. */
     @get:JsonIgnore
-    val primitiveTypes: Set<String>
-        get() = simpleTypes.keys
+    val primitiveTypeNames: Set<String>
+        get() = builtInTypes.keys
 
-    // TODO: Maybe make this abstract?
-    @get:JsonIgnore
-    @Transient
-    open val simpleTypes: Map<String, Type> =
-        mapOf(
-            "boolean" to BooleanType("boolean", 1, this),
-            "char" to IntegerType("char", 8, this, NumericType.Modifier.NOT_APPLICABLE),
-            "byte" to IntegerType("byte", 8, this, NumericType.Modifier.SIGNED),
-            "short" to IntegerType("short", 16, this, NumericType.Modifier.SIGNED),
-            "int" to IntegerType("int", 32, this, NumericType.Modifier.SIGNED),
-            "long" to IntegerType("long", 64, this, NumericType.Modifier.SIGNED),
-            "float" to FloatingPointType("float", 32, this, NumericType.Modifier.SIGNED),
-            "double" to FloatingPointType("double", 64, this, NumericType.Modifier.SIGNED),
-        )
+    /** The built-in types of this language. */
+    @get:JsonIgnore abstract val builtInTypes: Map<String, Type>
 
     /** The access modifiers of this programming language */
     open val accessModifiers: Set<String>
@@ -87,7 +74,7 @@ abstract class Language<T : LanguageFrontend> : Node() {
         scopeManager: ScopeManager = ScopeManager(),
     ): T
 
-    fun getSimpleTypeOf(typeString: String) = simpleTypes[typeString]
+    fun getSimpleTypeOf(typeString: String) = builtInTypes[typeString]
 
     /** Returns true if the [file] can be handled by the frontend of this language. */
     fun handlesFile(file: File): Boolean {
@@ -103,7 +90,7 @@ abstract class Language<T : LanguageFrontend> : Node() {
         result = 31 * result + fileExtensions.hashCode()
         result = 31 * result + namespaceDelimiter.hashCode()
         result = 31 * result + frontend.hashCode()
-        result = 31 * result + primitiveTypes.hashCode()
+        result = 31 * result + primitiveTypeNames.hashCode()
         result = 31 * result + accessModifiers.hashCode()
         return result
     }
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CPPLanguage.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CPPLanguage.kt
@@ -57,7 +57,7 @@ class CPPLanguage :
     @Transient
     override val simpleTypes =
         mapOf(
-            "bool" to BooleanType("bool", 1, this, NumericType.Modifier.SIGNED),
+            "bool" to BooleanType("bool", 1, this),
             "char" to IntegerType("char", 8, this, NumericType.Modifier.NOT_APPLICABLE),
             "byte" to IntegerType("byte", 8, this, NumericType.Modifier.SIGNED),
             "short" to IntegerType("short", 16, this, NumericType.Modifier.SIGNED),
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CPPLanguage.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CPPLanguage.kt
@@ -55,9 +55,9 @@ class CPPLanguage :
     override val unknownTypeString = listOf("auto")
 
     @Transient
-    override val simpleTypes =
+    override val builtInTypes =
         mapOf(
-            "bool" to BooleanType("bool", 1, this),
+            "bool" to BooleanType("bool", language = this),
             "char" to IntegerType("char", 8, this, NumericType.Modifier.NOT_APPLICABLE),
             "byte" to IntegerType("byte", 8, this, NumericType.Modifier.SIGNED),
             "short" to IntegerType("short", 16, this, NumericType.Modifier.SIGNED),
--- a/cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
+++ b/cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
@@ -34,7 +34,7 @@ import de.fraunhofer.aisec.cpg.graph.types.*
 import org.neo4j.ogm.annotation.Transient
 
 /** The Go language. */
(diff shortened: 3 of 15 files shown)


cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguage.kt
```diff
@@ -57,18 +57,34 @@ open class JavaLanguage :
...
     @Transient
     @JsonIgnore
     override val simpleTypes =
         mapOf(
-            "boolean" to IntegerType("boolean", 1, this, NumericType.Modifier.SIGNED),
+            // Boolean Types:
+            // https://docs.oracle.com/javase/specs/jls/se19/html/jls-4.html#jls-4.2.5
+            "boolean" to BooleanType("boolean", 1, this),
+            "Boolean" to BooleanType("java.lang.Boolean", 1, this),
+            "java.lang.Boolean" to BooleanType("java.lang.Boolean", 1, this),
```
oxisto, line 72: This is debatable. It is actually an ObjectType, not a boolean type. however you can do a direct assignment and boolean comparison on it, so I would suggest to model it as a boolean type:
```java
Boolean b = false;
if(b) {
}
```

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguage.kt
```diff
@@ -57,18 +57,34 @@ open class JavaLanguage :
...
+
+            // Integral Types:
```
oxisto, line 74: Should we also add `java.lang.Integer` and such here? They also "behave" like their respective primitive types.

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
```diff
@@ -46,21 +43,62 @@ open class GoLanguage : Language<GoLanguageFrontend>(), HasShortCircuitOperators
...
             "uint16" to IntegerType("uint16", 16, this, NumericType.Modifier.UNSIGNED),
+            // https://pkg.go.dev/builtin#uint32
             "uint32" to IntegerType("uint32", 32, this, NumericType.Modifier.UNSIGNED),
+            // https://pkg.go.dev/builtin#uint64
             "uint64" to IntegerType("uint64", 64, this, NumericType.Modifier.UNSIGNED),
+            // https://pkg.go.dev/builtin#uintptr
+            "uintptr" to
+                IntegerType(
+                    "uint64",
+                    null /* probably 64 bits */,
```
oxisto, line 82: It is defined as the maximum bit size to support a pointer (which is probably 64 bits "currently"), should we set it to 64 bits and add a comment?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -67,7 +67,7 @@ abstract class Language<T : LanguageFrontend> : Node() {
     @Transient
     open val simpleTypes: Map<String, Type> =
         mapOf(
```
oxisto, line 69: Do these default values really make any sense? Because any language is / needs to overwrite it anyway?

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
KuechA: I wouldn't make any assumptions on the architecture if we cannot be sure about it or make the "worst case" assumption. Otherwise, if we set it to 64 bits and e.g. run it on a 32 bit architecture, we will run into problems and we won't be able to catch that bug (even if it won't be likely in this case)

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguage.kt
konradweiss: I would be fine with it but I am conerned that when the variable of type `Boolean` is actually used as an object something is weird, like a call on the object. Would WrapperType<BooleanType> make sense in the future to differentiate. 
The Save thing would be to make it an `ObjectType` or we try having it as `BooleanType` and see if something breaks?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
KuechA: Probably not. We can delete them

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CPPLanguage.kt
```diff
@@ -57,7 +57,7 @@ class CPPLanguage :
     @Transient
     override val simpleTypes =
         mapOf(
-            "bool" to BooleanType("bool", 1, this, NumericType.Modifier.SIGNED),
+            "bool" to BooleanType("bool", 1, this),
```
KuechA, line 60: If BooleanType is always size 1, we can probably make this the default

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
```diff
@@ -42,6 +42,7 @@ class PythonLanguage : Language<PythonLanguageFrontend>(), HasShortCircuitOperat
     override val conjunctiveOperators = listOf("and")
     override val disjunctiveOperators = listOf("or")
 
+    /** See [Documentation](https://docs.python.org/3/library/stdtypes.html#). */
     @Transient
     override val simpleTypes =
         mapOf(
```
KuechA, line 48: Should we make `bool` also a `BooleanType`? The documentation states that it's an integer type but for consistency reasons (and iff the bitwidth is 1), we should use the new type here as well.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
konradweiss: I am fine with default values, do you think they are to distinct for most languages? For char byte and short I would expect these sizes to be reasonable defaults

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
```diff
@@ -46,21 +43,62 @@ open class GoLanguage : Language<GoLanguageFrontend>(), HasShortCircuitOperators
     override val startCharacter = '['
     override val endCharacter = ']'
 
+    /** See [Documentation](https://pkg.go.dev/builtin). */
     @Transient
     override val simpleTypes =
         mapOf(
+            // https://pkg.go.dev/builtin#any
+            // TODO: Actually, this should be a type alias to interface{}
+            "any" to ObjectType("any", listOf(), false, this),
```
KuechA, line 52: Are these ObjectTypes really "simple types"? Is it like `java.lang.Object` or so?

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
konradweiss: Is a uint64 downgraded to a 32 bits on 32 bit architectures or simply not compilable?

konradweiss (changes requested):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
oxisto: > I am fine with default values, do you think they are to distinct for most languages? For char byte and short I would expect these sizes to be reasonable defaults

They are overwritten anyway by the specific type in each supported language that we currently have.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
oxisto: Would be fine by me. 
Any opinion on this, @maximiliankaul ?

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguage.kt
oxisto: Lets have it as boolean type then.

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
oxisto: > Is a uint64 downgraded to a 32 bits on 32 bit architectures or simply not compilable?

Oops that's a copy/paste, error. It should be uintptr. And yes it is 32 bits on a 32 bit architecture, so I will keep it as null because we do not know the architecture.

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguage.kt
KuechA: Actually, all `NumericType`s (and thus `BooleanType`) inherit from `ObjectType`, so nothing should really go wrong here.

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguage.kt
KuechA: Same as above

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CPPLanguage.kt
oxisto: It is the default, but the language argument comes the bitfield :D

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguage.kt
konradweiss: Yes

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
KuechA: uint64 is not the same as a pointer. The available address space is specific to the architecture, OS and a bunch of things whereas the uint64 would be arranged somewhere in memory (but wouldn't fit into a register)

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
KuechA: Actually, I convinced him to model it like this because I didn't want a specific `BooleanType` and referred to the documentation 😄

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
oxisto: That's complicated :) Actually it is a so called "interface type", a concept which is probably unique to Go. Where as an ObjectType is describing something that is of a particular class (or Java interface, which is also just a name basically), the "interface type" in Go says, this object is something that implements this specific interface, e.g. a set of functions. And it's really dependent on the functions you specify in this interface and not the name, the name is totally irrelevant actually. This is very close to the duck typing concept in ruby, but probably unique to those two languages so not quite 100 % sure how we should model it.

In the end we could have an interface type that has a list of function types describing it. So the ingredients would be there, I am just not sure if we want to have it. And also whether we then want to for example model Java "interfaces" as an "InterfaceType" or an "ObjectType".

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
oxisto: Yeah, although the type "behaves" like a real boolean (unlike C, where it is really only a typedef to int usually for example). 

The documentation you refer to is probably the C API (which I also found): https://docs.python.org/3/c-api/bool.html, which specifies that its implemented as an integer of 0 / 1 internally. But this is probably due to the fact that C has no boolean to begin with, so it needs to be implemented as an int internally.

By vote is on `BooleanType`.

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
KuechA: Ok. Let's keep it like this

KuechA (approved):

PR #1114, opened 2023-03-04
Inferring missing functions in namespace declarations

oxisto:
Previously, we put all inferred function declarations in the global scope (the current TU). However, there are cases where the function is actually residing in a namespace declaration and we need to put it there.

This will not (yet) infer the namespace declaration if it is missing. This might even something that is specific to the frontend, e.g., generate namespace declarations out of import statements.

This will also now use the `ScopeManager` in the inference system, allowing for a better integration of inferred nodes. This for example reduces the amount of duplicate inferred nodes, since now just one is created and can be invoked by others.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
@@ -672,27 +672,30 @@ class ScopeManager : ScopeProvider {
     }
 
     /**
-     * Directly jumps to a given scope.
+     * Directly jumps to a given scope. Returns the previous scope. Do not forget to set the scope
+     * back to the old scope after performing the actions inside this scope.
      *
      * Handle with care, here be dragons. Should not be exposed outside of the cpg-core module.
      */
     @PleaseBeCareful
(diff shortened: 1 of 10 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -650,17 +658,41 @@ class ScopeManager : ScopeProvider {
...
         }
 
-        return resolve(s) { it.name.lastPartsMatch(call.name) && it.hasSignature(call.signature) }
+        return s
+    }
+
+    /**
+     * Directly jumps to a given scope.
+     *
+     * Handle with care, here be dragons. Should not be exposed outside of the cpg-core module.
```
KuechA, line 677: We should probably add "Returns the previous scope. Do not forget to set the scope back to the old scope after performing the actions inside this scope." or so to the documentation

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
oxisto: I made `jumpTo` private now.

Issue #1066, opened 2023-01-26
PDG for Branch nodes


KuechA:
* [ ] We should generate a new Class/Interface for branching nodes and loops and implement an (optional) program dependence graph.
* [ ] This graph should be used for implicit data flows which are currently modeled as DFG edges to the respective branching node.

PR #1135, opened 2023-03-18
Improvements to EOG iteration and more applications

### LINKED ISSUE #1067: Propagate UNREACHABLE Property for EOG edges
It would be good to propagate the `UNREACHABLE` property of EOG edges in the whole statement.

### LINKED ISSUE #1141: Missing DataFlow Edge for InitializerListExpression
# Unexpected behavior

DataFlow edge to `InitializerListExpression` is missing.

# Code of interest

```c++
int foo() {
    return 0;
}

int main() {
    int i{foo()};

    return i;
}

```

image

From the `CallExpression` `foo()` there should be a DFG edge to the `InitializerListExpression`

None

None

KuechA:
Multiple (future) passes or analyses compute a fix-point by traversing the EOG and performing some operations. We do not want to implement the same strategy multiple times, so I tried to generalize the EOG iteration and the respective transformation of a state.

This PR...
* [x] adds a (basic) implementation of the "worklist" fix-point iteration and provides generic interface for lattices and the program state.
* [x] changes the ControlFlowSensitiveDFGPass to make use of this algorithm
* [x] handles InitializerListExpression (closes #1141 but turned out to be a bug in the frontend)
* [x] adds a new node abstracting all kinds of nodes which have a conditional execution of paths (`BranchingNode`) as discussed in (#1066). It has the field `branchingDecision` which is the condition or similar. The following nodes implement this interface:
    - `IfStatement`
    - `WhileStatement`
    - `ForStatement`
    - `ForEachStatement`
    - `SwitchStatement`
    - `CatchClause` (even if it is incorrectly handled in the EOGPass)
    - `ConditionalExpression` (see #1142 )
    - `ShortCircuitOperator` which extends BinaryOperator. This one has to be adapted for each language (i.e., the language frontend has to decide if a binary op is a normal `BinaryOperator` or a `ShortCircuitOperator`. There's already a configuration in the `Language` class).
* [x] adds the fields `nextCDGEdges` and `prevCDGEdges` to the class `Node`. These hold the control dependence graph.
* [x] implements the new (optional?) pass `ControlDependenceGraphPass` which adds the CDG edges for each function (intraprocedurally).
* [x] Propagates the `Unreachable` properties of EOG edges (Closes #1067)


KuechA:
There seem to be some issues with identifying loops in the EOG for the ControlFlowSensitiveDFGPass.

KuechA:
@peckto Could you test if there are performance issues with the new and redesigned passes which haven't been there before?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
```diff
@@ -206,4 +206,7 @@ interface HasShortCircuitOperators : LanguageTrait {
     val conjunctiveOperators: List<String>
     // '||', 'or', 'v'
     val disjunctiveOperators: List<String>
+
+    val operatorCodes: Set<String>
```
oxisto, line 210: Missing documentation - also for conjunctive and disjunctive, even if its not your code ;) I wonder if this is also misleading because these are only *boolean* operators, right? Not all the operator codes a language potentially has.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/BranchingNode.kt
```diff
@@ -0,0 +1,32 @@
...
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph
+
+/** A node triggering a conditional execution of other code. */
+interface BranchingNode {
+    /** The node which affects the next EOG edge. Typically, this is a condition or similar. */
+    val branchingDecision: Node?
```
oxisto, line 31: Is "decision" the correct term here?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/ExpressionBuilder.kt
```diff
@@ -68,7 +69,17 @@ fun MetadataProvider.newBinaryOperator(
...
 ): BinaryOperator {
-    val node = BinaryOperator()
+    val node =
+        if (
+            this is LanguageProvider &&
+                (this.language as? HasShortCircuitOperators)
+                    ?.operatorCodes
+                    ?.contains(operatorCode) == true
+        ) {
+            ShortCircuitOperator()
```
oxisto, line 79: We should probably document this magic in the documentation of this node builder, since it is a little bit different than the others.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
```diff
@@ -25,22 +25,63 @@
...
     /** This field contains the iteration subject of the loop. */
     @AST var iterable: Statement? = null
 
     /** This field contains the body of the loop. */
     @AST var statement: Statement? = null
 
+    override val branchingDecision: Node?
+        get() = iterable
+
+    override var statementEdges: MutableList<PropertyEdge<Statement>>
```
oxisto, line 56: What is the reason for implementing statement holder here?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
```diff
@@ -25,22 +25,63 @@
...
+        get() {
+            val statements = mutableListOf<PropertyEdge<Statement>>()
+            variable?.let { statements.add(PropertyEdge(this, it)) }
+            iterable?.let { statements.add(PropertyEdge(this, it)) }
+            statement?.let { statements.add(PropertyEdge(this, it)) }
```
oxisto, line 61: this includes the compound statement itself (not the contents of it), why? this whole thing confuses me a little bit :)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -0,0 +1,301 @@
...
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdge
+import java.util.IdentityHashMap
+
+/**
+ * An abstract class representing complete lattices, i.e., an ordered structure of values of type
+ * [T]. Implementations of this class have to implement the comparator, the least upper bound of two
+ * lattices.
+ */
+abstract class Lattice<T>(open val elements: T) : Comparable<Lattice<T>?> {
```
oxisto, line 37: shouldn't this be a collection of T or something?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -0,0 +1,301 @@
...
+ * A worklist. Essentially, it stores mappings of nodes to the states which are available there and
+ * determines which nodes have to be analyzed.
+ */
+class Worklist<K : Any, N : Any, V>() {
+    /** A mapping of nodes to the state which is currently available there. */
+    var globalState: MutableMap<K, State<N, V>> = mutableMapOf()
+        private set
+
+    /** A list of all nodes which have already been visited. */
+    private val alreadySeen = mutableListOf<K>()
```
oxisto, line 141: you probably want an identityhashset for the seen-list, otherwise you might run into performance problems if lists are large and equals/hashcode must be used.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -0,0 +1,301 @@
...
+        for ((_, v) in globalState) {
+            state?.lub(v)
+        }
+
+        return state ?: State()
+    }
+}
+
+/** A very simple implementation of the worklist algorithm. */
+class EOGWorklist {
```
oxisto, line 217: a lot of classes.. why is EOGworklist not deriving from a "wordlist", but somehow has it in the transformation as an argument? Maybe different naming?

oxisto: You probably could just get rid of this class altogether and directly have those two functions as top-level functions.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
```diff
@@ -0,0 +1,241 @@
...
+import de.fraunhofer.aisec.cpg.graph.functions
+import de.fraunhofer.aisec.cpg.graph.statements.IfStatement
+import de.fraunhofer.aisec.cpg.helpers.EOGWorklist
+import de.fraunhofer.aisec.cpg.helpers.Lattice
+import de.fraunhofer.aisec.cpg.helpers.State
+import de.fraunhofer.aisec.cpg.helpers.Worklist
+import de.fraunhofer.aisec.cpg.passes.order.DependsOn
+
+@DependsOn(EvaluationOrderGraphPass::class)
+class ControlDependenceGraphPass : Pass() {
```
oxisto, line 44: What is it and what is it good for?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
```diff
@@ -0,0 +1,241 @@
...
+                        }
+                    // Map this branching node to its merging points
+                    Pair(branchingNode as Node, mergingPoints)
+                }
+                .toTypedArray()
+        )
+
+    companion object {
+        @JvmStatic
+        fun handleEdge(
```
oxisto, line 151: Why a companion object here? just use a top-level function

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
```diff
@@ -0,0 +1,241 @@
...
+                1
+            else 0
+        } else {
+            -1
+        }
+    }
+}
+
+/**
+ * A state which actually holds a state for all [PropertyEdge]s, one only for declarations and one
```
oxisto, line 238: This sentence confuses me.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
konradweiss: Kinda, however if other operators are interpreted more loosely as a question "is this non null", then it is not an issue. I think these are the best terms we can use here but maybe we find another.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
konradweiss: I am fine with the term. How would be a `branchedBy`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForStatement.kt
```diff
@@ -26,11 +26,13 @@
...
 
 import de.fraunhofer.aisec.cpg.graph.AST
+import de.fraunhofer.aisec.cpg.graph.BranchingNode
+import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.declarations.Declaration
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
 import java.util.*
 
-class ForStatement : Statement() {
+class ForStatement : Statement(), BranchingNode {
```
konradweiss, line 35: If we have the `ForEachStatement` as `StatementHolder` we should do that with every loop right? And switches as well?

konradweiss (changes requested):
We have to decide whether all Nodes holding other statements are `StatementHolder` and in that case include `IfStatement`, `ForStatement`, `SwitchStatement` etc.

cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/MultiValueEvaluatorTest.kt
```diff
@@ -167,21 +167,25 @@ class MultiValueEvaluatorTest {
 
         printB = main.bodyOrNull<CallExpression>(1)
         assertNotNull(printB)
+        evaluator.clearPath()
         value = evaluator.evaluate(printB.arguments.firstOrNull()) as ConcreteNumberSet
```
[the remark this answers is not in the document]
KuechA: I think the path is interesting if we want to have the information which nodes were considered when evaluating the expression. But it's very error-prone like this if you want to reuse the evaluator.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
KuechA: I think I needed a `StatementHolder` for the Fluent-DSL which I use in a test. But maybe I just don't understand when to use it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
KuechA: Maybe I made a mistake in Fluent-DSL definition. Somehow, this worked for me :D Feel free to fix the `forEachStmt` function in `Fluent.kt` and remove this code.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
KuechA: No, it could be anything. I updated the documentation

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -30,8 +30,14 @@ import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdge
 import java.util.IdentityHashMap
 
 /**
- * An abstract class representing complete lattices, i.e., an ordered structure of values of type
- * [T]. Implementations of this class have to implement the comparator, the least upper bound of two
+ * A complete lattices is an ordered structure of values of type [T]. [T] could be anything, e.g., a
```
oxisto, line 33:
```suggestion
 * A complete lattice is an ordered structure of values of type [T]. [T] could be anything, e.g., a
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -215,89 +222,86 @@ class Worklist<K : Any, N : Any, V>() {
...
-        transformation:
-            (PropertyEdge<Node>, State<N, V>, Worklist<K, N, V>) -> Pair<State<N, V>, Boolean>
-    ): State<N, V> {
-        val globalState = mutableMapOf<K, State<N, V>>()
-        for (startEdge in startEdges) {
-            globalState[startEdge] = startState
-        }
-        val worklist = Worklist(globalState)
-        startEdges.forEach { worklist.push(it, startState) }
+inline fun <reified K : PropertyEdge<Node>, N : Node, V> iterateEOGEN(
```
oxisto, line 283: What is the difference between iterateEOG and iterateEOGEN?

KuechA: I merged them into one

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
oxisto: I will have a look at it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForStatement.kt
oxisto: See above, this was introduced because of the fluent DSL and I will try to get rid of it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
oxisto: I tried to remove the `StatementHolder` and rework the fluent DSL part for for-each a little bit. It's not really pretty but creating the node works. However, now the `ControlDependenceGraphPassTest` seems to hang; which is weird because the for-each statement that is created is correct. Maybe it was not correct before and should have never worked? :D

KuechA: Removing the `StatementHolder` adds the nodes in the parent and the for-each statement which means that it's necessary right now. However, I managed to fix the test even in the presence of the bug/loop.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -0,0 +1,277 @@
...
+ * set, a new data structure (like a range), or anything else. [T] depends on the analysis and
+ * typically has to abstract the value for the specific purpose.
+ *
+ * This class is actually used to hold individual instances of the lattice's elements and to compute
+ * bigger elements depending on these two elements.
+ *
+ * Implementations of this class have to implement the comparator, the least upper bound of two
+ * lattices.
+ */
+abstract class Lattice<T>(open val elements: T) : Comparable<Lattice<T>?> {
```
oxisto, line 43: As discussed, this data structure seems to represent a lattice element, rather than a complete lattice?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -0,0 +1,277 @@
...
+    /**
+     * Computes the least upper bound of this lattice and [other]. It returns a new object and does
+     * not modify either of the objects.
+     */
+    abstract fun lub(other: Lattice<T>?): Lattice<T>
```
oxisto, line 48: Why nullable? Seems to introduce a lot of unnecessary nulls checks.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * Implements the [Lattice] over a set of nodes. The lattice itself is constructed by the powerset.
+ */
+class PowersetLattice(override val elements: Set<Node>) : Lattice<Set<Node>>(elements) {
+    override fun lub(other: Lattice<Set<Node>>?) =
+        PowersetLattice((other?.elements ?: setOf()).union(this.elements))
+
+    override fun duplicate() = PowersetLattice(this.elements.toSet())
+
+    override fun compareTo(other: Lattice<Set<Node>>?): Int {
```
oxisto, line 63: compareTo confuses me. Is this a workaround in order to compute the greatest lower bound? because for the powersets glb we could just use `intersect` if I understand lattices correctly 🤯

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -0,0 +1,277 @@
...
+        } else {
+            -1
+        }
+    }
+}
+
+/**
+ * Stores the current state. I.e., it maps [K] (e.g. a [Node] or [PropertyEdge]) to a [Lattice]. It
+ * provides some useful functions e.g. to check if the mapping has to be updated (e.g. because there
+ * are new nodes or because a new lattice is bigger than the old one).
```
oxisto, line 75: wording: lattice element?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -0,0 +1,277 @@
...
+        var update = false
+        for ((node, newLattice) in other) {
+            update = push(node, newLattice) || update
+        }
+        return Pair(this, update)
+    }
+
+    /**
+     * Checks if an update is necessary, i.e., if [other] contains nodes which are not present in
+     * `this` and if the lattice of a node in [other] "is bigger" than the respective lattice in
```
oxisto, line 96: wording: lattice element?

cpg-core/src/test/resources/dfg/initializerListExpression.cpp
```diff
@@ -0,0 +1,9 @@
+int foo() {
```
oxisto, line 1: leftover?

cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/MultiValueEvaluatorTest.kt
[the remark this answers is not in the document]
oxisto: Ok, lets keep it then

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
oxisto: You should be able to get rid of them. Makes the code much easier to read and maintain.

```patch
diff --git a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
index 7083dd7fc..dbe7ba5c5 100644
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
@@ -40,12 +40,12 @@ import java.util.IdentityHashMap
  * Implementations of this class have to implement the comparator, the least upper bound of two
  * lattices.
  */
-abstract class Lattice<T>(open val elements: T) : Comparable<Lattice<T>?> {
+abstract class Lattice<T>(open val elements: T) : Comparable<Lattice<T>> {
     /**
      * Computes the least upper bound of this lattice and [other]. It returns a new object and does
      * not modify either of the objects.
      */
-    abstract fun lub(other: Lattice<T>?): Lattice<T>
+    abstract fun lub(other: Lattice<T>): Lattice<T>
 
     /** Duplicates the object, i.e., makes a deep copy. */
     abstract fun duplicate(): Lattice<T>
@@ -55,14 +55,14 @@ abstract class Lattice<T>(open val elements: T) : Comparable<Lattice<T>?> {
  * Implements the [Lattice] over a set of nodes. The lattice itself is constructed by the powerset.
  */
 class PowersetLattice(override val elements: Set<Node>) : Lattice<Set<Node>>(elements) {
-    override fun lub(other: Lattice<Set<Node>>?) =
-        PowersetLattice((other?.elements ?: setOf()).union(this.elements))
+    override fun lub(other: Lattice<Set<Node>>) =
+        PowersetLattice(other.elements.union(this.elements))
 
     override fun duplicate() = PowersetLattice(this.elements.toSet())
 
-    override fun compareTo(other: Lattice<Set<Node>>?): Int {
-        return if (this.elements.containsAll(other?.elements ?: setOf())) {
-            if (this.elements.size > (other?.elements?.size ?: 0)) 1 else 0
+    override fun compareTo(other: Lattice<Set<Node>>): Int {
+        return if (this.elements.containsAll(other.elements)) {
+            if (this.elements.size > (other.elements.size)) 1 else 0
         } else {
             -1
         }
@@ -99,7 +99,8 @@ open class State<K, V> : IdentityHashMap<K, Lattice<V>>() {
     open fun needsUpdate(other: State<K, V>): Boolean {
         var update = false
         for ((node, newLattice) in other) {
-            update = update || node !in this || newLattice > this[node]
+            val current = this[node]
+            update = update || current == null || newLattice > current
         }
         return update
     }
@@ -122,13 +123,15 @@ open class State<K, V> : IdentityHashMap<K, Lattice<V>>() {
         if (newLattice == null) {
             return false
         }
-        if (newNode in this && this[newNode]?.let { it >= newLattice } == true) {
+
+        val current = this[newNode]
+        if (current != null && current >= newLattice) {
             // newLattice is "smaller" than the currently stored one. We don't add it anything.
             return false
-        } else if (newNode in this) {
+        } else if (current != null) {
             // newLattice is "bigger" than the currently stored one. We update it to the least
             // upper bound
-            this[newNode] = newLattice.lub(this[newNode])
+            this[newNode] = newLattice.lub(current)
         } else {
             this[newNode] = newLattice.duplicate()
         }
@@ -211,7 +214,7 @@ class Worklist<K : Any, N : Any, V>() {
     fun hasAlreadySeen(currentNode: K) = currentNode in alreadySeen
 
     /** Computes the meet over paths for all the states in [globalState]. */
-    fun mop(): State<N, V>? {
+    fun mop(): State<N, V> {
         val firstKey = globalState.keys.firstOrNull()
         val state = globalState[firstKey]
         for ((_, v) in globalState) {
@@ -236,7 +239,7 @@ inline fun <reified K : Node, V> iterateEOG(
     startNode: K,
     startState: State<K, V>,
     transformation: (K, State<K, V>, Worklist<K, K, V>) -> State<K, V>
-): State<K, V>? {
+): State<K, V> {
     val worklist = Worklist(mutableMapOf(Pair(startNode, startState)))
     worklist.push(startNode, startState)
 
diff --git a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
index 15a972ea7..094a283bf 100644
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
@@ -224,9 +224,8 @@ private fun <T : Node> PropertyEdge<T>.isConditionalBranch(): Boolean {
  */
 class PrevEOGLattice(override val elements: Map<Node, Set<Node>>) :
     Lattice<Map<Node, Set<Node>>>(elements) {
-    override fun lub(other: Lattice<Map<Node, Set<Node>>>?): Lattice<Map<Node, Set<Node>>> {
-        val newMap =
-            (other?.elements ?: mapOf()).mapValues { (_, v) -> v.toMutableSet() }.toMutableMap()
+    override fun lub(other: Lattice<Map<Node, Set<Node>>>): Lattice<Map<Node, Set<Node>>> {
+        val newMap = other.elements.mapValues { (_, v) -> v.toMutableSet() }.toMutableMap()
         for ((key, value) in this.elements) {
             newMap.computeIfAbsent(key, ::mutableSetOf).addAll(value)
         }
@@ -235,16 +234,14 @@ class PrevEOGLattice(override val elements: Map<Node, Set<Node>>) :
 
     override fun duplicate() = PrevEOGLattice(this.elements.toMap())
 
-    override fun compareTo(other: Lattice<Map<Node, Set<Node>>>?): Int {
+    override fun compareTo(other: Lattice<Map<Node, Set<Node>>>): Int {
         return if (
-            this.elements.keys.containsAll(other?.elements?.keys ?: setOf()) &&
-                this.elements.all { (k, v) -> v.containsAll(other?.elements?.get(k) ?: setOf()) }
+            this.elements.keys.containsAll(other.elements.keys) &&
+                this.elements.all { (k, v) -> v.containsAll(other.elements[k] ?: setOf()) }
         ) {
             if (
-                this.elements.keys.size > (other?.elements?.keys?.size ?: 0) ||
-                    this.elements.any { (k, v) ->
-                        v.size > (other?.elements?.get(k) ?: setOf()).size
-                    }
+                this.elements.keys.size > (other.elements.keys.size) ||
+                    this.elements.any { (k, v) -> v.size > (other.elements.get(k) ?: setOf()).size }
             )
                 1
             else 0
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -79,391 +91,238 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : TranslationUni
...
-        }
+/**
+ * A state which actually holds a state for all nodes, one only for declarations and one for
+ * ReturnStatements.
+ */
+class DFGPassState<V>(
+    /** A mapping of a [Node] to its [Lattice]. */
+    var generalState: State<Node, V> = State(),
+    /** A mapping of [Declaration] to its [Lattice]. */
+    var declarationsState: State<Node, V> = State(),
```
oxisto, line 292: The declaration state seems to contain more than just declarations - is that intended?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
KuechA: It's used to determine if we have to continue with the iteration, update something, etc. (e.g., if the new element > current element, we have to continue). By definition, a group has an ordering operator, which we model it with the `compareTo`.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
KuechA: Yes it is because the "generalState" collects the nodes which will later receive DFG edges whereas the nodes here don't. I documented this but we can also change the name if necessary.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
```diff
@@ -28,71 +28,196 @@ package de.fraunhofer.aisec.cpg.passes
...
+    override fun lub(other: LatticeElement<Reachability>) =
+        ReachabilityLattice(maxOf(this.elements, other.elements))
+
+    override fun duplicate() = ReachabilityLattice(this.elements)
+
+    override fun compareTo(other: LatticeElement<Reachability>) =
+        this.elements.compareTo(other.elements)
 }
+
+/** The ordering will be as follows: BOTTOM (no information) < UNREACHABLE < REACHABLE < TOP */
```
konradweiss, line 211: Can you give information on what Top stands for like you did for Bottom?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/ShortCircuitOperator.kt
```diff
@@ -0,0 +1,34 @@
...
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.statements.expressions
+
+import de.fraunhofer.aisec.cpg.graph.BranchingNode
+import de.fraunhofer.aisec.cpg.graph.Node
+
+class ShortCircuitOperator : BinaryOperator(), BranchingNode {
```
konradweiss, line 31: I think we need documentation here, ideally with a reference to the language trait.

konradweiss (changes requested):

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
KuechA: No because I did not even use it, so it didn't make sense. I just deleted it

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
```diff
@@ -0,0 +1,261 @@
...
+ *
+ * Note: This method does not account for return statements in the conditional part or endless loops
+ * where the other branch is actually also conditionally executed (or not). It should be easy to
+ * change this if we do not want this behavior (just remove the condition on the start node of the
+ * "false" branch).
+ */
+private fun <T : Node> PropertyEdge<T>.isConditionalBranch(): Boolean {
+    return if (this.getProperty(Properties.BRANCH) == true) {
+        true
+    } else this.start is IfStatement && this.getProperty(Properties.BRANCH) == false
```
seelchen, line 218: The `else` Expression of `ConditionalExpression` and possible branches of `ShortCircuitOperators` are not recognized as conditional branch with this condition
```suggestion
    } else
        (this.start is IfStatement ||
            this.start is ConditionalExpression ||
            this.start is ShortCircuitOperator) && this.getProperty(Properties.BRANCH) == false
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -238,6 +262,11 @@ open class Node : IVisitable<Node>, Persistable, LanguageProvider, ScopeProvider
         prev.nextDFG.add(this)
     }
 
+    fun addPrevCDG(prev: Node) {
+        prevCDGEdges.add(PropertyEdge(this, prev))
+        prev.nextCDGEdges.add(PropertyEdge(prev, this))
```
seelchen, line 267: The edge in both nodes should have the same start and end
```suggestion
        val edge = PropertyEdge(prev, this)
        prevCDGEdges.add(edge)
        prev.nextCDGEdges.add(edge)
```

oxisto (approved):
LGTM once the conflict is resolved

PR #1136, opened 2023-03-21
Made the `FunctionPointerCallResolver` compatible with the `ControlFlowSensitiveDFGPass`

oxisto:
The function pointer call resolver pass had a hidden blocking dependency on the `ControlFlowSensitiveDFGPass`, in a way that it needed to be executed BEFORE the control flow sensitive DFG was built. This was/is because of inconsistencies between both the "regular" DFG pass and the control-flow sensitive one (which we need to tackle in a seperate PR).

This PR makes resolving function pointers a little bit cleaner, since basically it tries to resolve every `callee` of a `CallExpression` that is of `FunctionPointerType`, regardless of any AST structure or other weird resolving. This also moves the `VariableUsageResovler` a little bit closer to resolving not just variables, but also functions (since in the long term we want to merge it with the `CallResolver`)

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.kt
@@ -118,9 +118,21 @@ open class VariableUsageResolver : SymbolResolverPass() {
 
         // For now, we need to ignore reference expressions that are directly embedded into call
         // expressions, because they are the "callee" property. In the future, we will use this
-        // property to actually resolve the function call.
+        // property to actually resolve the function call. However, there is a special case that
+        // we want to catch already, that is if we are "calling" a reference to a variable. This
+        // can be done in several languages, e.g., in C/C++ as function pointers or in Go as
+        // function references. In this case, we want to resolve the declared reference expression
+        // of this call expression back to its original variable declaration. In the future, we want
+        // to extend this particular code to resolve all callee references to their declarations,
+        // i.e., their function definitions and get rid of the separate CallResolver.
         if (parent is CallExpression && parent.callee === current) {
-            return
+            // Peek into the declaration, and if it is a variable, we can proceed normally, as we
+            // are running into the special case explained above. Otherwise, we abort here (for
+            // now).
+            val wouldResolveTo = scopeManager.resolveReference(current, current.scope)
+            if (wouldResolveTo !is VariableDeclaration) {
+                return
+            }
         }
 
         // only consider resolving, if the language frontend did not specify a resolution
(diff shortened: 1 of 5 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.kt
```diff
@@ -118,9 +118,21 @@ open class VariableUsageResolver : SymbolResolverPass() {
...
+        // function references. In this case, we want to resolve the declared reference expression
+        // of this call expression back to its original variable declaration. In the future, we want
+        // to extend this particular code to resolve all callee references to their declarations,
+        // i.e., their function definitions and get rid of the separate CallResolver.
         if (parent is CallExpression && parent.callee === current) {
-            return
+            // Peek into the declaration, and if it is a variable, we can proceed normally, as we
+            // are running into the special case explained above. Otherwise, we abort here (for
+            // now).
+            val wouldResolveTo = scopeManager.resolveReference(current, current.scope)
```
oxisto, line 132: Unfortunately, this now calls `resolveReference` twice in the case of a call expression reference. Anyone got a good idea how to merge this with line 139?

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontendTest.kt
```diff
@@ -1413,6 +1413,42 @@ internal class CXXLanguageFrontendTest : BaseTest() {
         assertRefersTo(callee.rhs, singleParam)
     }
 
+    @Test
```
KuechA, line 1416: We should test if the resolution works with and without the `ControlFlowSensitiveDFGPass`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.kt
KuechA: Maybe not a good idea but you could make a `var wouldResolveTo: Declaration? = nukk` before line 128, set it in line 132, and only call `resolveReference` in line 139 if it is null

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontendTest.kt
oxisto: Is that really needed? I thought the idea was to merge two DFG passes anyway?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/VariableUsageResolver.kt
oxisto: Its better than my solution, so sold!

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cpp/CXXLanguageFrontendTest.kt
KuechA: Yes, but this means that the `FunctionPointerCallResolver` has to handle the case of the state after the control flow sensitive pass and the one without it. It's just to make sure that we don't miss something because our tests always have the "control flow sensitive" configuration by accident or so (just as we missed that the old version didn't work properly with a slightly different but acceptable configuration)

oxisto: Ok, I duplicated the test with and without the control-flow-DFG.

KuechA (approved):

PR #1140, opened 2023-03-24
Transfer more of the core-functionality tests to the DSL to get rid of language frontends

KuechA:
In preparation to extract the cpp language frontend and to add tests to core again which were moved to the java language module before.


KuechA:
Many tests of the call resolver (including the constructor tests) and variable resolver seem to be rather language-specific and seem to test a pass but also the language frontend. We could probably split these tests into different ones but need to understand better what they are supposed to do

oxisto:
In addition to this PR, I was exploring how to aggregate jacoco results across different subprojects into one (e.g. the `cog-all` project). See https://github.com/Fraunhofer-AISEC/cpg/pull/1144 and https://docs.gradle.org/current/userguide/jacoco_report_aggregation_plugin.html. It works locally (I get an HTML and XML report with all frontends contributing), but somehow I could not yet convince sonar cloud to take this XML file.

KuechA:
Also, for a few days, the SonarCloud coverage badge reports that the measure cannot be found.

oxisto (approved):

PR #1147, opened 2023-04-02
Add --custom-pass-list and --list-passes

shenniger:
Adds a CLI option to cpg-neo4j that allows a user to choose specific passes to run. Example: Use `--custom-pass-list=DFGPass` to run only DFGPass. This is especially useful for debugging.

This feature requires TranslationConfiguration to keep a list of all choosable passes. Currently, the list is stored in a global variable of `KClass` objects (that reflection is used on in order to get their names and create instances of them). For now, this variable only contains passes that are part of `defaultPasses()`. This could be easily changed by adding other passes to the variable.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -372,6 +385,26 @@ private constructor(
             return this
         }
 
+        /** Produces the [Pass] with the given name. */
```
oxisto, line 388: Sounds like a private or internal function

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -372,6 +385,26 @@ private constructor(
...
+        fun producePass(passName: String): Pass {
+            val list = passList.map { Pair(it.simpleName, it) }.toMap()
+            if (!list.contains(passName)) {
+                throw ConfigurationException("Asked to produce unknown pass")
+            }
+            return list[passName]!!.createInstance()
+        }
+
+        /** Returns the list of available passes that can be registered. */
+        fun getPassList(): List<String> {
```
oxisto, line 398: Should be converted to a (virtual) property instead of a getter function

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -165,6 +165,12 @@ class Application : Callable<Int> {
     )
     private var noDefaultPasses: Boolean = false
 
+    @CommandLine.Option(
+        names = ["--custom-pass-list"],
+        description = ["Add custom list of passes (includes --no-default-passes)"]
```
KuechA, line 170: This param receives a comma-separated list, right? Could be included in the documentation.

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -334,8 +346,13 @@ class Application : Callable<Int> {
             translationConfiguration.sourceLocations(filePaths)
         }
 
-        if (!noDefaultPasses) {
+        if (!noDefaultPasses && customPasses == "ALL") {
```
KuechA, line 349: Why will "ALL" result in the default passes and not in really all available passes?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -228,6 +228,19 @@ private constructor(
         private var compilationDatabase: CompilationDatabase? = null
         private var matchCommentsToNodes = false
         private var addIncludesToGraph = true
+        private var passList =
```
KuechA, line 231: This list is not complete. I.e., there are language frontend-dependent passes (e.g. `JavaExternalTypeHierarchyResolver`, the `CompressLLVMPass`) or maybe a user might add another custom pass. If we could auto-generate this list/mapping somehow, this would be more cool :)

KuechA:
Note: Many passes depend on another pass, so what you specify here and what you will actually observe will differ

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
konradweiss: All I can think of is some form of reflection that shows all Classes which inherit from Pass, also we then need a command line option that prints the list of available passes. When passes are then chosen they should be instantiable by getting the class from the provided class name. Required passes will have to be added as well. the question is if we print that information in the previously mentioned option that shows all available passes. @shenniger does that sound like something solving what you had in mind?

oxisto: Maybe we can explicitly state that these are only the "core" passes or something like that? I also see an issue by making this configurable because some passes *really* need to be run, otherwise things are just plain wrong (e.g. the language-frontend dependant passes). The only pass that really is sort of "optional" is the control-flow-sensitive DFG. And there we discussed about merging this with the regular DFG pass  and making the control flow sensitive stuff configurable.

TL;DR: I do not really see a use case for making the pass list configurable.

konradweiss: I could see it for the use case where people implement their own passes, or in the future when our pass system is more diverse. However as you said, the dependent passes would have to be ran the same.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
shenniger: Resolved in 979866b61.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
shenniger: Resolved in 979866b61.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -372,6 +385,26 @@ private constructor(
             return this
         }
 
+        /** Produces the [Pass] with the given name. */
+        fun producePass(passName: String): Pass {
+            val list = passList.map { Pair(it.simpleName, it) }.toMap()
+            if (!list.contains(passName)) {
```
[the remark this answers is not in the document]
shenniger: Resolved in 979866b61.

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
shenniger: Good point. Renamed it to "DEFAULT".

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -372,6 +385,26 @@ private constructor(
             return this
         }
 
+        /** Produces the [Pass] with the given name. */
+        fun producePass(passName: String): Pass {
+            val list = passList.map { Pair(it.simpleName, it) }.toMap()
```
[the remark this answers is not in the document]
shenniger: Resolved in 979866b61.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
shenniger: @konradweiss, regarding auto-generation: I was looking at that, but it seems that there is no easy way to get the list of subclasses of a (non-sealed) Kotlin class. Hence the idea to store a list somewhere and then at least use reflection to get the names of the classes and instantiate. But maybe TranslationConfiguration is not a good place for such a list – we could put it somewhere more central, or allow for sublists per module – let me know what you think.

@oxisto regarding pass dependencies: The idea is to use this mostly for debugging and in cases where no full DFG is needed. But I agree that it should be in the docs that pass dependencies are still resolved with this, so you will likely get more passes than you asked for.

shenniger:
Hi everyone, thanks for your feedback! I believe I resolved most issues with the last commit. Also wrote some more notes on auto-generating the list and pass dependencies above.

peckto (approved):

oxisto (commented):
Sorry for dragging this along (and be somewhat negative towards this), but somehow this doesn't feel right to me. It introduces three list of passes - which are completely static, so I am not quite sure what the point of returning them is via the config. You could just return the list directly in the neo4j application if you really need this information in the CLI.

With regards to registering custom passes, I would suggest following the same as we have with the languages:

https://github.com/Fraunhofer-AISEC/cpg/blob/747d435d700df04ca38bbb28637fa3cf034c2140/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt#L493C13-L498

This has the slight disadvantage that you need to specify the complete FQN, but it has the advantage that you can register any pass that is somehow on the class path.

If we really want to have a list of all *possible* passes, we could use https://github.com/ronmamo/reflections#usage (JVM only) or possibly move towards a dependency injection framework like https://github.com/InsertKoinIO/koin

oxisto (changes requested):
Sorry for dragging this along (and be somewhat negative towards this), but somehow this doesn't feel right to me. It introduces three list of passes - which are completely static, so I am not quite sure what the point of returning them is via the config. You could just return the list directly in the neo4j application if you really need this information in the CLI.

With regards to registering custom passes, I would suggest following the same as we have with the languages:

https://github.com/Fraunhofer-AISEC/cpg/blob/747d435d700df04ca38bbb28637fa3cf034c2140/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt#L390-L411

This has the slight disadvantage that you need to specify the complete FQN, but it has the advantage that you can register any pass that is somehow on the class path.

If we really want to have a list of all *possible* passes, we could use https://github.com/ronmamo/reflections#usage (JVM only) or possibly move towards a dependency injection framework like https://github.com/InsertKoinIO/koin

konradweiss:
> Sorry for dragging this along (and be somewhat negative towards this), but somehow this doesn't feel right to me. It introduces three list of passes - which are completely static, so I am not quite sure what the point of returning them is via the config. You could just return the list directly in the neo4j application if you really need this information in the CLI.
> 
> With regards to registering custom passes, I would suggest following the same as we have with the languages:
> 
> https://github.com/Fraunhofer-AISEC/cpg/blob/747d435d700df04ca38bbb28637fa3cf034c2140/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt#L390-L411
> 
> This has the slight disadvantage that you need to specify the complete FQN, but it has the advantage that you can register any pass that is somehow on the class path.
> 
> If we really want to have a list of all _possible_ passes, we could use https://github.com/ronmamo/reflections#usage (JVM only) or possibly move towards a dependency injection framework like https://github.com/InsertKoinIO/koin (in the future)

So I will try to sum up and find tasks to solve those concerns and find some middle ground to the current implementation:

- I agree that we should try to keep as much static pass configuration out of the cpg-core and move it to the cpg-neo4j, expecialy because the neo4j is the only component that needs to register passes by name. I suggest moving these functions to cpg-neo4j as suggested.
- When it comes to the name based registration: I like having the ability to register all existing passes, on the other hand it is unhandy to specify all passes over their FQN. Suggestion: allow both, create passes from their name, if that fails, e.g. because the pass name was a simple name and not a FQN, prepend out default pass package name.

The last part will allow to register passes not shown among the available passes list, but that is a feature then. Inclusion of an additional framework is something that we should look into but I think it would drag this PR on even longer. Lets investigate that in the future if we see the need to have a complete list of specifiable passes.

How does that sound @oxisto @shenniger?

shenniger:
Thanks all for the good ideas!

I have pushed two major changes to this:
1. Everything about the custom pass list was moved to cpg-neo4j and out of the main module.
2. As suggested, FQDNs can now be used in addition to the passes in the list.

oxisto (approved):

PR #1158, opened 2023-04-18
Use `ConcurrentHashMap` to keep language-specific `UnknownType`s

KuechA:
We had to observe some random crashes when using the unknown type map recently with concurrent modification exceptions. This PR should fix this issue.


oxisto (approved):

oxisto:
In the long run we might want to have an `UnknownLanguage` instead of a lot of null languages.

PR #1169, opened 2023-05-06
Overhaul of the pass system

oxisto:
This PR contains a slight overhaul of the pass system. Specifically, instead of just applying a pass to the whole translation result, it is now either applied to individual components, a translation unit or the whole result. Instead of directly subclassing from `Pass`, it is now needed to subclass `ComponentPass`, `TranslationResultPass` or `TranslationUnitPass`.

That will allow us in the future, to have language-specific execution of passes in individual software components (e.g., server code and client code) or translation units. 

Furthermore, this PR changes the way when passes are actually instantiated. Previously, they were implicitly (or explicitly) created at the time one called `defaultPasses()` or `registerPass` on the `TranslationConfiguration`. This meant, that later important state information, such as the config or the scope manager was injected weirdly in the `accept` method and the single `Pass` object was re-used for different executions. This led to errors if passes were incorrectly cleaned between executions.

In the new system, a `TranslationConfiguration` is only holding a reference to the `KClass` of a pass. The pass is then instantiated out of the class dynamically, once it is actually needed and only "lives" for its scope of execution (e.g. translation result, component, tu).

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
@@ -25,10 +25,9 @@
  */
 package de.fraunhofer.aisec.cpg
 
-import com.fasterxml.jackson.annotation.JsonIdentityInfo
-import com.fasterxml.jackson.annotation.JsonIdentityReference
-import com.fasterxml.jackson.annotation.ObjectIdGenerators
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
 import de.fraunhofer.aisec.cpg.frontends.CompilationDatabase
+import de.fraunhofer.aisec.cpg.frontends.KClassSerializer
 import de.fraunhofer.aisec.cpg.frontends.Language
 import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend
 import de.fraunhofer.aisec.cpg.frontends.cpp.CLanguage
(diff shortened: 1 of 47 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -93,7 +93,9 @@ private constructor(
      * always take priority over those in the whitelist.
      */
     val includeBlocklist: List<Path>,
-    passes: List<Pass>,
+    passes: List<KClass<out Pass<*>>>,
+    val replacedPasses:
```
KuechA, line 97: Some documentation could be useful. What if there are different replacements of a pass for different languages? If I see it correctly, this would require a `Map<KClass<out Pass<*>>, Map<KClass<out Language<*>>, KClass<out Pass<*>>>>`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -217,7 +215,9 @@ private constructor(
         private val includePaths = mutableListOf<Path>()
         private val includeWhitelist = mutableListOf<Path>()
         private val includeBlocklist = mutableListOf<Path>()
-        private val passes = mutableListOf<Pass>()
+        private val passes = mutableListOf<KClass<out Pass<*>>>()
+        private val replacedPasses =
```
KuechA, line 219: See comment above

KuechA (commented):
Overall, I like the new feature. I think that some more passes could operate on the `TranslationUnit` instead of the `Component`. However, it would be good to have some documentation on how to use this new system but I'm not quite sure where this belongs (wiki, readme, language-spec?).

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
oxisto: Should be addressed now, I didn't completely understand it before :) Now it is possible to have different replacements of a pass for different languages. As suggest by you, I settled on a map with the pair of (oldPass, language) as key and the replacement as value - should be most efficient for looking it up.

KuechA (approved):

Issue #767, opened 2022-04-25
Crash when resolving calls to variadic template functions


peckto:
# Stack trace

```
java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at de.fraunhofer.aisec.cpg_vis_neo4j.Application.call(Application.kt:402)
	at de.fraunhofer.aisec.cpg_vis_neo4j.Application.call(Application.kt:70)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
	at picocli.CommandLine.access$1300(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2346)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2311)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
	at picocli.CommandLine.execute(CommandLine.java:2078)
	at de.fraunhofer.aisec.cpg_vis_neo4j.ApplicationKt.main(Application.kt:442)
Caused by: java.lang.NullPointerException
	at de.fraunhofer.aisec.cpg.passes.CallResolver.getCallSignature(CallResolver.java:733)
	at de.fraunhofer.aisec.cpg.passes.CallResolver.handleTemplateFunctionCalls(CallResolver.java:557)
	at de.fraunhofer.aisec.cpg.passes.CallResolver.handleNormalCallCXX(CallResolver.java:1062)
	at de.fraunhofer.aisec.cpg.passes.CallResolver.handleNormalCalls(CallResolver.java:1024)
	at de.fraunhofer.aisec.cpg.passes.CallResolver.handleCallExpression(CallResolver.java:299)
	at de.fraunhofer.aisec.cpg.passes.CallResolver.resolve(CallResolver.java:260)
	at de.fraunhofer.aisec.cpg.helpers.SubgraphWalker$ScopedWalker.lambda$registerHandler$0(SubgraphWalker.java:408)
	at de.fraunhofer.aisec.cpg.helpers.SubgraphWalker$ScopedWalker.handleNode(SubgraphWalker.java:445)
[...]

PR #1181, opened 2023-05-25
Reduce Sonar bugs warnings

KuechA:
Reduces the usage of the `!!` operators as they can result in null pointer exceptions and it triggers numerous Sonar warnings.


KuechA:
Should also fix the problem in #767

oxisto (approved):

PR #1183, opened 2023-05-26
Make `TypeManager` not a singleton anymore

oxisto:
This PR makes the `TypeManager` not a singleton anymore. This PR also changes the following:

- [x] Introduce a new `TranslationContext` which holds all necessary managers during translation, such as the `ScopeManager`, `TypeManager` as well as the `TranslationConfiguration`.
- [x] Functions that previously took a `TranslationResult` as an input, now take a `TranslationContext` (with very few exceptions in the `TranslationManager`)
- [x] One type manager is spawned for the `TranslationResult` instead of one global singleton. This should make testing easier and potentially make it easier for component-wise type managers in the future
- [x] Replaced all occurrences of `TypeParser.createFrom` (outside of the type parser/manager) with `parseType`

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/HasType.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/HasType.kt
@@ -25,9 +25,11 @@
  */
 package de.fraunhofer.aisec.cpg.graph
 
+import de.fraunhofer.aisec.cpg.frontends.TranslationException
 import de.fraunhofer.aisec.cpg.graph.types.Type
+import java.util.Optional
 
-interface HasType {
+interface HasType : ContextProvider {
     var type: Type
 
     /**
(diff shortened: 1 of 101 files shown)


KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/HasType.kt
```diff
@@ -25,9 +25,11 @@
  */
 package de.fraunhofer.aisec.cpg.graph
 
+import de.fraunhofer.aisec.cpg.frontends.TranslationException
 import de.fraunhofer.aisec.cpg.graph.types.Type
+import java.util.Optional
 
-interface HasType {
+interface HasType : ContextProvider {
```
konradweiss, line 32: I kinda don't like  having every node that has a type, hold a reference to the context, don't we just need it in the handlers? I can see that we will have the need for multiple translation contexts associated to different nodes in one result.

oxisto: We need a reference to the context, more specifically to the type manager in each node which has a type because the we are using functions of the `TypeManager` in `setType` and so on. Previously we were using the type manager singleton, not we need a reference to the appropriate type manager in each node.

konradweiss (approved):
Looks good to me, only thing is that I don't like having the context stored with every node that has a type, but I see why we would need it.

PR #1196, opened 2023-06-07
Fixed a bug in `ScopeManager::mergeFrom`

oxisto:
There was a rather delicate bug in the merging functionality of the ScopeManager, that always resulted in a wrong merged global scope.


oxisto:
Ok this confused me... they are now tests failing with this *correct* behaviour. How the heck did the tests even pass in the first place *sigh*

PR #1197, opened 2023-06-07
Infer DFG for calls to implicit functions

KuechA:
Currently, we only infer dfg edges to calls for unresolved functions. This PR also generalizes it to implicit function declarations.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
```diff
@@ -390,9 +390,16 @@ class DFGPass(ctx: TranslationContext) : ComponentPass(ctx) {
             // Unresolved call expression
             handleUnresolvedCalls(call)
         } else if (call.invokes.isNotEmpty()) {
-            call.invokes.forEach {
-                Util.attachCallParameters(it, call.arguments)
-                call.addPrevDFG(it)
+            call.invokes.forEach { funDecl ->
+                if (funDecl.isInferred && inferDfgForUnresolvedSymbols) {
+                    if (call is MemberCallExpression && !call.isStatic) {
```
oxisto, line 395: You should probably call handleUnresolvedCalls instead of copy/pasting its content here :)

oxisto (approved):

PR #1199, opened 2023-06-11
Type system / manager overhaul

oxisto:
## Prep Work
- [x] Converting `TypeManager` to Kotlin
- [x] Introduce a new `ProblemType`, similar to `ProblemDeclaration` and `ProblemExpression`
- [x] Create two new functions `newPrimitiveType` and `newObjectType` as part of the node builders 

### C/C++ Frontend
- [x] Replace creation of primitive types with `newPrimitiveType` instead of calling the type parser
- [x] Replace all calls to `parseType` with type parsing in frontend and `newObjectType`

### Java Frontend
- [x] Replace creation of primitive types with `newPrimitiveType` instead of calling the type parser
- [x] Replace all calls to `parseType` with type parsing in frontend and `newObjectType`

### LLVM Frontend
- [x] Replace creation of primitive types with `newPrimitiveType` instead of calling the type parser
- [x] Replace all calls to `parseType` with type parsing in frontend and `newObjectType`

### TypeScript Frontend
- [x] Replace creation of primitive types with `newPrimitiveType` instead of calling the type parser
- [x] Replace all calls to `parseType` with type parsing in frontend and `newObjectType`

### Go Frontend
- [x] Replace creation of primitive types with `newPrimitiveType` instead of calling the type parser
- [x] Replace all calls to `parseType` with type parsing in frontend and `newObjectType`

### Python Frontend
- [x] Replace creation of primitive types with `newPrimitiveType` instead of calling the type parser
- [x] Replace all calls to `parseType` with type parsing in frontend and `newObjectType`

## Cleanup
- [x] Removing `TypeParser`
- [x] Cleaning up some stuff
- [x] Build C++ primitive types in a canonical way (e.g., `unsigned long long int`) to correctly parse all primitive types
- [x] Harmonise the way frontends parse/construct type information (e.g. using a `typeOf` function in the `LanguageFrontend`)


KuechA (commented):
I think it's a very good idea to let the frontends generate and parse the types as they have more (specific and relevant) information available and we do not have to assemble a string only to parse it again. I also think that adding this to a frontend shouldn't be problematic in most cases.

It would be good to have one method in the `LanguageFrontend` which has to be implemented by their instances to generate the type. Currently, they have a different name everywhere. I'd also appreciate to have a common usage pattern provided by the node builder (e.g. `pointer(primitiveType("int"))` instead of `primitiveType("int").reference(POINTER)` or so). I think this is not consistent yet.

konradweiss (approved):
LGTU

PR #1200, opened 2023-06-12
Draft: Revert IncludeDeclaration hashCode overload

peckto:
This PR reverts the changes introduced during conversion to kotlin.
The `hashCode` is not stable and results in uncontrolled recursion.


peckto:
Needs more investigation

PR #1203, opened 2023-06-12
Clean up old list of file extensions

peckto:


oxisto (approved):

konradweiss (approved):

KuechA:
There's another place where we have some hardcoded extensions for c/cpp but I'm not sure if we can get rid of this one that easily: https://github.com/Fraunhofer-AISEC/cpg/blob/cf8375d61273f69839ec3dbaad3b28e65be812f4/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt#L188

KuechA (approved):

oxisto:
> There's another place where we have some hardcoded extensions for c/cpp but I'm not sure if we can get rid of this one that easily:
> 
> https://github.com/Fraunhofer-AISEC/cpg/blob/cf8375d61273f69839ec3dbaad3b28e65be812f4/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt#L188

Since `CLanguage` is not a part of core anymore, we would need reflection to do it, which is probably a bit tricky.

Issue #935, opened 2022-10-09
.json format of CPG's through this library is not understandable and not giving full information


anushka192001:
lets generating .JSON format of CPG of this sum.cpp file
![image](https://user-images.githubusercontent.com/63954873/194773884-32350a4f-f420-4d9c-8192-4b83bc95f0a8.png)

using cpg-neo4j subproject
![image](https://user-images.githubusercontent.com/63954873/194774460-f1bdad92-cce9-4bfe-94c2-481c4c493b1f.png)

downloading json from here
![image](https://user-images.githubusercontent.com/63954873/194774434-553b9d72-3d8d-402a-8a65-47f09b311d97.png)

![image](https://user-images.githubusercontent.com/63954873/194774595-5abd571f-7c87-4aa1-849b-31e22f98157c.png)

like as you can see this .json file , this is not giving even edges and relations information and completely not understandable . 

any other better way to get json file with complete information thought this library , i urgently need json files to train my machine learning model .

KuechA:
Your query only matches on nodes but not on relationships between nodes (the edges). That's why the json only includes all the nodes but no edges. To fix this, you have to change the query. E.g., `MATCH (n) OPTIONAL MATCH (n)-[r]-(m) RETURN n,r,m` matches on all nodes and the edges. The json will now consist of a list of json objects with 3 elements (n, r, m), where the element "n" is the start node of an edge, "r" is the edge and "m" is the end node of the edge.

Concerning the interpretation of the output, the "identity" is an ID of the node, the "labels" determine the type of nodes, the "properties" are the properties of the node (please check the documentation of the graph). It just lists all the information which is available in the nodes.

anushka192001:
@KuechA can you suggest me , which format is best for generating embeddings like graphson(json) or graphml(xml) ?

KuechA:
We implemented a json export functionality in #1215 and thus close this issue for now.

Issue #776, opened 2022-05-12
A attempt to export cpg graph（have seen#247, solution miss）


HickeyHsu:
I've been following this amazing project for a long time.  I hope to extract cpg to files(.dot/.gml/.json) for deep learning usage. I noticed that #247 gave an solution"cpg-to-dot". However, this page was gone.

I'm wondering if there will be an official solution of cpg-export?

Thanks a lot!

HickeyHsu:
I try to convert the result into jgrapht format, which can be easily export as different files. But I'm still struggling.

oxisto:
> > I try to convert the result into jgrapht format, which can be easily export as different files. But I'm still struggling.
> 
> [https://github.com/HickeyHsu/SCPG](url) I make a demo to convert cpg into jgrapht format. Not a perfect attempt.

Cool idea!

KuechA:
Closed by #1215

Issue #247, opened 2020-09-25
Export cpg graph


peckto:
At the moment, the cpg graph can only be used within the cpg library.
(We have the `cpg-vis-neo4j`, but this is only feasible for small graphs)
For compatibility reasons, it would be nice, to have the capability, to export/serialize the graph to a machine readable format.
This allows a bulk import of the cpg to other tools/databases.
When discussing the right format to use, we need to take the properties of the cpg into account.
One standard format for graph representation is `GraphML`. 
But as far as I know, `GraphML` does not support edge properties.
For this reason, I would suggest a `csv` based format, similar to the format accepted by the neo4j database:
https://neo4j.com/docs/operations-manual/current/tutorial/import-tool/

I submit this issue to the cpg repo for now, because I do not know a better place.

This issue may depend on #246

oxisto:
I would suggest a separate project for this or an extension of `cpg-vis-neo4j`, we want to keep this project solely as a library. The main reason for this is the difficult minefile of different licenses for different graph connectors. This project is and should stay as Apache 2.0.

peckto:
I see your point with the license and the library.
I'm still not sure, if the `csv` export should replace the `cpg-vis-neo4j` or if there should be something like `cpg-vis-csv`.

anon767:
https://github.com/anon767/cpg-to-dot

I just made a tool to extract the cpg into the Dot format. Maybe this helps.

oxisto:
> https://github.com/anon767/cpg-to-dot
> 
> I just made a tool to extract the cpg into the Dot format. Maybe this helps.

Very interesting! BTW, i noted that you were using a quite old version of cpg (`2.x`). The current one is `3.1.0`.

anon767:
Ah thank you! I updated it to use the current cpg version :)

aifight:
> https://github.com/anon767/cpg-to-dot
> 
> I just made a tool to extract the cpg into the Dot format. Maybe this helps.

Great solution! But I noticed the link is not found, could you share it?

oxisto:
Related to https://github.com/Fraunhofer-AISEC/cpg/issues/246

KuechA:
Closed by #1215

PR #1215, opened 2023-06-15
Export cpg as json via Neo4j OGM

peckto:
This PR adds the functionality to export the cpg as json.
The serialization is done via the Neo4j OGM.
Basically the same as when importing to Neo4j Database, just without Database.
So the json export will suffer from known problems, like recursion and bigInt.

Json format might be useful for machine learning or other data science projects.

Proposed json format:
```json
{
"nodes": [],
"edges": []
}
```
Example json file: 
[cpg.txt](https://github.com/Fraunhofer-AISEC/cpg/files/11757931/cpg.txt)

- [x] compare json export with data from Neo4j Database import
- [x] add test case
- [x] support pre-save hooks

### DIFF
--- a/cpg-neo4j/README.md
+++ b/cpg-neo4j/README.md
@@ -4,7 +4,7 @@ A simple tool to export a *code property graph* to a neo4j database.
 
 ## Requirements
 
-The application requires Java 11 or higher.
+The application requires Java 17 or higher.
 
 ## Build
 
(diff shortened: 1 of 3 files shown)


peckto:
The current implementation does not create any Neo4j Session object, which allows the json export to be used without a connection to the database server, which is of course desirable. 
This has unfortunately the side effect, that pre-save hooks are not executed, because they are configured in the session.
I have currently two ideas how this could be fixed:
a) create some dummy session, without actual database connection
b) configure pre-save hooks without session

KuechA:
Related issues: #247 #776  #935 . We may want to enable customizing the output format somehow.

peckto:
> We may want to enable customizing the output format somehow

Good point! I created a separate function to export in the proposed json format (`exportSimpleJson`) .
In future, here one could step in to support other formats.

konradweiss:
- [ ] Make a pass `PrepareForNeo4jSerializationPass`
- [ ] Turn it on as a default pass for the Neo4j module.

cpg-neo4j/README.md
```diff
@@ -14,57 +14,95 @@ Build (and install) a distribution using Gradle
...
 ## Known issues:
 
 - While importing sufficiently large projects with the parameter <code>--save-depth=-1</code> 
         a <code>java.lang.StackOverflowError</code> may occur.
     - This error could be solved by increasing the stack size with the JavaVM option: <code>-Xss4m</code>
+    - Otherwise the depth must be limited (e.g. 3 or 5)
 
 - While pushing a constant value larger than 2^63 - 1 a <code>java.lang.IllegalArgumentException</code> occurs.
-
-## Experimental Languages
```
konradweiss, line 58: Why did we remove the lines for experimental languages?

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -232,6 +253,94 @@ class Application : Callable<Int> {
     private val passList: List<String>
         get() = passClassList.mapNotNull { it.simpleName }
 
+    private val packages: Array<String> =
+        arrayOf("de.fraunhofer.aisec.cpg.graph", "de.fraunhofer.aisec.cpg.frontends")
+
+    fun serializeCpgViaOGM(
```
konradweiss, line 259: I would document this function for better maintainability. Also I would rename some function as I was confused and that might happen to others IMO. Here `translateCPGToOGMBuilders`?

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -232,6 +253,94 @@ class Application : Callable<Int> {
...
+        val newNodeBuilders =
+            (getNewNodeBuilders[compiler] as? ArrayList<*>)?.filterIsInstance<DefaultNodeBuilder>()
+        val newRelationshipBuilders =
+            (getNewRelationshipBuilders[compiler] as? ArrayList<*>)?.filterIsInstance<
+                DefaultRelationshipBuilder
+            >()
+        return newNodeBuilders to newRelationshipBuilders
+    }
+
+    fun exportSimpleJson(
```
konradweiss, line 290: Here `buildJsonGraph`?

konradweiss (changes requested):
I would change some names and add more explicit explanations in the form of function comments. Just to make it a bit more maintainable. I did also think whether or not it would be good to extract the export to json functionality into its own class to not overload the `Application.kt` but for now its fine so that is optional.

cpg-neo4j/README.md
peckto: The option does no longer exists. As far as I understand, experimental language frontends are treated as any other frontend, as all are now optional?

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
peckto: done

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
peckto: done

cpg-neo4j/README.md
KuechA: Correct. Each frontend is basically optional

konradweiss (approved):

PR #1224, opened 2023-06-23
Convert Neo4j pre-save hooks to cpg pass

peckto:
In preparation for #1215

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PrepareSerialization.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PrepareSerialization.kt
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2023, Fraunhofer AISEC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


oxisto (commented):
Looks good. Maybe just a quick discussion: Should we have this pass in `cog-core` or in `cpg-neo4j`?

oxisto:
Ok it seems are test design for neo4j is really ... crap. We are setting up our own translation result here in the test instead of re-using the function that exists in the `Application`:

https://github.com/Fraunhofer-AISEC/cpg/blob/182b9ec991e484db543296c2f2fc5e37968c5c27/cpg-neo4j/src/test/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/ApplicationTest.kt#L47-L62

@peckto Can you maybe merge this as part of this PR so that the test is actually using the prepared TR of the application?

oxisto (approved):

peckto:
I reworked the `ApplicationTest` to use the `setupTranslationConfiguration`, including the arg parse interface.

peckto:
> Looks good. Maybe just a quick discussion: Should we have this pass in `cog-core` or in `cpg-neo4j`?

I don't have a strong opinion on this.
Maybe the pass is (in future) also helpful for other serialization methods?
We could also rename the pass to something more generic like `PrepareSerialization` or similar.

peckto:
The pass is executed before `FilenameMapper`, as `FilenameMapper` is last, this will be second last.

oxisto:
> > Looks good. Maybe just a quick discussion: Should we have this pass in `cog-core` or in `cpg-neo4j`?
> 
> I don't have a strong opinion on this. Maybe the pass is (in future) also helpful for other serialization methods? We could also rename the pass to something more generic like `PrepareSerialization` or similar.

Renaming it to `PrepareSerialization` would be a good idea.

oxisto (approved):

PR #1227, opened 2023-06-29
Add Program Dependency Graph

seelchen:
This PR adds the Program Dependency Graph (PDG). A new pass is introduced which collects the DFG and CDG edges of each node to create its PDG edges.


KuechA:
Closes #1066

seelchen:
On hold until the direction of PropertyEdges stored in `prevPDGEdges` is decided.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ProgramDependencyGraphPass.kt
```diff
@@ -0,0 +1,66 @@
...
+import de.fraunhofer.aisec.cpg.graph.*
+import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
+import de.fraunhofer.aisec.cpg.passes.order.DependsOn
+import de.fraunhofer.aisec.cpg.processing.IVisitor
+import de.fraunhofer.aisec.cpg.processing.strategy.Strategy
+
+@DependsOn(ControlDependenceGraphPass::class)
+@DependsOn(DFGPass::class)
+@DependsOn(ControlFlowSensitiveDFGPass::class, softDependency = true)
+class ProgramDependencyGraphPass(ctx: TranslationContext) : TranslationUnitPass(ctx) {
```
oxisto, line 38: Small nitpick: It's a program *dependence* graph, not dependency graph. Also missing javadoc for this class

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ProgramDependencyGraphPass.kt
```diff
@@ -0,0 +1,66 @@
...
+    private val pdgSetter =
```
oxisto, line 39: setter might be a misleading name for this variable.

oxisto:
> On hold until the direction of PropertyEdges stored in `prevPDGEdges` is decided.

We should probably also store the information whether this is a data flow or a control dependence edge in the PDG edge

seelchen:
We decided that the direction of the dependence edges should be the same as the direction of the flow edges

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ProgramDependencyGraphPass.kt
seelchen: I couldn't think of a good name so I just changed it to `visitor`.

seelchen:
> > On hold until the direction of PropertyEdges stored in `prevPDGEdges` is decided.
> 
> We should probably also store the information whether this is a data flow or a control dependence edge in the PDG edge

I added a new edge Property called DEPENDENCE and a new enum class to differentiate between the dependence types

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -208,6 +206,24 @@ open class Node : IVisitable<Node>, Persistable, LanguageProvider, ScopeProvider
     @PopulatedByPass(DFGPass::class, ControlFlowSensitiveDFGPass::class)
     var nextDFG: MutableSet<Node> by PropertyEdgeSetDelegate(Node::nextDFGEdges, true)
 
+    /** Outgoing Program Dependence Edges. */
+    @PopulatedByPass(ProgramDependenceGraphPass::class)
+    @Relationship(value = "PDG", direction = Relationship.Direction.OUTGOING)
+    var nextPDGEdges: MutableList<PropertyEdge<Node>> = mutableListOf()
```
KuechA, line 212: I think we should make it a `Set` in accordance with the DFG edges

seelchen: I'm not sure what the conclusion for the DFG edges was, should I use `Sets` for both `nextPDGEdges` and `nextPDG` or only for `nextPDG`?

KuechA: The conclusion was that in some cases (e.g. DFG, EOG, CDG, PDG) `Set`s make more sense because there's no ordering and multiple identical edges don't make sense. We will migrate those in the (near) future. In other cases we will keep `List`s. So I'd suggest to use a `Set` here right away.

KuechA: The conclusion was that in some cases (e.g. DFG, EOG, CDG, PDG) `Set`s make more sense because there's no ordering and multiple identical edges don't make sense. We will migrate those in the (near) future. In other cases we will keep `List`s. So I'd suggest to use a `Set` here right away.

oxisto (approved):
The tests follow a bit „unusual“ convention, with compared to the rest of the project but this seems to be ok

PR #1233, opened 2023-07-07
Change DFG to use PropertyEdges

seelchen:
This PR changes the DFG to use PropertyEdges instead of storing the Nodes directly.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
@@ -39,6 +39,7 @@ import de.fraunhofer.aisec.cpg.graph.edge.Properties
 import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdge
 import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdge.Companion.unwrap
 import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdgeDelegate
+import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdgeSetDelegate
 import de.fraunhofer.aisec.cpg.graph.scopes.GlobalScope
 import de.fraunhofer.aisec.cpg.graph.scopes.RecordScope
 import de.fraunhofer.aisec.cpg.graph.scopes.Scope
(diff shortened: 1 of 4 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -164,13 +165,25 @@ open class Node : IVisitable<Node>, Persistable, LanguageProvider, ScopeProvider
...
     @PopulatedByPass(DFGPass::class, ControlFlowSensitiveDFGPass::class)
-    var prevDFG: MutableSet<Node> = HashSet()
+    var prevDFGEdges: MutableList<PropertyEdge<Node>> = mutableListOf()
+        internal set
 
+    /** Virtual property for accessing [prevDFGEdges] without property edges. */
     @PopulatedByPass(DFGPass::class, ControlFlowSensitiveDFGPass::class)
-    @Relationship(value = "DFG")
-    var nextDFG: MutableSet<Node> = HashSet()
+    var prevDFG: MutableSet<Node> by PropertyEdgeSetDelegate(Node::prevDFGEdges, false)
```
oxisto, line 176: Do we really want to keep this as a `Set`, when the internal representation is a list anyway? Then we would not need the additional delegate.

KuechA: This actually makes me wonder why the other fields are `List`s and we allow duplicate entries (i.e., we don't filter that in the `add...` methods as far as I remember).

oxisto: Good point. Probably all of them should be a set instead of a list?

KuechA: Probably. But this could lead to a bigger changes and might even be API breaking for users

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
KuechA: We will keep the `Set` here and later review other edges

PR #1268, opened 2023-07-29
Overhaul of type propagation

### LINKED ISSUE #204: Heuristic for subtypes

### LINKED ISSUE #705: Overhaul of the type system
```[tasklist]
### Tasks
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/29
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/691
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/204
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/189
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/28
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/512
```

### LINKED ISSUE #1071: Variable type confusion after assignment
# Unexpected behavior

The type of a variable is changed based on the type of the assigned variable, not on the variable type itself.

# Code of interest

Different forms of this behavior have been observed:

1)

```c++
int main() {
        int i;
        short s1 = 1;
        short s2 = i;
        return 0;
}
```

image

2)

```c++
int main() {
	int i;
	short s1 = 1;
	short s2 = 1;
	s2 = i;
	return 0;
}
```

image

3)

```c++
int main() {
	int i;
	short s1;
	short s2 = 1;
	s1 = i;
	s1 = s2;
	return 0;
}
```
image

None

None

### LINKED ISSUE #1260: Use `AssignExpression` in all frontends
We should use the `AssignExpression` instead of a `BinaryOperator` with a `==`

```[tasklist]
### Tasks
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/1263
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/1269
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/1262
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/1261
- [x] Use `AssignExpression` in TypeScript
- [x] Use `AssignExpression` in Go
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/1071
```

### LINKED ISSUE #1262: Use `AssignExpression` in Python

### LINKED ISSUE #1263: Use `AssignExpression` in C++

### LINKED ISSUE #1269: Use `AssignExpression` in Java

### LINKED ISSUE #1270: Implement logic for numeric-based type propagation based on bit width

oxisto:
This PR is a major overhaul of the existing type listeners. This should greatly improve type parsing. The PR is actually slightly too big, but unfortunately a lot of things were dependent on each other and could not be fixed in smaller steps.

## Changes

- Renamed `TypeListener` into `TypeObserver` and made its documentation a lot clearer
- Looked carefully at every `Node`, to see whether we really need a `TypeObserver` and (hopefully) removed all loops in type propagation
- Smaller changes and fixes to the type handling of the `CXXLanguageFrontend`
- Replaced all usages of a `BinaryOperator` using the `=` operator with an `AssignExpression`. Creating a binary operator with `=` or a compound assignment will now throw a `TranslationException`
- Added support for compound assignments using `AssignExpression` in the `ValueEvaluator` and `MultiValueEvaluator`
- Removed the type cache that was used in parallel parsing. Now, the type system is active even if parallel parsing is used. This probably needs a bit more testing, if this does not lead to any concurrent modification errors, but for now it seems to be fine
- With the removal of the type caching, I also got rid of the slightly weird `_type` variables and the `setType` function. Now only a `type` variable exists and a Kolin setter is used to inform the type observers.
- Moved the creation of DFG edges of an `InitializerListExpression` back to the `DFGPass` instead of the class itself. It was moved there because of performance issues, which seem to be resolved now.
- Introduced a new shortcut that allows to "invoke" a list of nodes using a Kotlin class and an index (which defaults to 0). This can be used to retrieve a certain index in a node list and cast it to a specific class - or return null if its not castable. For example: `assign.lhs<DeclaredReferenceExpression>()` will retrieve the first element in `lhs` of an `AssignExpression` and cast it to `DeclaredReferenceExpression?`.

## Fixes

Fixes #204
Fixes #705 
Fixes #1071 
Fixes #1260 
Fixes #1262 
Fixes #1263
Fixes #1269 
Fixes #1270


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/HasType.kt
```diff
@@ -23,91 +23,116 @@
...
 import de.fraunhofer.aisec.cpg.frontends.TranslationException
-import de.fraunhofer.aisec.cpg.graph.types.Type
-import java.util.Optional
+import de.fraunhofer.aisec.cpg.graph.ContextProvider
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.declarations.ValueDeclaration
+import java.util.*
 
 interface HasType : ContextProvider {
+
```
KuechA, line 36: I understand the difference between `declaredType` and `assignedType` but I keep wondering what `type` should be used for

oxisto: I will probably merge declaredType with type or type is an alias for declared type? not sure yet...

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/ExpressionBuilder.kt
```diff
@@ -617,7 +620,7 @@ fun <T> Literal<T>.duplicate(implicit: Boolean): Literal<T> {
     duplicate.code = this.code
     duplicate.location = this.location
     duplicate.locals = this.locals
-    duplicate.possibleSubTypes = this.possibleSubTypes
+    // duplicate.possibleSubTypes = this.possibleSubTypes
```
KuechA, line 623:
```suggestion
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationContext.kt
```diff
@@ -45,5 +47,8 @@ class TranslationContext(
      * The type manager is responsible for managing type information. Currently, we have one
      * instance of a [TypeManager] for the overall [TranslationResult].
      */
-    val typeManager: TypeManager
+    val typeManager: TypeManager,
+
+    // TODO(oxisto): merge to statistics holder
+    val typeObserverInvocations: AtomicInteger = AtomicInteger(0)
```
KuechA, line 53: What should this be used for? I think it was a way to exit loops of type propagations

cpg-neo4j/build.gradle.kts
```diff
@@ -58,5 +58,7 @@ dependencies {
 
     // Command line interface support
     api(libs.picocli)
+    testImplementation(project(mapOf("path" to ":cpg-core")))
+    testImplementation(project(mapOf("path" to ":cpg-core")))
```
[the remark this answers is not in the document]
oxisto: It should not be there at all :)

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/enhancements/InferenceTest.kt
```diff
@@ -27,41 +27,48 @@ package de.fraunhofer.aisec.cpg.enhancements
...
+
+            val valueField = record.fields["value"]
+            assertNotNull(valueField)
+            assertLocalName("int", valueField.type)
+
+            val nextField = record.fields["next"]
+            assertNotNull(nextField)
+            assertLocalName("T*", nextField.type)
+
+            assertTrue(result.finalCtx.typeObserverInvocations.get() < 30)
```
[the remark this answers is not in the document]
oxisto: random. should be removed, was for debugging only

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
```diff
@@ -272,14 +272,9 @@ class DFGPass(ctx: TranslationContext) : ComponentPass(ctx) {
...
      * this expression.
-     *
-     * TODO: This change seems to have performance issues!
      */
     protected fun handleInitializerListExpression(node: InitializerListExpression) {
-        node.initializers.forEach {
-            it.registerTypeListener(node)
-            node.addPrevDFG(it)
-        }
+        node.initializers.forEach { node.addPrevDFG(it) }
```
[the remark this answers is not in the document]
oxisto: that is already remove from `InitializerListExpression` and it seems to be fine now

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -145,33 +148,25 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : TranslationUni
...
-                doubleState.declarationsState[writtenDecl] = PowersetLattice(setOf(currentNode.lhs))
-            }
         } else if (isCompoundAssignment(currentNode)) {
             // We write to the lhs, but it also serves as an input => We first get all previous
             // writes to the lhs and then add the flow from lhs and rhs to the current node.
 
             // The write operation goes to the variable in the lhs
-            writtenDecl =
-                ((currentNode as BinaryOperator).lhs as? DeclaredReferenceExpression)?.refersTo
+            val lhs = currentNode.lhs.singleOrNull()
```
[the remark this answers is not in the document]
oxisto: Yes, a compound assignment only works with one lhs

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/BinaryOperator.kt
```diff
@@ -180,15 +114,25 @@ open class BinaryOperator :
...
+        if (operatorCode == ".*" || operatorCode == "->*" && src === rhs) {
+            // Propagate the function pointer type to the expression itself. This helps us later in
+            // the call resolver, when trying to determine, whether this is a regular call or a
+            // function pointer call.
+            this.type = newType
+        } else {
+            // Otherwise, we have a special language-specific function to deal with type propagation
+            val type = language?.propagateTypeOfBinaryOperation(this)
+            if (type != null) {
+                this.type = type
```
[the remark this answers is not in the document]
oxisto: If its null, there is pretty much nothing we can do?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationContext.kt
oxisto: We can probably remove it

cpg-language-python/src/main/python/CPGPython/_statements.py
```diff
@@ -599,15 +608,17 @@ def bar(self):
...
                 if rhs is not None:
+                    self.log_with_loc("this")
                     v = DeclarationBuilderKt.newVariableDeclaration(
                         self.frontend, lhs.getName(),
-                        rhs.getType(),
+                        TypeBuilderKt.autoType(self.frontend),
                         self.get_src_code(stmt),
                         False)
                 else:
+                    self.log_with_loc("or that")
```
maximiliankaul, line 618:
```suggestion

```

cpg-language-python/src/main/python/CPGPython/_statements.py
```diff
@@ -599,15 +608,17 @@ def bar(self):
                     "Could not resolve -> creating a new variable for: %s"
                     % (lhs.getName()))
                 if rhs is not None:
+                    self.log_with_loc("this")
```
maximiliankaul, line 611:
```suggestion
```

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/MultiValueEvaluator.kt
```diff
@@ -92,6 +94,43 @@ class MultiValueEvaluator : ValueEvaluator() {
...
+            for (lhs in lhsValue) {
+                if (rhsValue is Collection<*>) {
+                    result.addAll(rhsValue.map { r -> computeBinaryOpEffect(lhs, r, node) })
+                } else {
+                    result.add(computeBinaryOpEffect(lhs, rhsValue, node))
+                }
+            }
+        } else {
+            result.addAll(
+                (rhsValue as Collection<*>).map { r -> computeBinaryOpEffect(lhsValue, r, node) }
```
maximiliankaul, line 127: a comment would be nice here. this is a bit confusing at first.
// `rhsValue` is a `Collection<*>` because of the check above and because `lhsValue` is not a `Collection<*>`

maximiliankaul (approved):
Python: LGTM.
Other changes as far as I understand them :+1:

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/MultiValueEvaluator.kt
oxisto: This is copy paste code from the value evaluators magic wizard @KuechA ;) Can you maybe add a comment here, @KuechA ?

KuechA: I added some comments to explain this craziness

PR #1279, opened 2023-08-10
Copy objects less frequently

### LINKED ISSUE #1194: StackOverflowError in subtype listening system
Hey everyone, 

when trying to launch the cpg-neo4j, I get a `java.lang.StackOverflowError`. 
See the full log output [here](https://github.com/Fraunhofer-AISEC/cpg/files/11663656/log.txt).
[This](https://github.com/Fraunhofer-AISEC/cpg/files/11663680/code.txt) is the file I supplied as parameter. 

With `cpg-console`, the error looks slightly different: 

```
[...]
17:57:00,100 WARN  Util code.c:9574:50: Did not find a declaration for int64_t
17:57:00,102 WARN  Util code.c:9580:57: Did not find a declaration for int64_t
17:57:00,934 WARN  Util code.c:9634:59: Did not find a declaration for int64_t
18:12:59,259 INFO  MeasurementHolder TranslationManager: Translation into full graph done in 986182 ms
java.util.concurrent.ExecutionException: java.lang.StackOverflowError
        at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
        at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
        at Line_18.<init>(Line_18.kts:1)
java.lang.NullPointerException: Cannot invoke "Line_18.getResult()" because "this.$$earlierScripts[17]" is null
        at Line_19.<init>(Line_19.kts:1)
```

Thanks for your help!

KuechA:
Currently, the EOG iteration copies the state for each step in the EOG. This consumes quite some resources and it would be great to copy the objects less frequently. One idea is to copy them only when entering/exiting a new basic block and otherwise operate on the same object.

Fixes #1194

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/LabelStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/LabelStatement.kt
@@ -51,9 +51,9 @@ class LabelStatement : Statement(), StatementHolder {
     }
 
     override var statementEdges: MutableList<PropertyEdge<Statement>>
-        get() = subStatement?.let { wrap(listOf(it), this) } ?: mutableListOf()
+        get() = subStatement?.let { PropertyEdge.wrap(listOf(it), this) } ?: mutableListOf()
         set(value) {
-            subStatement = unwrap(value).firstOrNull()
+            subStatement = PropertyEdge.unwrap(value).firstOrNull()
         }
 
     override fun equals(other: Any?): Boolean {
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/LabelStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/LabelStatement.kt
@@ -51,13 +51,10 @@ class LabelStatement : Statement(), StatementHolder {
     }
 
     override var statementEdges: MutableList<PropertyEdge<Statement>>
-        get() =
-            subStatement?.let { mutableListOf(PropertyEdge(this, it, mutableMapOf())) }
-                ?: mutableListOf()
+        get() = subStatement?.let { wrap(listOf(it), this) } ?: mutableListOf()
         set(value) {
-            subStatement = value.firstOrNull()?.end
+            subStatement = unwrap(value).firstOrNull()
         }
-
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
         if (other !is LabelStatement) return false
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/LabelStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/LabelStatement.kt
@@ -26,6 +26,8 @@
 package de.fraunhofer.aisec.cpg.graph.statements
 
 import de.fraunhofer.aisec.cpg.graph.AST
+import de.fraunhofer.aisec.cpg.graph.StatementHolder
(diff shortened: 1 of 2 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -248,9 +248,18 @@ inline fun <reified K : Node, V> iterateEOG(
     while (worklist.isNotEmpty()) {
         val (nextNode, state) = worklist.pop()
 
-        val newState = transformation(nextNode, state.duplicate(), worklist)
+        val insideBB =
+            (nextNode.nextEOG.size == 1 &&
+                nextNode.prevEOG.size == 1 &&
+                nextNode.prevEOG.first().nextEOG.size == 1)
```
oxisto, line 254: you could use `single` or `singleOrNull` instead of checking for size 1 and using first()

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/EOGWorklist.kt
```diff
@@ -248,9 +248,18 @@ inline fun <reified K : Node, V> iterateEOG(
     while (worklist.isNotEmpty()) {
         val (nextNode, state) = worklist.pop()
 
-        val newState = transformation(nextNode, state.duplicate(), worklist)
+        val insideBB =
```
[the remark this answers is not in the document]
KuechA: This should check if we're not near the beginning/end of a basic block (i.e., there are no merge points or branches of the EOG nearby). If that's the case, we just parse the whole basic block and do not want to duplicate the state. Near the beginning/end, we do want to copy the state to avoid terminating the iteration too early by messing up with the state-changing checks.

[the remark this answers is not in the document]
oxisto: Can you maybe add this as a comment to the variable. Otherwise it is confusing because we don’t actually do any check if we are in a compound statement or similar but it’s purely based on the “entry” within the block.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/LabelStatement.kt
```diff
@@ -48,6 +50,14 @@ class LabelStatement : Statement() {
...
     }
 
+    override var statementEdges: MutableList<PropertyEdge<Statement>>
+        get() =
+            subStatement?.let { mutableListOf(PropertyEdge(this, it, mutableMapOf())) }
+                ?: mutableListOf()
+        set(value) {
+            subStatement = value.firstOrNull()?.end
+        }
+
```
oxisto, line 60: I would prefer using the `wrap` / `unwrap` function instead of doing it manually, since they already do "some" magic and might do more magic in the future.

```suggestion
override var statementEdges: MutableList<PropertyEdge<Statement>>
        get() {
            return subStatement?.let { wrap(listOf(it), this) } ?: mutableListOf()
        }
        set(value) {
            subStatement = unwrap(value).firstOrNull()
        }
```

oxisto, line 60: I would prefer using the `wrap` / `unwrap` functions here for consistency.

```suggestion
    override var statementEdges: MutableList<PropertyEdge<Statement>>
        get() =
            subStatement?.let { mutableListOf(PropertyEdge(this, it, mutableMapOf())) }
                ?: mutableListOf()
        set(value) {
            subStatement = value.firstOrNull()?.end
        }

```

oxisto (approved):

PR #1282, opened 2023-08-17
Refactor of cpg-language-python

maximiliankaul:
This PR aims at minimizing Python code in favor of Kotlin code in our CPG. This will benefit the overall experience as
- the code will be easier to maintain (debugging, similarity to other frontends, ...)
- we can get rid of a lot of file handling background magic

This goal is achieved in a similar manner as with the other frontends (i.e. cpg-language-go). We will still use Python's `ast` module to parse Python code but then map all Python objects to a Kotlin interface. All logic will be implemented in Kotlin.

Open todos:
- [ ] make everything work as good as before (or better ;))
- [ ] resolve all `TODO`s
- [ ] update README
- [ ] fix the `PythonLanguageFrontend` functions (`getCode()`, ...) -> need to store the file name somewhere
- [ ] remove all `pycodestyle` references (github runner, ...)
- [ ] remove spotless checks for Python files (as there will be no more Python files) @oxisto


oxisto:
I did a first cleanup pass on this. I removed everything which was breaking things in other language frontends. I also added initial type parsing and some type inference work, which we sill can improve later on, but it works for the tests.

PR #1285, opened 2023-08-22
Added `TupleDeclaration`

oxisto:
This PR adds a new type of `Declaration`, which represenst a tuple of variable declarations. This is primarily needed for languages, such as Go, which support the assignment of multiple values in an initializer. In this case we cannot assign the initializer of the individual variables, but assigned it to the `TupleDeclaration` instead.

Currently, the type of the `TupleDeclaration` and its elements MUST be set to the `AutoType`.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
@@ -535,10 +535,6 @@ val Node?.namespaces: List<NamespaceDeclaration>
 val Node?.variables: List<VariableDeclaration>
     get() = this.allChildren()
 
-/** Returns all [TupleDeclaration] children in this graph, starting with this [Node]. */
-val Node?.tuples: List<TupleDeclaration>
-    get() = this.allChildren()
-
 /** Returns all [Literal] children in this graph, starting with this [Node]. */
 val Node?.literals: List<Literal<*>>
     get() = this.allChildren()
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
@@ -535,6 +535,10 @@ val Node?.namespaces: List<NamespaceDeclaration>
 val Node?.variables: List<VariableDeclaration>
     get() = this.allChildren()
 
+/** Returns all [TupleDeclaration] children in this graph, starting with this [Node]. */
+val Node?.tuples: List<TupleDeclaration>
+    get() = this.allChildren()
+
 /** Returns all [Literal] children in this graph, starting with this [Node]. */
 val Node?.literals: List<Literal<*>>
     get() = this.allChildren()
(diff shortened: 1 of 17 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -535,6 +535,10 @@ val Node?.namespaces: List<NamespaceDeclaration>
 val Node?.variables: List<VariableDeclaration>
     get() = this.allChildren()
 
+/** Returns all [TupleDeclaration] children in this graph, starting with this [Node]. */
+val Node?.tuples: List<TupleDeclaration>
```
KuechA, line 539: This might be a bit misleading as 1) not all of the declarations where we use this are actually tuples and 2) one could think we mean variables of type tuple (e.g., a variable `a: Tuple`). I'm wondering if this is really relevant/would be frequently used

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/TupleDeclaration.kt
```diff
@@ -0,0 +1,79 @@
...
+import de.fraunhofer.aisec.cpg.graph.AST
+import de.fraunhofer.aisec.cpg.graph.Name
+import de.fraunhofer.aisec.cpg.graph.newTupleDeclaration
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.CallExpression
+import de.fraunhofer.aisec.cpg.graph.types.AutoType
+import de.fraunhofer.aisec.cpg.graph.types.TupleType
+
+/**
+ * This declaration models a tuple of different [VariableDeclaration] nodes. This is primarily used
+ * in languages that support multiple assignments, such as Go as part of a top-level declaration.
```
KuechA, line 37:
```suggestion
 * in languages that support multiple assignments in a declaration, such as Go.
```

docs/docs/CPG/specs/dfg.md
```diff
@@ -54,6 +54,49 @@ Scheme:
     expression -- DFG --> node;
 ```
 
+
+## AssignExpression (`operatorCode: =`)
```
KuechA, line 58: @oxisto I guess we also have to add a documentation for `+=`, etc., right? And we should probably delete the part in the `BinaryOperator` even if I believe we do not prevent a user from modeling an assignment that way (I think we can't block `=` as an operator in case some language uses this in another way)

oxisto: We actually *do* block `=` and all other compound operators in the `BinaryOperator` by throwing a translation exception. So yes, it should be removed from `BinaryOperator` and compound assignments should be added to `AssignExpression`.

KuechA: ok...I tried to adapt the specs. Maybe, throwing the translation exception might not be a good idea for each language... 

Do we support `+=` with multiple elements on `lhs` and `rhs`?

oxisto: No, compound assignments are only allowed with one lhs and rhs

KuechA (approved):

Issue #902, opened 2022-09-07
Bring tutorial up-to-date with new query extensions and changes to query API


fwendland:
# Stack trace

When using the console for `v4.6.2` with the example file `cpg-console/src/test/resources/array.cpp` I'm getting:
```
[25] var expr = result.all<ArraySubscriptionExpression>().map { Triple(
        it.subscriptExpression.evaluate() as Int,
        it.arrayExpression.dfgFrom<ArrayCreationExpression>().first().capacity,
        it
    ) }
java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')
	at Line_26.<init>(Line_26.kts:2)
```

Thinking it might be a type change not reflected in the tutorial yet, I've tried changing the type to `Long`:
```
[26] var expr = result.all<ArraySubscriptionExpression>().map { Triple(
        it.subscriptExpression.evaluate() as Long,
        it.arrayExpression.dfgFrom<ArrayCreationExpression>().first().capacity,
        it
    ) }
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap')
	at Line_27.<init>(Line_27.kts:2)
```
... producing the opposite exception.

# Unexpected behavior

_Not applicable_

# Code of interest

Part of this repo

# Extensions to the library, e.g. added or changed LanguageFrontends or Passes

Used default console configuration

# Changes to the configuration

None

oxisto:
We also need to re-work the tutorial to incorporate changes of https://github.com/Fraunhofer-AISEC/cpg/pull/900

PR #1290, opened 2023-08-24
More precise types in arithmetic evaluation

oxisto:
Previously, we just casted most of our numeric types to `Long` when doing arithmetic operations in the `ValueEvalutor`. This PR tries to optimize that by introducing a number of extension functions that under the hood call the necessary Kotlin functions to properly deal with the resulting type based on the bit width.


KuechA (approved):

KuechA:
We may want to add a note to our gh page and say that the evaluation could return different types of numbers. This means that it might be required/useful to transfer the value e.g. via `toLong()` (this is related to #902 )

PR #1303, opened 2023-08-28
Renaming nodes to be shorter and more consistent

konradweiss:
This PR introduces some renamings with the aim to make the node names more intuitive and more fluently usable when working with the query language and any console scripting interface. The main idea i followed was to name more consistently and omit unnecessary details

- `ParamVariableDeclaration` to `ParameterDeclaration`
- `DeclaredReferenceExpression` to `Reference` to make it shorter and omitting the `Expression` consistent with `Literal`.
- `ArraySubscriptionExpression` to `SubscriptionExpr` shorten further and because we use it for other subscriptable constructs that are no arrays.
- `ClassTemplateDeclaration` to `RecordTemplateDeclaration` to be consistent with `RecordDeclaration`
- `ArrayCreationExpression` to `NewArrayExpression`: to better reflect that is only used in the new context and can be 
- Rename `CompoundStatement` to `Block`: I think would be more intuitive.
- Unify `CompoundStatement` and `CompoundStatementExpression`
- `UsingDirective` to `UsingDeclaration` due to it being a declaration
- `ExplicitConstructorInvocation` to `ConstructorCallExpression` to remove unnecessary explicit and invocation and signal that it is a call expression
- `TypeParamDeclaration` to `TypeParameterDeclaration` for consistency


oxisto:
Personally, I wouldn't shorten `Statement` to `Stmt`, etc. I kinda like that our classes are complete words and not just weird abbreviations.

I do agree with

- `Reference`
- `ParameterDeclaration`
- `SubscriptionExpression`
- `CompoundStatement` -> `BlockStatement`

KuechA:
`Tmp` for `Template` is too confusing to me, the rest sounds reasonable. I wouldn't be declined to abbreviate Statement, Declaration, Expression since they occur quite often but I'd also be fine with dropping this information completely. I think the reason to keep it was to know quickly whether it inherits from `Statement`, `Declaration` or `Expression` but at the same time, assigning a wrong type wouldn't be possible anyway

konradweiss:
@oxisto @KuechA I don't see us dropping the suffix of Statement, Expression, Declaration completely because there are some node names that would be misleading or collide, e.g. `CompoundStatement` and `CompoundStatementExpression` and as @KuechA its also good for developers to know what type of node it is. I would incorporate the feedback and still keep the shortening of these three names as my objection to the Decl-Declaration-Declarator Issue is that Decl always stands for Declaration and users who differentiate between Declarator, Declaration and something Declared are so involved that they know our convention. That ofc means we do not abbreviate  Declarator and Declared. Unless you have a objection @oxisto ?

KuechA:
> e.g. `CompoundStatement` and `CompoundStatementExpression` 

Is this something that we can get rid of? Why can't `Block` inherit from `Expression` (which extends `Statement`)

konradweiss:
> > e.g. `CompoundStatement` and `CompoundStatementExpression`
> 
> Is this something that we can get rid of? Why can't `Block` inherit from `Expression` (which extends `Statement`)

This would make EVERY block of statements an expression, but although I had an initial repulsion to the thought I see it as less problematic the more I think of it. @oxisto What do you think?

oxisto:
> > > e.g. `CompoundStatement` and `CompoundStatementExpression`
> > 
> > 
> > Is this something that we can get rid of? Why can't `Block` inherit from `Expression` (which extends `Statement`)
> 
> This would make EVERY block of statements an expression, but although I had an initial repulsion to the thought I see it as less problematic the more I think of it. @oxisto What do you think?

We could do it similar to the `AssignExpression`,  which is also used as a statement in 90% of the time, but it derives from expression (because of C++ of course).

We then have a property that denotes the usage as expression:
https://github.com/Fraunhofer-AISEC/cpg/blob/463ac6094bcb3217972db1f4bee576daf0905f7b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/AssignExpression.kt#L79-L84

We afaik then check in the DFG (EOG?) passes if this property is set and set an additional DFG edge to itself.

konradweiss:
> > > > e.g. `CompoundStatement` and `CompoundStatementExpression`
> > > 
> > > 
> > > Is this something that we can get rid of? Why can't `Block` inherit from `Expression` (which extends `Statement`)
> > 
> > 
> > This would make EVERY block of statements an expression, but although I had an initial repulsion to the thought I see it as less problematic the more I think of it. @oxisto What do you think?
> 
> We could do it similar to the `AssignExpression`, which is also used as a statement in 90% of the time, but it derives from expression (because of C++ of course).
> 
> We then have a property that denotes the usage as expression:
> 
> https://github.com/Fraunhofer-AISEC/cpg/blob/463ac6094bcb3217972db1f4bee576daf0905f7b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/AssignExpression.kt#L79-L84
> 
> We afaik then check in the DFG (EOG?) passes if this property is set and set an additional DFG edge to itself.

Ok, I see we ar edoing this, so I would say we just rename `CompoundStatementExpression` to `Block` and remove `CompundStatement`. Here I am fine with removing the suffix cause we use them as Expression and Statement so it does not matter that much. 

Can it work without using a variable like `usedAsExpression`?

oxisto:
> > > > > e.g. `CompoundStatement` and `CompoundStatementExpression`
> > > > 
> > > > 
> > > > Is this something that we can get rid of? Why can't `Block` inherit from `Expression` (which extends `Statement`)
> > > 
> > > 
> > > This would make EVERY block of statements an expression, but although I had an initial repulsion to the thought I see it as less problematic the more I think of it. @oxisto What do you think?
> > 
> > 
> > We could do it similar to the `AssignExpression`, which is also used as a statement in 90% of the time, but it derives from expression (because of C++ of course).
> > We then have a property that denotes the usage as expression:
> > https://github.com/Fraunhofer-AISEC/cpg/blob/463ac6094bcb3217972db1f4bee576daf0905f7b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/AssignExpression.kt#L79-L84
> > 
> > We afaik then check in the DFG (EOG?) passes if this property is set and set an additional DFG edge to itself.
> 
> Ok, I see we ar edoing this, so I would say we just rename `CompoundStatementExpression` to `Block` and remove `CompundStatement`. Here I am fine with removing the suffix cause we use them as Expression and Statement so it does not matter that much.
> 
> Can it work without using a variable like `usedAsExpression`?

I think we need some kind of property to make the differentiation in the EOG and/or DFG because probably the last statement flows to the expression itself or something. Not sure how the EOG/DFG is implemented for `CompoundStatementExpression`.

konradweiss:
> > > > > > e.g. `CompoundStatement` and `CompoundStatementExpression`
> > > > > 
> > > > > 
> > > > > Is this something that we can get rid of? Why can't `Block` inherit from `Expression` (which extends `Statement`)
> > > > 
> > > > 
> > > > This would make EVERY block of statements an expression, but although I had an initial repulsion to the thought I see it as less problematic the more I think of it. @oxisto What do you think?
> > > 
> > > 
> > > We could do it similar to the `AssignExpression`, which is also used as a statement in 90% of the time, but it derives from expression (because of C++ of course).
> > > We then have a property that denotes the usage as expression:
> > > https://github.com/Fraunhofer-AISEC/cpg/blob/463ac6094bcb3217972db1f4bee576daf0905f7b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/AssignExpression.kt#L79-L84
> > > 
> > > We afaik then check in the DFG (EOG?) passes if this property is set and set an additional DFG edge to itself.
> > 
> > 
> > Ok, I see we ar edoing this, so I would say we just rename `CompoundStatementExpression` to `Block` and remove `CompundStatement`. Here I am fine with removing the suffix cause we use them as Expression and Statement so it does not matter that much.
> > Can it work without using a variable like `usedAsExpression`?
> 
> I think we need some kind of property to make the differentiation in the EOG and/or DFG because probably the last statement flows to the expression itself or something. Not sure how the EOG/DFG is implemented for `CompoundStatementExpression`.

I will see if it is necessary

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/MultiValueEvaluator.kt
```diff
@@ -220,10 +220,7 @@ class MultiValueEvaluator : ValueEvaluator() {
      * In contrast to the implementation of [ValueEvaluator], this one can handle more than one
      * value.
      */
-    override fun handleDeclaredReferenceExpression(
-        expr: DeclaredReferenceExpression,
-        depth: Int
-    ): Collection<Any?> {
+    override fun handleDeclaredReferenceExpression(expr: Reference, depth: Int): Collection<Any?> {
```
oxisto, line 223: handleReference

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/BlockStatement.kt
```diff
@@ -38,7 +38,7 @@ import org.neo4j.ogm.annotation.Relationship
  * A statement which contains a list of statements. A common example is a function body within a
  * [FunctionDeclaration].
  */
-class CompoundStatement : Statement(), StatementHolder {
+class BlockStatement : Statement(), StatementHolder {
```
oxisto, line 41: Rename to `Block`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/Reference.kt
```diff
@@ -134,7 +133,7 @@ open class DeclaredReferenceExpression : Expression(), HasType.TypeObserver {
         if (this === other) {
             return true
         }
-        if (other !is DeclaredReferenceExpression) {
+        if (other !is Reference) {
             return false
         }
         return super.equals(other) && refersTo == other.refersTo
```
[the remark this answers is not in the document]
konradweiss: I would not remove that in this PR, its not related

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/ValueEvaluator.kt
```diff
@@ -380,10 +379,7 @@ open class ValueEvaluator(
      * Tries to compute the constant value of a reference. It therefore checks the incoming data
      * flow edges.
      */
-    protected open fun handleDeclaredReferenceExpression(
-        expr: DeclaredReferenceExpression,
-        depth: Int
-    ): Any? {
+    protected open fun handleDeclaredReferenceExpression(expr: Reference, depth: Int): Any? {
```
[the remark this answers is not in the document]
konradweiss: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/RecordTemplateDeclaration.kt
```diff
@@ -35,7 +35,7 @@ import kotlin.collections.ArrayList
 import org.neo4j.ogm.annotation.Relationship
 
 /** Node representing a declaration of a ClassTemplate */
```
KuechA, line 37: Do the comments also need to be changed?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/UsingDeclaration.kt
```diff
@@ -28,12 +28,12 @@ package de.fraunhofer.aisec.cpg.graph.declarations
 import java.util.Objects
 
 // TODO: Documentation
-class UsingDirective : Declaration() {
+class UsingDeclaration : Declaration() {
```
KuechA, line 31: Documentation missing

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/BlockScope.kt
```diff
@@ -25,10 +25,10 @@
  */
 package de.fraunhofer.aisec.cpg.graph.scopes
 
+import de.fraunhofer.aisec.cpg.graph.statements.BlockStatement
 import de.fraunhofer.aisec.cpg.graph.statements.BreakStatement
-import de.fraunhofer.aisec.cpg.graph.statements.CompoundStatement
 
-class BlockScope(blockStatement: CompoundStatement) :
+class BlockScope(blockStatement: BlockStatement) :
```
KuechA, line 31: Documentation

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/StatementHandler.kt
```diff
@@ -503,10 +516,8 @@ class StatementHandler(lang: JavaLanguageFrontend?) :
         return switchStatement
     }
 
-    private fun handleExplicitConstructorInvocation(
-        stmt: Statement
-    ): ExplicitConstructorInvocation {
-        val eciStatement = stmt.asExplicitConstructorInvocationStmt()
+    private fun handleConstructorCallExpression(stmt: Statement): ConstructorCallExpression {
+        val eciStatement = stmt.asConstructorCallExpressionStmt()
```
KuechA, line 520: eciStatement -> constructorCall?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/RecordTemplateDeclaration.kt
oxisto: Ideally we should use the "imported name" syntax in Kolin, e.g. `[ClassTemplate]` instead of `ClassTemplate`, because this way they also get renamed.

konradweiss: Updated description

oxisto:
> * `DesignatedInitializerExpression` to `InitializerExpression` : @oxisto Does this still reflect what it is, is there a better name?

I would keep it as `DesignatedInitializerExpression` for now and merge it as a combination of the initializerlist and the key value expression at some point.

oxisto (commented):
Some smaller things. Looks good overall, I support the renamings

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/MultiValueEvaluator.kt
```diff
@@ -73,17 +73,17 @@ class MultiValueEvaluator : ValueEvaluator() {
...
             is Literal<*> -> return node.value
-            is DeclaredReferenceExpression -> return handleDeclaredReferenceExpression(node, depth)
+            is Reference -> return handleReference(node, depth)
             is UnaryOperator -> return handleUnaryOp(node, depth)
             is AssignExpression -> return handleAssignExpression(node, depth)
             is BinaryOperator -> return handleBinaryOperator(node, depth)
             // Casts are just a wrapper in this case, we are interested in the inner expression
             is CastExpression -> return this.evaluateInternal(node.expression, depth + 1)
-            is ArraySubscriptionExpression -> return handleArraySubscriptionExpression(node, depth)
+            is SubscriptionExpression -> return handleSubscriptionExpression(node, depth)
```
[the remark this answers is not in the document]
konradweiss: Was thinking similarly, but did not want stay conservative, but if I have the support I will gladly rename that

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
```diff
@@ -633,61 +626,58 @@ fun LanguageFrontend<*, *>.elseIf(init: IfStatement.() -> Unit): IfStatement {
...
+    val node = newBlock()
     init(node)
     statement = node
 
     return node
 }
 
 /**
- * Creates a new [CompoundStatement] in the Fluent Node DSL and sets it to the
+ * Creates a new [BlockStatement] in the Fluent Node DSL and sets it to the
```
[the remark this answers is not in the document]
konradweiss: Done

oxisto (approved):

oxisto (approved):

KuechA (approved):

PR #1304, opened 2023-08-28
Only store one single unique `Type` nodes for each type

oxisto:
Instead of gathering individual type node objects and then squashing them back together, we just make sure that `registerType` only returns the single unique type object. To reduce the number of `Type` nodes needed as input to `registerType`, the `objectType` function does a pre-check now, if a `ObjectType` with the given name already exists, so no new `Type` object is even created at all.

This also removes the "secondary type edge" interface, which was only needed to merge extra usages of types in nodes. Now all type nodes are always referring to the single instance.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
@@ -36,8 +36,15 @@ import de.fraunhofer.aisec.cpg.graph.scopes.Scope
 import de.fraunhofer.aisec.cpg.graph.scopes.TemplateScope
 import de.fraunhofer.aisec.cpg.graph.types.*
 import java.util.*
+import java.util.concurrent.ConcurrentHashMap
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 
 class TypeManager {
+    companion object {
+        val log: Logger = LoggerFactory.getLogger(TypeManager::class.java)
+    }
+
     /**
      * Stores the relationship between parameterized RecordDeclarations (e.g. Classes using
      * Generics) to the ParameterizedType to be able to resolve the Type of the fields, since
@@ -50,8 +57,8 @@ class TypeManager {
             mutableMapOf<TemplateDeclaration, MutableList<ParameterizedType>>()
         )
 
-    val firstOrderTypes: MutableSet<Type> = Collections.synchronizedSet(HashSet())
-    val secondOrderTypes: MutableSet<Type> = Collections.synchronizedSet(HashSet())
+    val firstOrderTypes: MutableSet<Type> = ConcurrentHashMap.newKeySet()
+    val secondOrderTypes: MutableSet<Type> = ConcurrentHashMap.newKeySet()
 
     /**
      * @param recordDeclaration that is instantiated by a template containing parameterizedtypes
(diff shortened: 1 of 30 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/TypeBuilder.kt
```diff
@@ -112,9 +110,25 @@ fun LanguageProvider.objectType(name: CharSequence, generics: List<Type> = listO
...
                 "Could not create type: translation context not available"
             )
-    val type = ObjectType(name, generics, false, language)
 
-    return c.typeManager.registerType(type)
+    synchronized(c.typeManager.firstOrderTypes) {
+        // We can try to look up the type by its name and return it, if it already exists.
+        var type =
+            c.typeManager.firstOrderTypes.firstOrNull {
+                it is ObjectType && it.name == name && it.generics == generics
```
KuechA, line 118: I think that we also want to check the language

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
```diff
@@ -187,13 +190,35 @@ class TypeManager {
...
+            return t
+        }
+
         if (t.isFirstOrderType) {
-            firstOrderTypes.add(t)
-        } else {
-            secondOrderTypes.add(t)
-            registerType((t as SecondOrderType).elementType)
+            // Make sure we only ever return one unique object per type
+            if (!firstOrderTypes.add(t)) {
```
konradweiss, line 201: Does this not add the element if already contained? I can't find any Doc on KeySets saying that it functions like this.

KuechA, line 205: Could probably be optimized by using a Map and `computeIfAbsent()`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
```diff
@@ -187,13 +190,35 @@ class TypeManager {
...
+                return firstOrderTypes.filterIsInstance<T>().first { it == t }
```
konradweiss, line 202: Doesn't filterIsInstance<T> return the same type of the same type type but not the one that was found in the condition above that is equal with `.equals`?

konradweiss (changes requested):
I would like some clarification on the KeySet and the use of isInstanceOf

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
oxisto: The doc says:
```
/**
     * Adds the specified element to the set.
     *
     * @return `true` if the element has been added, `false` if the element is already contained in the set.
     */
```

And since it is a `Set`, it will only contain each "equal" object exactly once.

konradweiss (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeResolver.kt
```diff
@@ -30,190 +30,37 @@ import de.fraunhofer.aisec.cpg.graph.Component
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration
 import de.fraunhofer.aisec.cpg.graph.types.*
-import de.fraunhofer.aisec.cpg.graph.types.HasSecondaryTypeEdge
 import de.fraunhofer.aisec.cpg.helpers.SubgraphWalker.IterativeGraphWalker
 
+/**
+ * The purpose of this [Pass] is to establish a relationship between [Type] nodes (more specifically
+ * [ObjectType]s) and their [RecordDeclaration].
+ */
```
KuechA, line 38: I'm wondering if we still need this pass or if we could just set/generate the `Type` to `RecordDeclaration` mapping when we parse the `RecordDeclaration`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeResolver.kt
```diff
@@ -30,190 +30,37 @@ import de.fraunhofer.aisec.cpg.graph.Component
...
     /**
      * Pass on the TypeSystem: Sets RecordDeclaration Relationship from ObjectType to
      * RecordDeclaration
      *
      * @param component
      */
     override fun accept(component: Component) {
-        removeDuplicateTypes()
         val walker = IterativeGraphWalker()
-        walker.registerOnNodeVisit(::ensureUniqueType)
```
KuechA, line 151: We still walk through every single node in the graph. Is this intended? Most of them are irrelevant as we only care about `RecordDeclaration`s in the handle function.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
oxisto: I tried that and its not possible because it only returns the value of the map which is then `True`, but I optimised it a little bit by only using `first` and not `filterIsInstance` and `first`.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeResolver.kt
oxisto: Not quite sure, I would explore this in a separate PR if this is possible.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeResolver.kt
oxisto: I will change it to an AST listener only listening for records.

PR #1311, opened 2023-09-01
Use FluentDSL in analysis module

KuechA:
We want to get rid of the de-facto required frontends java and cxx. The last blocking piece are the tests in the analysis module. This PR re-writes the tests to use the fluent DSL.


cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/MultiValueEvaluatorTest.kt
```diff
@@ -192,7 +184,8 @@ class MultiValueEvaluatorTest {
 
     @Test
     fun testLoop() {
-        val topLevel = Path.of("src", "test", "resources", "value_evaluation")
+        val tu = ValueEvaluationTests.getCfExample().components.first().translationUnits.first()
+        /*val topLevel = Path.of("src", "test", "resources", "value_evaluation")
```
oxisto, line 188: Remove

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/DoStatement.kt
```diff
@@ -50,6 +51,15 @@ class DoStatement : Statement() {
             .append("statement", statement)
             .toString()
 
+    override fun addArgument(expression: Expression) {
+        this.condition = expression
+    }
+
+    override fun replaceArgument(old: Expression, new: Expression): Boolean {
+        this.condition = new
```
oxisto, line 59: This should check whether this condition holds "old" or not and only replace if condition is "old" and return true/false accordingly.

oxisto:
We should also remove `defaultLanguages` from the `AnalysisTest` (and possibly other places)

KuechA:
> We should also remove `defaultLanguages` from the `AnalysisTest` (and possibly other places)

I completely removed the `defaultLanguages()` method.

PR #1315, opened 2023-09-06
Symbol resolver with EOG power

oxisto:
This PR leverages the power of the already established EOG to establish a relatively good order of resolving symbols. This may not be 100 % accurate yet, but gives us a good enough approximation for now.

In this PR, we also merged the `VariableUsageResolver` and `FunctionCallResolver` into a single `SymbolResolver`. The code is mostly copy/paste from both resolvers and still has a lot of potential for more refactoring in the future, because both did similar things.

The resolver will look for `EOGStartHolder` nodes (a new interface), to start the resolving cycle. Examples include function declarations, top-level declarations or catch-clauses of try-statements. After gathering all start nodes, it will follow the EOG edges and resolve variables and functions, depending on the type of node it encounters.

This allows us to resolve scenarios where a variable's type depends on the outcome of a function call and then a method call is called on it (which was not possible to resolve before), like the following:

```Java
var variable = method();
variable.method2();
```

It also slightly improves symbol resolving in the individual languages, primarily Go, by modelling more language specifics, such as built-in functions or type derivation logic.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/EOGBreaker.kt
```diff
@@ -0,0 +1,36 @@
...
+ *                   $$ |      $$$$$$$  |$$ |$$$$\
+ *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph
+
+/**
```
oxisto, line 28: I wonder if it would be better to iterate the AST, but order them (partially) in the EOG, maybe using the lattice? So that we really catch all nodes?

KuechA: I'm not completely sure what you mean with it but what we could do in the `SubgraphWalker` is to add AST-children if they are disconnected from the EOG. But this sounds to me like a solution to address

1) the case where we cannot precisely model the EOG (e.g. because it's evaluated at a later point in time or it depends on an uncertain trigger),
2) the case where there should be an EOG but we forgot to model it in the EOG Pass and
3) there case where a node is not in EOG but for some reason it holds relevant information but we never explicitly thought of it being relevant (What is this?! Should never happen because it would somewhat mean that we don't know what we're doing) ;)

Numbers 2 and 3 simply shouldn't happen and I'd prefer to see a test failing if that's the case

The lattice-based fixed-point iteration won't help us in this case but in fact, I should check that we do not miss visiting some parts there

oxisto: I added the resolution start holder interface also to the TryStatement, so that resolution starts with all catch clauses. That seems to work for now. This way we don't need an extra interface

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -406,7 +406,12 @@ open class Node : IVisitable<Node>, Persistable, LanguageProvider, ScopeProvider
                 code == other.code &&
                 comment == other.comment &&
                 location == other.location &&
-                file == other.file &&
+                // We need to exclude "file" here, because in C++ the same header node can be
+                // imported in two different files and in this case, the "file" property will be
```
konradweiss, line 410: Thanks for the Doc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/FunctionDeclaration.kt
```diff
@@ -111,27 +112,35 @@ open class FunctionDeclaration : ValueDeclaration(), DeclarationHolder {
...
-        val signature =
-            parameters
-                .stream()
-                .sorted(Comparator.comparingInt(ParameterDeclaration::argumentIndex))
-                .collect(Collectors.toList())
+    fun hasSignature(call: CallExpression): Boolean {
+        return hasSignature(call.signature, call.arguments)
+    }
+
+    // TODO: Documentation required. It's not completely clear what this method is supposed to do.
```
konradweiss, line 119: In this function we have several todos and we should solve them either here if we can figure out what is relevant for documentation, if not we have to postpone it but mark it as an issue

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/SubgraphWalker.kt
```diff
@@ -392,13 +395,22 @@ object SubgraphWalker {
...
+        }
+
         /**
          * Wraps [IterativeGraphWalker] to handle declaration scopes.
          *
          * @param root The node where AST descent is started
          */
         fun iterate(root: Node) {
             walker = IterativeGraphWalker()
+            walker!!.strategy = this.strategy
```
konradweiss, line 413: Can we get rid of `!!` here?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
```diff
@@ -191,15 +189,19 @@ open class EvaluationOrderGraphPass(ctx: TranslationContext) : TranslationUnitPa
             eogNodes
                 .filter { node ->
                     node is FunctionDeclaration ||
+                        node is VariableDeclaration ||
```
konradweiss, line 192: Is this to not filter out eog s built for default values?

konradweiss (changes requested):
Some minor changes, and an idea for how to document and change a Todo in code. I did also ask some questions to help me understand some things that are unclear to me.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -774,24 +765,35 @@ class ScopeManager : ScopeProvider {
...
+                    scope.valueDeclarations.filterIsInstance(klass).filter(predicate)
+                )
             }
 
             if (scope is StructureDeclarationScope) {
-                var list = scope.structureDeclarations.filterIsInstance<T>().filter(predicate)
+                var list = scope.structureDeclarations.filterIsInstance(klass).filter(predicate)
 
                 // this was taken over from the old resolveStructureDeclaration.
                 // TODO(oxisto): why is this only when the list is empty?
```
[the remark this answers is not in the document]
oxisto: We can't use `list.addAll` because what we return from `filter(predicate)` is a `List`, not a `MutableList`, in any case it would not make a difference because `list` is empty at line 796 line. 

BUT in both cases the contents of `list` are already added to the `declarations` list using `addAll` and only `declarations` is returned, so it is not only returning the last matched template.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
oxisto: > Yeah, this should prevent removing EOG edges for nodes which are not reachable by the EOG from the starting node for different reasons. Similar to the filters which already existed. I'm wondering though if we can use the `ResolutionStartHolder` interface instead of listing all these classes?

That might work. I'll try

oxisto: I changed it to 
```kotlin
var validStarts =
            eogNodes.filter { it is ResolutionStartHolder || it is VariableDeclaration }.toSet()
```

The tests in cog-core work, but I have no idea if this is "correct"

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/Type.kt
```diff
@@ -221,3 +222,14 @@ abstract class Type : Node {
         }
     }
 }
+
+/** A shortcut to return [ObjectType.recordDeclaration], if this is a [ObjectType]. */
+var Type.recordDeclaration: RecordDeclaration?
```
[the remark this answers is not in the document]
oxisto: Because it's a shortcut to check if this type is an object type and if yes, to retrieve the record declaration. Not all types have a record declaration, so I would prefer this as an extension rather than a property.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -656,3 +935,7 @@ open class CallResolver(ctx: TranslationContext) : SymbolResolverPass(ctx) {
         }
     }
 }
+
+interface ResolutionStartHolder {
```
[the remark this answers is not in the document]
oxisto: Done. Please have a look at it; currently, I only describe the usage in the symbol resolver. But we might also extend this to other things, e.g., as an entry point to the c-sensitive DFG pass. Should we rename it?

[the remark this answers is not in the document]
KuechA: Looks good. Yes, the ControlFlowSensitiveDFGPass is another application of this. Maybe EOGStart or similar?

KuechA (approved):

oxisto:
We also need to update https://fraunhofer-aisec.github.io/cpg/CPG/specs/eog/#variabledeclaration because the EOG of variable declarations changed.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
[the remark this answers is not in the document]
konradweiss: Ah ok I missed the `declarations.addAll(list)`. Can we update the Todo comment then? Or is there still information missing on why this is done when the list is empty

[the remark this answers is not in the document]
oxisto: It is basically still copy paste from very old legacy code and I have no idea why it is done this way. Therefore I would keep this here and hopefully somebody somewhere will overhaul the template code.

konradweiss (approved):

PR #1332, opened 2023-10-11
Several improvements and fixes to the DFG

KuechA:
This PR aims to improve the DFG which currently fails to consider readwrite and write accesses in many cases. This PR fixes the `access` property of `Reference` and uses this information more frequently in the `ControlFlowSensitiveDFGPass`. It also adds some properties for the DFG edges


oxisto:
Will be split up into individual PRs.

PR #1333, opened 2023-10-16
Support for `<<`, `>>` and `|`, `&`, `^` in `ValueEvaluator` and `BinaryOperation`

oxisto:


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/ValueEvaluator.kt
```diff
@@ -163,6 +163,9 @@ open class ValueEvaluator(
             "/=" -> handleDiv(lhsValue, rhsValue, expr)
             "*",
             "*=" -> handleTimes(lhsValue, rhsValue, expr)
+            "<<" -> handleShiftLeft(lhsValue, rhsValue, expr)
+            ">>" -> handleShiftRight(lhsValue, rhsValue, expr)
+            "|" -> handleXor(lhsValue, rhsValue, expr)
```
KuechA, line 168: This is quite tricky. In some languages (e.g. C, Java, Kotlin), the `|` operator is a bitwise or or an or without terminating early. And instead the `^` or `xor`  is an xor operator... I'm not sure how to address this problem. Maybe, we have to model operators by the language?

oxisto: Ah, you are correct. I was actually using it incorrectly. Its the same in Go as in C

&    bitwise AND            integers
|    bitwise OR             integers
^    bitwise XOR            integers

PR #1334, opened 2023-10-16
Caching symbols on symbol resolution

oxisto:
This PR caches symbol resolutions. This can can significantly speed up the resolution process.


oxisto:
> I like this. Maybe provide some tests to make sure corner cases like variable shadowing work as expected?

Good idea. One caveat though: variable shadowing is not really supported even without caching :)

konradweiss (approved):
LGTM

PR #1341, opened 2023-10-19
Removed SonarQube warnings

oxisto:
One might need to copy the example grade.properties manually, since it seems the `sonar.gradle.skipCompile` is a system property, which needs to be set in `gradle.properties`.


maximiliankaul (approved):
No more sonar warnings on my machine :)

PR #1357, opened 2023-11-20
Improved `guessCastExpressions` in C++ frontend

### LINKED ISSUE #1347: Incorrect parsing of nested casts and unary operators in the CXX language frontend
# Unexpected behavior

A BinaryOperation `&` with lhs CastExpression(int64_t) and rhs x is generated instead of a CastExpression of the UnaryOperator `&` with input `x`

# Code of interest

```
(int64_t)&x
```

oxisto:
There were more ambiguities when guesssing cast expressions, so I added them.
Fixes #1347


KuechA (approved):
Extremely hacky but since the current output has been wrong in some cases, it seems to be necessary.

PR #1365, opened 2023-11-23
Experimental support to run passes in parallel

oxisto:


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -45,12 +45,13 @@ import kotlin.contracts.contract
 @OptIn(ExperimentalContracts::class)
 @DependsOn(EvaluationOrderGraphPass::class)
 @DependsOn(DFGPass::class)
-open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : TranslationUnitPass(ctx) {
+open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : FunctionPass(ctx) {
```
oxisto, line 48: @KuechA should we do it for each resolution start holder instead?

KuechA: In the long term it should definitely work for each `ResolutionStartHolder` but we can filter for `FunctionDeclarations` for now.

KuechA (approved):

PR #1369, opened 2023-11-23
Use `IdentitySet` and `IdentityHashMap` in EOG iteration for CF-DFG to improve performance

KuechA:


oxisto:
Could this also be used in the implementation of `PowersetLattice`? I noticed when stepping through a large program that it seems that a lot of time is spent in `duplicate`, which seems to use a regular `Set`.

oxisto:
> Could this also be used in the implementation of `PowersetLattice`? I noticed when stepping through a large program that it seems that a lot of time is spent in `duplicate`, which seems to use a regular `Set`.

I have implemented this very quickly and it seems that this provides another performance boost. In my example, the CF-DFG went from 26s to 17s.

oxisto (approved):

PR #1371, opened 2023-11-24
Add fields to `Reference`s and `VariableDeclaration`s

KuechA:
Adds the following fields:
* `isGlobal` (getter) to determine if a variable is a local or a global variable. This can make a difference ;)
* `aliases` a list collecting aliases of a reference or a variable (more precisely `ValueDeclaration`)

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/ValueDeclaration.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/ValueDeclaration.kt
@@ -61,7 +61,7 @@ abstract class ValueDeclaration : Declaration(), HasType, HasAliases {
             }
         }
 
-    override var aliases = mutableSetOf<Declaration>()
+    override var aliases = mutableSetOf<HasAliases>()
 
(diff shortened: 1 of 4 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/ValueDeclaration.kt
```diff
@@ -61,6 +61,8 @@ abstract class ValueDeclaration : Declaration(), HasType {
             }
         }
 
+    var aliases = mutableSetOf<Node>()
```
oxisto, line 64: Can we somehow restrict this into set of specific nodes? Does it always contain declarations or references? Or both?

Does it make sense to have an `HasAliases` that both references and declarations implement.

KuechA: Yes, makes sense

KuechA: Actually I think it always contains declarations but it could be potentially also something different

oxisto (approved):

Issue #1373, opened 2023-11-25
Label from abstract classes are not persisted to Neo4j


konradweiss:
Abstract classes are not recognized as sources for labels anymore, while labels are still added to a node if the super class can be instantiated. Bellow we can see that a `ConstructorDeclaration` node has the labels `MethodDeclaration` and `FunctionDeclaration` but is missing the important `Declaration` label. This prevents queries to identify by the abstract types `Statement`, `Expression`, `Declaration`, `Type`, etc.

![image](https://github.com/Fraunhofer-AISEC/cpg/assets/12885000/a3b8d1d7-1e9b-46ff-bf39-5c7a808e70b3)

oxisto:
Closed by #1374

PR #1374, opened 2023-11-25
Add NodeEntity to abstract node classes to get them persisted to neo4j

konradweiss:
Solves #1373 

By adding the @NodeEntity to abstract classes of nodes, neo4j again persists labels like `Type`, `Expression`, `Declaration`. This is only necessary for abstract classes as super classes that are not abstract get the label persisted.


oxisto (approved):
I was wondering why `Node` does not have the `@NodeEntity`. Can you check, whether this also solve the issue?

konradweiss:
> I was wondering why `Node` does not have the `@NodeEntity`. Can you check, whether this also solve the issue?

Checked it, unfortunately it did not solve it. `Node` is an open class and apparently adding the annotation does not propagate to the abstract classes. The `Node` label, however, was also present before adding the annotation. so the behavior is as what I expected after seeing the `FunctionDeclaration` label at a `ConstructorDeclaration`. The OGM assumes that a non-abstract Class represents a label and `NodeEntity` is only needed in the abstract class case.

PR #1380, opened 2023-11-27
Cleanup NodeBuilder

maximiliankaul:
NodeBuilder uses `rawNode` to apply metadata. Optional parameters `location` and `code` are not in the rest of the codebase used and removed to cleanup the NodeBuilder code.


oxisto:
The remaining failing tests are in the python frontend. Since we are re-writing this anyway, we are not fixing this in the old frontend, but are blocked by https://github.com/Fraunhofer-AISEC/cpg/pull/1282, which we will merge soon.

PR #1382, opened 2023-11-28
Fixes to DFG

### LINKED ISSUE #2485: Missing support for `reference-to-pointer` parameters in CPP
Currently we aren't handling [`reference-to-pointer`](https://www.geeksforgeeks.org/cpp/passing-reference-to-a-pointer-in-c/) parameters (T*&) in C++.  The following example illustrates the issue: 

```
void badSource(char*& data)
{
    data = new char[100];
    memset(data, 'A', 100 - 1);
    data[99] = '\0';

    delete[] data;
}

void bad()
{
    char* data = NULL;
    badSource(data);
    printLine(data);  // Use-after-free
}
```

oxisto:
# New Nodes
+ MemoryAddress: (nearly) every node has a memoryAddress now, for which we track the current value stored there. For Declarations, we create a new memoryAddress, the References use the same memoryAddress as their refersTo
+ UnknownMemoryValue: The value stored might be unknown, in which case this node comes in
+ ParameterMemoryValue: The value that comes in via a parameter. 
+ PointerReference (&a): Has an input (in this case Reference „a“), and indicates that we actually use the Reference (as in C, so the memory address, not as in CPG-Reference) of the input
+ PointerDereference (*p): Has an input (in this case Reference „p“), and indicates that we actually use the Dereference, as in the value we find when using the input as address

# FunctionSummaries
  What does the function do? What’s the new value of arguments after the function is finished? 
Mostly for internal processing. Allows us to process each function only once and then process the affects of function calls

# New (DFG)-Edges
+ currentDeref(Deref)Value: Mostly relevant for Expression/Pointerhandling. I.e, if we have „int i=0; int* p=&i“, then the currentDerefValue of „p“ points to „0“.
+ shortFS: Gives the DFG-Summaries-Edge that does not enter other functions, but only indicates that i.e. an argument is handed to a function and that the function does something with it. Handy if we want to travers larger DFGs and don’t want to enter every single function
+ memoryValue: The current value stored for a Node, mostly relevant for Expressions. I.e. for „i=0; printf(i)“, the 2nd Reference „i“ should have a memoryValue-Edge pointing to the Literal 0.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/MermaidPrinter.kt
```diff
@@ -69,6 +98,158 @@ fun AstNode.printAST(
...
+ * current [Node] as Markdown, with an embedded [Mermaid](https://mermaid.js.org) graph. The output
+ * can either be pasted into a Markdown document (and then rendered) or directly pasted into GitHub
+ * issues, discussions or pull requests (see
+ * https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/).
+ *
+ * @param strategies The strategies to use when iterating the graph. See [Strategy] for
+ *   implementations.
+ * @return The Mermaid graph as a string encapsulated in triple-backticks.
+ */
+fun <EdgeType : Edge<out Node>> Node.printGraphNew(
```
KuechA, line 168: I think this never really worked.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/crypto/hashing/Hashing.kt
```diff
@@ -0,0 +1,77 @@
...
+ */
+package de.fraunhofer.aisec.cpg.graph.concepts.crypto.hashing
+
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.concepts.Concept
+import de.fraunhofer.aisec.cpg.graph.concepts.Operation
+import de.fraunhofer.aisec.cpg.graph.concepts.crypto.IsCryptography
+import java.util.Objects
+
+enum class HashingAlgorithm(val algorithmName: String) {
```
KuechA, line 34: Probably not required any more

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/crypto/IsCryptography.kt
```diff
@@ -0,0 +1,28 @@
...
+ *                   $$ |      $$$$$$$  |$$ |$$$$\
+ *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.concepts.crypto
+
+interface IsCryptography
```
KuechA, line 28: probably not required anymore

KuechA (approved):
Let's finally get rid of this

PR #1385, opened 2023-11-30
Implement `ScopeProvider` in `Pass`

oxisto:
This PR implements the `ScopeProvider` interface in the `Pass` class. This should now enable passes to use the information in the stored `scopeManager` when creating new nodes with the node builder.


maximiliankaul (approved):
:tada:

PR #1388, opened 2023-12-01
Restoring Comment Matcher Functionality

maximiliankaul:
CommentMatcher functionality was not implemented in the recent python rewrite. This pr restores the functionality.

Reference to the old code: https://github.com/Fraunhofer-AISEC/cpg/blob/c81ef60152baab938bb594d11c70a9e1f3d3e0d2/cpg-language-python/src/main/python/cpg.py#L61


maximiliankaul:
@konradweiss this restores the old logic. However the test fails. I think it's somewhere in the CommentMatcher logic, as the debugger says there are only 7 comments created while I see 10 calls to the CommentMatcher.

oxisto (approved):

PR #1400, opened 2023-12-24
Incorporating Gradle tasks for automatically installing JEP

akuleshov7:
### What's Done:
1) Introduced the `installJep` task to facilitate the automatic installation of JEP from [https://github.com/icemachined/jep-distro/](https://github.com/icemachined/jep-distro/).

2) Centralized and encapsulated the logic for JEP search and installation: all the relevant logic has been relocated to the Gradle task, separating it from the main code. The main code will now simply utilize the JEP stored in the `build/jep` directory.

### DIFF
--- a/cpg-language-python/build.gradle.kts
+++ b/cpg-language-python/build.gradle.kts
@@ -1,3 +1,8 @@
+import java.nio.file.Path
+import kotlin.io.path.*
+import kotlin.io.path.Path
+import kotlin.io.path.exists
+
 /*
  * Copyright (c) 2021, Fraunhofer AISEC. All rights reserved.
  *
@@ -23,6 +28,15 @@
  *                    \______/ \__|       \______/
  *
  */
+
+val jepDistroVersion = "4.1.1"
+// Python version can be set with `-Ppython` property: `./gradlew cpg-language-python:build -Ppython=3.12`, default is 3.10
+val pythonVersion = if (project.hasProperty("python")) project.property("python") else "3.10"
+val jepCpgDir: Path = Path(projectDir.path, "build", "jep")
+val regularJepPathInVirtualEnv: Path = Path(
+    "lib", "python${pythonVersion}", "site-packages", "jep"
+)
+
 plugins {
     id("cpg.frontend-conventions")
 }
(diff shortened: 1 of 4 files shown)


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/JepSingleton.kt
```diff
@@ -25,96 +25,50 @@
...
-                    "lib",
-                    "python${version}",
-                    "site-packages",
-                    "jep",
-                    "libjep.dll"
-                )
-            )
-        }
-        // try system-wide paths, too
-        // TODO: is this still needed?
```
akuleshov7, line 100: No, it is not...

akuleshov7:
Please check @oxisto

0x6675636b796f75676974687562 (approved):

oxisto:
cc @maximiliankaul

oxisto:
Thanks for your contribution! (And signing the CLA ;) Looks like an interesting idea, I will have some time between Christmas and New Years to look at this as well.

akuleshov7:
> Thanks for your contribution! (And signing the CLA ;) Looks like an interesting idea, I will have some time between Christmas and New Years to look at this as well.

Thanks! Happy holidays! Please also approve a workflow (I added 1 commit, trying to fix build with JEP)

cpg-language-python/build.gradle.kts
```diff
@@ -39,10 +53,112 @@ publishing {
...
+    } else {
+        // As mentioned above, ARM machines are not supported, so you need to provide your own JEP distribution
+        // in that case with CPG_JEP_LIBRARY, or other installations like homebrew, virtual env., etc.
+        if (!System.getProperty("os.arch").contains("x86")) {
+            throw IllegalStateException(
+                """
+                    | Your system architecture, identified as <${System.getProperty("os.arch")}>, is not supported in icemachined/jep-distro. 
+                    | Consequently, you are required to install it manually. Here are the potential solutions:
+                    |  1. Utilize Homebrew/pip package manager to facilitate the JEP installation process; 
+                    |  2. Create a virtual environment with the specified environment variable CPG_PYTHON_VIRTUALENV set to /(user.home)/.virtualenv/CPG_PYTHON_VIRTUALENV;
```
oxisto, line 117: sometimes its `.virtualenv` and sometimes ``.virtualenvs`

oxisto (commented):
In general, I like that idea of using the jep-distro files, but I wonder if moving the logic from runtime to compile time is a good idea for non-distro-supported systems.

maximiliankaul:
Hi @akuleshov7 ,
thanks so much for our PR. I'm a bit undecided whether I like it or not:
- The entire JEP installation / configuration process has always been a pain for us and we applaud any effort to improve this situation :+1:
- I'm very hesitant to include a dependency to `icemachined` as this does not seem to be related with the JEP project / it's unclear to me how well this will be maintained.
- Your PR does not work on my machine (Linux). Setting the `CPG_PYTHON_VIRTUALENV=cpg` env variable still results in the `Your system architecture, identified as <amd64>, is not supported in icemachined/jep-distro. ` error message.
- Can you please comment on the following usecases which are important to us:
  - We are using the CPG in other projects as a dependency. How does this work with your PR?
  - Your PR / icemachined only supports x86, correct? We do not want do drop ARM support.
  - We have to fix a Python version in the build process (i.e. on Github) and this version is required on any client using the Github build, correct? We would like to avoid a scenario like this.

We will discuss testing this with some other projects and see if anything breaks. Any improvement in the JEP install logic is very welcome to us.

akuleshov7:
> * I'm very hesitant to include a dependency to `icemachined` as this does not seem to be related with the JEP project / it's unclear to me how well this will be maintained.

@icemachined, Dima, how well do you maintain JEP-distro? 😄

icemachined (changes requested):

icemachined:
> > * I'm very hesitant to include a dependency to `icemachined` as this does not seem to be related with the JEP project / it's unclear to me how well this will be maintained.
> 
> @icemachined, Dima, how well do you maintain JEP-distro? 😄

I'm not modifying jep I just perform jep build on github and publish it on maven, you can see the code here https://github.com/icemachined/jep-distro. If there will be any issues I can fix it or accept your fix.

maximiliankaul:
>       * Your PR / icemachined only supports x86, correct? We do not want do drop ARM support.
>       * We have to fix a Python version in the build process (i.e. on Github) and this version is required on any client using the Github build, correct? We would like to avoid a scenario like this.

Is there any update on these two points? @oxisto, if I remember correctly, we can live without the ARM support (although we would like to keep it as some of our team work on ARM machines), but the second point is a deal breaker for us, right?

icemachined:
> >   * Your PR / icemachined only supports x86, correct? We do not want do drop ARM support.
> >   * We have to fix a Python version in the build process (i.e. on Github) and this version is required on any client using the Github build, correct? We would like to avoid a scenario like this.
> > ```
> 
> Is there any update on these two points? @oxisto, if I remember correctly, we can live without the ARM support (although we would like to keep it as some of our team work on ARM machines), but the second point is a deal breaker for us, right?

Hi @maximiliankaul , jep-distro currently support ARM, latest release have ARM libraries tested on M2. so if you need we can update pr with ARM support

maximiliankaul:
> > > * Your PR / icemachined only supports x86, correct? We do not want do drop ARM support.
> > > * We have to fix a Python version in the build process (i.e. on Github) and this version is required on any client using the Github build, correct? We would like to avoid a scenario like this.
> > 
> > 
> > Is there any update on these two points? @oxisto, if I remember correctly, we can live without the ARM support (although we would like to keep it as some of our team work on ARM machines), but the second point is a deal breaker for us, right?
> 
> Hi @maximiliankaul , [jep-distro](https://github.com/icemachined/jep-distro) currently support ARM, latest release have ARM libraries tested on M2. so if you need we can update pr with ARM support

That's great to hear :) When we first checked this PR, there was no support / we missed this.
However, as we mentioned above, the deal breaker is fixing the Python version at compile time. Please don't get me wrong: we don't like the current solution and welcome any improvements (and this PR is an improvement in many ways), but we don't want to lose the ability to build the CPG once (e.g. here on Github) and distribute it to machines running different versions of Python. This is not achievable with this PR right now, is it?

icemachined:
> That's great to hear :) When we first checked this PR, there was no support / we missed this. However, as we mentioned above, the deal breaker is fixing the Python version at compile time. Please don't get me wrong: we don't like the current solution and welcome any improvements (and this PR is an improvement in many ways), but we don't want to lose the ability to build the CPG once (e.g. here on Github) and distribute it to machines running different versions of Python. This is not achievable with this PR right now, is it?

Yes you are right seems like this is no achievable in this pr. But it is achievable in general because jep-distro contains builds for all architectures and versions of python starting from 3.6 till 3.12. You could include all required versions in distro and unpack one of them in runtime to tmp depending on current version of python.

konradweiss:
Hello everyone,
 
we are cleaning up old PRs and have decided to close this one as it does not really fit our target use case, as discussed above. 
Shipping multiple versions of jep and unpacking them at runtime does have some drawbacks to the current approach. As stated above, we are not happy with the current mechanism either and are open to improvements. This PR, however, is too far away from what we want to achieve.

Thank you for contributing!

Best regards,
The CPG Team

PR #1403, opened 2024-01-02
Added `registerReplacement` to `ScopedWalker`

oxisto:
In some frontends (e.g., the Go frontend) we are replacing certain nodes in an extra pass, while iterating with the `ScopedWalker`. One such usecase is to replace calls with casts. Previously, when doing so, chained replacement of calls were a problem, since the original replaced value was stil registered in the walker and was used to determine further AST children to "walk" on, instead of the new node.

This adds a function to register such replacements with the walker.


konradweiss (approved):
LGTM.

PR #1407, opened 2024-01-11
Several improvements and cleanups to the `Inference` system

oxisto:
This PR does some cleanup and improvements to the inference systems. It is made more configurable, with more options added to `InferenceConfiguration`.

For example, currently, the `Inference` class only infers fields. However, some languages (such as C/C++) have global variables, which we can also infer.

Note, that we intentionally *NOT* infer local variables, since we assume that we have incomplete code, that is probably complete within a function, but not complete in the sense that translation unit files or headers are missing.

This also aligns the DFG of inferred functions more to non-inferred ones and adds inferred methods to the AST.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
@@ -31,6 +31,7 @@ import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.declarations.FunctionDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
+import de.fraunhofer.aisec.cpg.graph.scopes.GlobalScope
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.CallExpression
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.MemberExpression
 import de.fraunhofer.aisec.cpg.graph.types.Type
@@ -43,7 +44,7 @@ import de.fraunhofer.aisec.cpg.passes.SymbolResolver
  *
  * Currently, this interface has no methods. However, in the future, this could be used to execute
  * language/frontend-specific code for the particular trait. This could help to fine-tune the
- * [CallResolver] for specific languages.
+ * [SymbolResolver] for specific languages.
  */
 interface LanguageTrait
 
@@ -237,6 +238,11 @@ interface HasAnonymousIdentifier {
         get() = "_"
 }
 
+/**
+ * A language trait, that specifies that this language has global variables directly in the
+ * [GlobalScope], i.e,. not within a namespace, but directly contained in a
+ * [TranslationUnitDeclaration].
+ */
 interface HasGlobalVariables {
     val globalVariableScopeClass: Class<out Node>
 }
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
@@ -27,6 +27,7 @@ package de.fraunhofer.aisec.cpg.frontends
 
 import de.fraunhofer.aisec.cpg.ScopeManager
 import de.fraunhofer.aisec.cpg.TranslationContext
+import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.declarations.FunctionDeclaration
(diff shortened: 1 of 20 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -341,6 +342,32 @@ class Inference(val start: Node, override val ctx: TranslationContext) :
...
+                "Inferred a new variable declaration {} with type {}",
+                inferred.name,
+                inferred.type
+            )
+
+            // In any case, we will observe the type of our reference and update our new variable
+            // declaration accordingly.
+            ref.typeObservers += TypeInferenceObserver(inferred)
+
+            // Add it the scope
```
KuechA, line 364:
```suggestion
            // Add it to the scope
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
```diff
@@ -398,11 +404,13 @@ class DFGPass(ctx: TranslationContext) : ComponentPass(ctx) {
...
-                } else {
-                    Util.attachCallParameters(it, call.arguments)
-                    call.addPrevDFG(it)
+                Util.attachCallParameters(it, call.arguments)
+                if (it is MethodDeclaration && call is MemberCallExpression && !call.isStatic) {
+                    it.receiver?.let { receiver -> call.base?.addNextDFG(receiver) }
+                }
+                call.addPrevDFG(it)
+                if (it.isInferred && it.prevDFG.isEmpty()) {
+                    handleFunctionDeclaration(it)
```
oxisto, line 413: We should add a comment here, why we are doing it this way, i.e., that inferred nodes are most likely not in the AST (yet) and that handleFunctionDeclaration only iterates on AST functions.

KuechA: I added a comment for it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -310,36 +350,90 @@ class Inference(val start: Node, override val ctx: TranslationContext) :
...
+            type.recordDeclaration = declaration
 
-        // update the type
-        type.recordDeclaration = declaration
+            scopeManager.addDeclaration(declaration)
+            declaration
+        }
+    }
+
+    fun inferVariableDeclaration(hint: Reference): VariableDeclaration? {
```
konradweiss, line 399: Small comment to specify which scope it will use

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXInferenceTest.kt
```diff
@@ -0,0 +1,45 @@
...
+package de.fraunhofer.aisec.cpg.frontends.cxx
+
+import de.fraunhofer.aisec.cpg.TestUtils
+import java.io.File
+import kotlin.test.Test
+import kotlin.test.assertNotNull
+
+class CXXInferenceTest {
+    @Test
+    fun testGlobals() {
```
konradweiss, line 35: Doesn't this just check if translation was successful? the name of the test gives me the impression the presence of some global variable should be checked

konradweiss (changes requested):
Pls add some small missing documentation and clarify if testGlobals is correct. I also noticed that the check if inference is turned on is fragmented over the subfunctions. while it would be cool to check that asap when the first inference function is called, the checks which inference, e.g. infering records etc. is turned on needs to happen further down so I don't know if we can make it better or have to accept that checking if inference is turned on has to be done in many different places.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
```diff
@@ -235,3 +236,7 @@ interface HasAnonymousIdentifier {
     val anonymousIdentifier: String
         get() = "_"
 }
+
+interface HasGlobalVariables {
```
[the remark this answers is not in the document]
oxisto: Ah this is a left-over, I will move the global trait things in a separate PR

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXInferenceTest.kt
oxisto: Yes, its missing an assert 🤣

oxisto:
@konradweiss Could you have another look, if the changes are ok please? thanks!

konradweiss (approved):

PR #1411, opened 2024-01-14
Code and location from children

konradweiss:
This PR allows the setting of a node's code and location region as the code and location of its children.
Sometimes, when we translate a parent node in the language-specific  AST with its children into the CPG AST, we have to set a specific intermediate Node between that has no language-specific AST that can give it a proper code and location.

This PR provides the function `codeAndLocationFromChildren(frontend: LanguageFrontend<S, *>, parentNode: S)` that is purely optional and can be used in language frontends to cover the use-case mentioned above and set the region of the node to the start and end region of its children. The code is then extracted from the language-specific AST parent using the newly determined location region.

The function modifies the base node, meaning it should be used after construction:
`newBlock(...).codecodeAndLocationFromChildren()`
or before returning the node if the children are processed and returned after construction:
`return block..codecodeAndLocationFromChildren()`

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
@@ -272,6 +272,18 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
     return this
 }
 
+/**
+ * This function allows the setting of a node's code and location region as the code and location of its children.
+ * Sometimes, when we translate a parent node in the language-specific AST with its children into the CPG AST,
+ * we have to set a specific intermediate Node between, that has no language-specific AST that can give it a proper
+ * code and location.
+ *
+ * While the location of the node is determined by the start and end of the child locations, the code is extracted from
+ * the parent node to catch separators and auxiliary syntactic elements that are between the child nodes.
+ *
+ * @param frontend Used to invoke language specific code and location generation
+ * @param parentNode Used to extract the code for this node
+ */
 fun <T : Node, S> T.codeAndLocationFromChildren(
     frontend: LanguageFrontend<S, *>,
     parentNode: S
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
@@ -272,7 +272,10 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
     return this
 }
 
-fun <T : Node> T.codeAndLocationFromChildren(): T {
+fun <T : Node, S> T.codeAndLocationFromChildren(
+    frontend: LanguageFrontend<S, *>,
+    parentNode: S
+): T {
     var nodesWithLocation =
         this.astChildren
             .filter { it.location?.region != null && it.location?.region != Region() }
@@ -297,14 +300,25 @@ fun <T : Node> T.codeAndLocationFromChildren(): T {
         )
 
     if (sortedNodes.isNotEmpty()) {
-        this.location?.region =
-            Region(
-                startLine = sortedNodes.first().location?.region?.startLine ?: -1,
+        // All regions are present and not default at this point
+        val startLine = sortedNodes.first().location?.region?.startLine ?: -1
+        val endLine = sortedNodes.last().location?.region?.endLine ?: -1
+        val newRegion = Region(
+                startLine = startLine,
                 startColumn = sortedNodes.first().location?.region?.startColumn ?: -1,
-                endLine = sortedNodes.last().location?.region?.endLine ?: -1,
+                endLine = endLine,
                 endColumn = sortedNodes.last().location?.region?.endColumn ?: -1,
             )
+        this.location?.region = newRegion
+
+        val parentCode = frontend.codeOf(parentNode)
+        val parentRegion = frontend.locationOf(parentNode)?.region
+        if(parentCode != null && parentRegion != null){
+            this.code = frontend.getCodeOfSubregion(parentCode, parentRegion, newRegion)
+        }
+
     }
+
 
     return this
 }
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
@@ -303,7 +303,8 @@ fun <T : Node, S> T.codeAndLocationFromChildren(
         // All regions are present and not default at this point
         val startLine = sortedNodes.first().location?.region?.startLine ?: -1
         val endLine = sortedNodes.last().location?.region?.endLine ?: -1
-        val newRegion = Region(
+        val newRegion =
+            Region(
                 startLine = startLine,
                 startColumn = sortedNodes.first().location?.region?.startColumn ?: -1,
                 endLine = endLine,
@@ -313,12 +314,10 @@ fun <T : Node, S> T.codeAndLocationFromChildren(
 
         val parentCode = frontend.codeOf(parentNode)
         val parentRegion = frontend.locationOf(parentNode)?.region
-        if(parentCode != null && parentRegion != null){
+        if (parentCode != null && parentRegion != null) {
             this.code = frontend.getCodeOfSubregion(parentCode, parentRegion, newRegion)
         }
-
     }
-
 
     return this
 }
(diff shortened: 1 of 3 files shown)


konradweiss:
The quality gate fails because I cannot test it without using the function in a frontend, which I will add and then test in #1388. And unfortunately cannot add with the Fluent Node DSL as I need access to rawNodes and a valid frontend unless I heavily restructure the code just to make it testable and jump through some more hoops.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -269,3 +270,53 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
 
     return this
 }
+
+fun <T : Node, S> T.codeAndLocationFromChildren(
+    frontend: LanguageFrontend<S, *>,
+    parentNode: S
+): T {
+    var nodesWithLocation =
```
KuechA, line 278: Why don't you use `var nodesWithLocation = this.allChildren<Node>{it.location != null && it.location?.region != Region()}` instead of making this worklist and while loop?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -269,3 +270,53 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
 
     return this
 }
+
+fun <T : Node, S> T.codeAndLocationFromChildren(
```
oxisto, line 274: I think this method needs a good documentation here, because it is not self-explanatory that I need to specify the parent node (in raw AST form) to retrieve the code of its children.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -269,3 +270,53 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
...
+    frontend: LanguageFrontend<S, *>,
```
oxisto, line 275: It's a bit cumbersome to specify a frontend here, because we usually already have the frontend as part of the meta-data provider. We could use a context receiver for a `CodeAndLocationProvider` and we should change the `CodeAndLocationProvider` to implement `codeOf` and `locationOf` instead of the weird `setCodeAndLocation`. But that is probably a bigger change. If you want I can try to implement this here or I can try to do it after we merge this one?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
konradweiss: I found a comparison function of single elements with comparator function that allowed me to implement the lexicographic order. So I managed to make it more efficient:
 - no sorting
 - comparing every node only once

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
konradweiss: I would prefer to keep this PR small, as I need it for another. Changing it later sound interesting but for now I would update the Documentation of the function and then merge.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
konradweiss: On it

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -269,3 +270,81 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
...
+    var last: Node? = null
+
+    // Search through all children to find the first and last node based on region startLine and
+    // startColumn
+    var worklist: MutableList<Node> = this.astChildren.toMutableList()
+    while (worklist.isNotEmpty()) {
+        var current = worklist.removeFirst()
+        if (current.location?.region == null || current.location?.region == Region()) {
+            // If the node has no location we use the same search on his children again
+            worklist.addAll(current.astChildren)
```
oxisto, line 301: Shoudn't you add the ast children in any case to traverse all the children? This no only traverses if the children's node does not have a location?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -269,3 +270,81 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
...
+fun <T : Node, S> T.codeAndLocationFromChildren(
+    frontend: LanguageFrontend<S, *>,
+    parentNode: S
+): T {
+    var first: Node? = null
+    var last: Node? = null
+
+    // Search through all children to find the first and last node based on region startLine and
+    // startColumn
+    var worklist: MutableList<Node> = this.astChildren.toMutableList()
```
oxisto, line 296: val

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -269,3 +270,81 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
...
+    while (worklist.isNotEmpty()) {
+        var current = worklist.removeFirst()
```
oxisto, line 298: val

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -269,3 +270,81 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
...
+        if (current.location?.region == null || current.location?.region == Region()) {
+            // If the node has no location we use the same search on his children again
+            worklist.addAll(current.astChildren)
+        } else {
+            // Compare nodes by line and column in lexicographic order, i.e. column is compared if
+            // lines are equal
+            if (first == null || last == null) {
```
oxisto, line 305: last == null always false

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -269,3 +270,81 @@ fun <T : Node, S> T.codeAndLocationFrom(frontend: LanguageFrontend<S, *>, rawNod
...
+                first = current
+                last = current
+            }
+            first =
+                minOf(
+                    first,
+                    current,
+                    compareBy(
+                        { it?.location?.region?.startLine },
```
oxisto, line 314: ? unnecessary. also the other ?s

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
konradweiss: No, because a child node should not have a location outside of the parent location, it is intentional to add the children only if the parent has no location.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
konradweiss: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
konradweiss: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
konradweiss: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
konradweiss: I removed the one after `it` but the others are necessary due to the compiler not understanding that I already ensured their non-nullness

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
oxisto: Ah, that makes sense.

oxisto (approved):

PR #1416, opened 2024-01-18
fix import of language frontends in kotlin script

peckto:


KuechA (approved):
Ruby is also missing but apart from that it looks ok

peckto:
> Ruby is also missing but apart from that it looks ok

Thanks, fixed

PR #1430, opened 2024-02-09
Enable using function summaries to get more precise and custom DFG edges for inferred functions.

KuechA:
Currently, we infer DFG edges for function/method declarations by applying a heuristic. However, this is not always correct and it is desirable to have an option to specify more precise function summaries. Here, we introduce a JSON file format to do this and integrate it into the `DFGPass`.


cpg-core/src/test/resources/function-dfg.json
```diff
@@ -0,0 +1,56 @@
+[
```
oxisto, line 1: Can we also support yaml files? Or maybe instead of JSON? JSON seems a little bit bloaty here I think

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/DFGFunctionSummaries.kt
```diff
@@ -0,0 +1,197 @@
...
+            if (mostPreciseClassEntries.size > 1) {
+                mostPreciseClassEntries.sortByDescending { it.signature?.size ?: 0 }
+            }
+            functionToDFGEntryMap[matchingEntries.first()]
+        } else {
+            null
+        }
+    }
+
+    private fun getAllSupertypes(type: Type?): Set<Type> {
```
oxisto, line 126: I am pretty sure, that we already have a function like this somewhere.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/DFGFunctionSummaries.kt
```diff
@@ -0,0 +1,197 @@
...
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.fasterxml.jackson.module.kotlin.registerKotlinModule
+import de.fraunhofer.aisec.cpg.graph.declarations.FunctionDeclaration
+import de.fraunhofer.aisec.cpg.graph.declarations.MethodDeclaration
+import de.fraunhofer.aisec.cpg.graph.objectType
+import de.fraunhofer.aisec.cpg.graph.parseName
+import de.fraunhofer.aisec.cpg.graph.types.Type
+import java.io.File
+
+class DFGFunctionSummaries {
```
oxisto, line 38: Some documentation for the class and most of its functions would be nice

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/DFGFunctionSummaries.kt
```diff
@@ -0,0 +1,197 @@
...
+                } else if (entry.to.startsWith("return")) {
+                    val returnIndex = entry.to.removePrefix("param").toInt()
+                    // TODO: It would be nice if we could model the index. Not sure how this is done
+                    functionDeclaration
+                } else {
+                    null
+                }
+            // TODO: It would make sense to model properties here. Could be the index of a return
+            // value, full vs. partial flow or whatever comes to our minds in the future
+            to?.let { from?.addNextDFG(it) }
```
oxisto, line 172: So this only connects the flow *inside* the function declaration, but not back to the call expression, correct?

So if I have a call like `foo(a, b)` and model a flow from param0 to param1, I will see a flow from one param declaration to the other, but I won't see a flow at the call site from `a`, to `b`?

cpg-core/src/test/resources/function-dfg.json
KuechA: We now support json and yml

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/DFGFunctionSummaries.kt
KuechA: Indeed, there was an `ancestor` property/method somewhere. It does not the exact same thing but it's good enough.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/DFGFunctionSummaries.kt
KuechA: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/DFGFunctionSummaries.kt
KuechA: This is correct and consistent with handling any other function declaration. I'll think about a solution for this issue

cpg-core/src/test/resources/function-dfg.json
```diff
@@ -0,0 +1,56 @@
...
+    "functionDeclaration": {
+      "language": "de.fraunhofer.aisec.cpg.frontends.java.JavaLanguage",
+      "methodName": "java.util.List.addAll",
+      "signature": ["int", "java.util.Object"]
+    },
+    "dataFlows": [
+      {
+        "from": "param1",
+        "to": "base",
+        "dfgType": "full"
```
konradweiss, line 12: Using Json to then specify partial DFGs is than not possible correct? because we cannot specify the target as string? Or can we just specify partial here and what expression is associated can be inferred?

oxisto: In theory we could use the shortcut API to lookup a node using a name for this, but it gets complicated when there are multiple nodes with the same name, but different types. I would keep this as future work.

konradweiss: The node that would be referenced in that string is somewhere along the DFG leading to that call, so it can be identified more easily i think, but we can do that at some future point as you said.

KuechA: I'm not sure if it's easy to identify a node based on the path in all cases. Most likely, the "partial" DFG edge would be annotated with a `FieldDeclaration`, right? Since that one should have a (unique) FQN, we could use that name here.

Another use-case for this property could be to pass some index for a return value like a tuple or multi-value return instead of using the `returnX` thing I'm using right now. I'm not completely sure if/how we want to note all these things though.

I'd like to finalize the JSON/YAML format within this PR as good as we can, so if you have ideas/wishes on other scenarios and on how to note them, it would be great to add this here.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/DFGFunctionSummaries.kt
KuechA: I proposed a solution for this issue. The idea is that we save "writes" to a parameter (i.e., when it receives an incoming DFG edge) in a map in the `DFGSummaries` object. We also collect all function calls to inferred functions in the standard `DFGPass`. After having iterated the whole graph, we check the calls to the collected functions again, look up the parameters which flow out of the function declaration and the parameters to the arguments and draw these edges. The edges between the arguments and the parameters are labeled with the `CALLING_CONTEXT_IN` and `CALLING_CONTEXT_OUT` to enable context sensitivity in this case.

The only downside of this approach is that neo4j-ogm requires a custom serialization. I used `toString()` for now but we could also rely on the `id` instead. For some reason, the neo4j test still fails and I'm not sure why :/

cpg-core/src/test/resources/function-dfg.json
KuechA: ok, we keep it for the future

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -388,6 +390,34 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : EOGStarterPass
                 currentNode,
                 PowersetLattice(identitySetOf(currentNode))
             )
+        } else if (currentNode is CallExpression) {
+            val functionsWithSummaries =
+                currentNode.invokes.filter {
+                    it in ctx.config.functionSummaries.functionToChangedParameters
+                }
+            if (functionsWithSummaries.isNotEmpty()) {
```
oxisto, line 398: Some documentation of these steps would be nice

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -420,6 +435,16 @@ private constructor(
             return this
         }
 
+        fun registerFunctionSummaries(functionSummaries: List<File>): Builder {
```
oxisto, line 438: Wouldn't it make more sense to make this a File varargs? Then we can either have one file or a list of files instead of having two functions.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -397,6 +401,17 @@ private constructor(
             return this
         }
 
+        inline fun <reified P : Pass<*>> unregisterPass(): Builder {
```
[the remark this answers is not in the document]
KuechA: Only a test. I replaced it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -122,11 +120,15 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : EOGStarterPass
                     )
                 }
             } else {
-                key.addAllPrevDFG(
-                    value.elements.filterNot {
-                        (it is VariableDeclaration || it is ParameterDeclaration) && key == it
+                value.elements.forEach {
+                    if ((it is VariableDeclaration || it is ParameterDeclaration) && key == it) {
+                        // Nothing to do
+                    } else if (it in edgePropertiesMap && edgePropertiesMap[it] is CallingContext) {
```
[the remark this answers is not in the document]
KuechA: Yes...this maps stores additional information for the dfg edges. I added some documentation and we may want to change this in the future and encode more information in the state directly.

oxisto (approved):

PR #1438, opened 2024-02-18
Adding specific classes for property edges, such as `Dataflow`

oxisto:
This PR adds specific classes for our property edges. Starting with `Dataflow`. We will extend this in the future.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
@@ -31,10 +31,7 @@ import de.fraunhofer.aisec.cpg.PopulatedByPass
 import de.fraunhofer.aisec.cpg.TranslationContext
 import de.fraunhofer.aisec.cpg.frontends.Handler
 import de.fraunhofer.aisec.cpg.frontends.Language
-import de.fraunhofer.aisec.cpg.graph.declarations.FieldDeclaration
-import de.fraunhofer.aisec.cpg.graph.declarations.MethodDeclaration
-import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration
-import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
+import de.fraunhofer.aisec.cpg.graph.declarations.*
 import de.fraunhofer.aisec.cpg.graph.edge.*
 import de.fraunhofer.aisec.cpg.graph.edge.Properties
 import de.fraunhofer.aisec.cpg.graph.scopes.GlobalScope
@@ -254,9 +251,8 @@ open class Node : IVisitable<Node>, Persistable, LanguageProvider, ScopeProvider
         next: Node,
         granularity: GranularityType = GranularityType.FULL,
         memberField: FieldDeclaration? = null,
-        legacyProperties: MutableMap<Properties, Any?> = EnumMap(Properties::class.java)
     ) {
-        val edge = Dataflow(this, next, granularity, memberField, legacyProperties)
+        val edge = Dataflow(this, next, granularity, memberField)
         nextDFGEdges.add(edge)
         next.prevDFGEdges.add(edge)
     }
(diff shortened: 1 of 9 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edge/Dataflow.kt
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * This edge class defines a flow of data between [start] and [end]. The flow can have a certain
+ * [granularity].
+ */
+@RelationshipEntity
+class Dataflow(
+    start: Node,
+    end: Node,
+    val granularity: GranularityType = GranularityType.FULL,
+    val memberField: FieldDeclaration? = null,
```
KuechA, line 51: Could we also directly add the `index` and `calling_context_in` and `calling_context_out` features? Do we really want this `legacyProperties` map? I'd prefer the ability to generate the map from the class' properties in a getter or make the data available to the common functions of the `PropertyEdge` base-class if necessary for backwards compatibility.

oxisto: No we don't want this legacy property map, at least not in the "public" functions I can have a look, whether I can generate the map behind the scenes. I will probably not get rid of it all-to-gether at this point.

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -251,9 +249,10 @@ open class Node : IVisitable<Node>, Persistable, LanguageProvider, ScopeProvider
 
     fun addNextDFG(
         next: Node,
-        properties: MutableMap<Properties, Any?> = EnumMap(Properties::class.java)
+        granularity: GranularityType = GranularityType.FULL,
+        memberField: FieldDeclaration? = null,
```
konradweiss, line 253: Here I think we need some doc. The function was simple previously but now we do have more parameters, especially what a `memberField` is an when it should be set would be good here.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -272,9 +271,10 @@ open class Node : IVisitable<Node>, Persistable, LanguageProvider, ScopeProvider
 
     open fun addPrevDFG(
         prev: Node,
-        properties: MutableMap<Properties, Any?> = EnumMap(Properties::class.java)
+        granularity: GranularityType = GranularityType.FULL,
+        partialTarget: Declaration? = null,
```
konradweiss, line 275: Similar here. I would also consider renaming `partialTarget` to `target` or something similar. At least to me it is confusing why the DFG that can be `FULL` is related to a declaration that is only a partial target.

konradweiss (changes requested):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
oxisto: It is only applicable to partial flows. See https://github.com/Fraunhofer-AISEC/cpg/blob/2083d20af8fc0eaf0a104088765bb35e7000d61b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edge/Dataflow.kt#L56-60.

I could add an additional check if granularity is really set to partial if partialTarget is set.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
oxisto: Agree. The membeField naming is a leftover. It should also be called partialTarget here.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
oxisto: I have tried another approach using sealed classes. Please have a look whether this is more intuitive @konradweiss @KuechA

konradweiss (approved):
The target being bound to the creation of the data flow and not having to set it when adding the DFG is definitely making it more intuitive.

Issue #1444, opened 2024-03-04
Frontend cannot assign type to modulo operation


robinmaisch:
Hello, CPG team,
Currently, in `de.fraunhofer.aisec.cpg.frontends.Language`, the modulo operator `%` is not handled in the method `propagateTypeOfBinaryOperation`, which causes BinaryOperator nodes with the operatorCode `%` to be assigned the `UNKNOWN` type, at least using the `JavaLanguage` . That should be a quick bug to fix, I think.

Cheers!

oxisto:
> Hello, CPG team, Currently, in `de.fraunhofer.aisec.cpg.frontends.Language`, the modulo operator `%` is not handled in the method `propagateTypeOfBinaryOperation`, which causes BinaryOperator nodes with the operatorCode `%` to be assigned the `UNKNOWN` type, at least using the `JavaLanguage` . That should be a quick bug to fix, I think.
> 
> Cheers!

Hi! Would you be able to supply a PR for this?

robinmaisch:
I'd be happy to, see #1449

oxisto:
Fixed by https://github.com/Fraunhofer-AISEC/cpg/pull/1449

PR #900, opened 2022-09-01
Introduction of new query extensions for `Node`

oxisto:
This PR introduces a couple of new extensions for `Node`, similar to `calls`, which has been introduced in #765 for functions, methods, literals and references. They return all children in a flattened AST starting from the particular node (which can also be a translation result).

Furthermore, the operator extensions `[]` and ' ()'. have been defined on `List<Node>`, which can be used in the following way(s):

## Retrieving nodes by their type

```kotlin
import de.fraunhofer.aisec.cpg.graph.*

var functions = result.functions
var locals = myFunction.variables
```

## Retrieve a particular node

Retrieving a particular node, such as a `VariableDeclaration` from the graph by a given name. The query can be further extended by so called `SearchModifiers`, e.g., to throw an error if the result is not unique.
```kotlin
import de.fraunhofer.aisec.cpg.graph.*

// returns the first variable in the graph which has the name "a"
var a = result.variables["a"]

// returns the only variable with the name "a" or an exception otherwise
var theOnlyA = result.variables["a", SearchModifiers.UNIQUE]
```

If one needs to filter by further attributes, a `predicate` can be used instead of the name.

```kotlin
import de.fraunhofer.aisec.cpg.graph.*

// returns the first variable in the graph which does have an initialiser
var anyWithInitializer = result.variables[{ it.initializer != null }]

// returns the only variable in the graph which does not have an initialiser or throws an exception
var uniqueWithInitializer = result.variables[{ it.initializer != null }, SearchModifiers.UNIQUE]
```

## Retrieve a list of nodes 

Sometimes, retrieving a filtered list of nodes is needed instead of a single node. In this case, the `()` operator can be used to quickly filter a list of nodes either according to a name or a predicate.

```kotlin
import de.fraunhofer.aisec.cpg.graph.*

// returns a list of all VariableDeclarations in the graph with the name "a"
var aList = result.variables("a")

// returns a list of FunctionDeclarations that have no parameter
var noArgs = result.functions { it.parameters.isEmpty() }
```

PR #1449, opened 2024-03-07
Fixes to type propagation of arithmetic expressions

robinmaisch:
Concerning #1444 (and more small problems came up when dealing with it):

 - `BinaryOperator`s with the operator code `%`, `||`, `&&`, and `>>>` are now typed correctly. I could verify that the propagation is valid for the first three in all supported languages, and for `>>>`, the left operand type is used, like with the other bit shift operators.
 - Java float literals, indicated by the flag `f` or `F`, now get the FloatingPointType for floats instead of doubles. The JavaParser library does not differentiate between float and double literals, so a manual check is necessary.
 - the type propagation of BooleanType binary operations is now included in  `arithmeticOpTypePropagation`, as BooleanType is a subclass of NumericType.
 - simplified the resulting logic of `propagateTypeOfBinaryOperation` a bit, as `arithmeticOpTypePropagation` can already handle invalid cases
 - Modified a test that fails under Windows because of hardcoded Linux path separators
 - Added a test for the above changes
 - Added a test file for the test


cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.kt
```diff
@@ -123,7 +123,7 @@ class ExpressionHandler(lang: JavaLanguageFrontend) :
 
     /**
      * Creates a new [NewArrayExpression], which is usually used as an initializer of a
-     * [VariableDeclaration].
+     * [de.fraunhofer.aisec.cpg.graph.declarations.VariableDeclaration].
```
oxisto, line 126: Would be more consistent to import `VariableDeclaration` instead.

cpg-language-java/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontendTest.kt
```diff
@@ -800,4 +797,65 @@ internal class JavaLanguageFrontendTest : BaseTest() {
...
+
+        val language = JavaLanguage()
+        val result =
+            TestUtils.analyze(listOf(file), file.parentFile.toPath(), true) {
+                it.registerLanguage(language)
+            }
+        val tu =
+            findByUniqueName(result.components.flatMap { it.translationUnits }, file.toString())
+
+        val record = tu.getDeclarationAs(0, RecordDeclaration::class.java)
```
oxisto, line 813: instead of using findByUniqueName, getDeclarationAs, etc. you should use our "new" Kotlin-based shortcut API. Unfortunately, almost all unit test in the java frontend still use the old syntax, but you can for example look at the Go frontend for some inspiration.

There is some documentation on how to use it also on the PR that introduced it: https://github.com/Fraunhofer-AISEC/cpg/pull/900

You can use it to quickly retrieve nodes by their name, e.g. `tu.functions["main"]`

robinmaisch: Right, that's already much nicer!

cpg-language-java/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontendTest.kt
```diff
@@ -807,52 +807,44 @@ internal class JavaLanguageFrontendTest : BaseTest() {
...
 
-        val mainMethod = record.methods[0]
+        val expressionLists = mainMethod.mcalls
+        assertEquals(6, expressionLists.size)
 
-        val intOperationsList = mainMethod.getBodyStatementAs(0, MemberCallExpression::class.java)
-        assertNotNull(intOperationsList)
+        val intOperationsList = expressionLists[0]
         assertEquals(14, intOperationsList.arguments.size)
         assertTrue { intOperationsList.arguments.all { language.builtInTypes["int"]!! == it.type } }
```
oxisto, line 821: small nitpick: We try to avoid the forced non-null `!!` in Kotlin. The "canonical" way to do would be to use `primitiveType("int")` in order to retrieve the type. But that needs a language context, so you would need to do something like this:

```Kotlin
val main = record.methods["main"]
with(main) {
   val expressions = it.mcalls
   val intOperations = expressions[0]
   assertTrue { intOperations.arguments.all { primitiveType("int") == it.type } }
}
```

You can also use `with(language)`

robinmaisch: I see – done.

oxisto (approved):

konradweiss (approved):
LGTM

PR #1468, opened 2024-03-20
Use currentClass for implicit this

### LINKED ISSUE #1455: MemberCall inside LambdaExpression causes Parsing failure
I have created a small failing Java example:

```
public class SquareMapper {

 	public int square(int i) { return i*i; }
	
	public List<Integer> squareAll(List<Integer> is) {
		return is.stream().map(i -> square(i)).toList(); 
	}
	
	public static void main(String[] args) {
		SquareMapper mapper = new SquareMapper();
		List<Integer> squares = mapper.squareAll(List.of(1,2,3,4));
		squares.forEach(sq -> System.out.println(sq));
	}

}
```

In `squareAll`, the instance method `square` is referenced, but the base `this` is left implicit.
In the Java frontend, this causes `ExpressionHandler.createImplicitThis()` (line 793ff.) to throw a ClassCastException because the `currentFunction` is a `FunctionDeclaration`, not a `MethodDeclaration`. The actual current method may be determined by iterating through the parents of the current scope, for example. Alternatively, the ScopeManager could offer a reference to the `currentMethod` in addition to the `currentFunction`. Or, `thisType` could be determined by the current record instead, I think.

If `this` is added as an explicit base for the `MemberCallExpression`, then the `MemberCallExpression` resolves correctly.

Remark: As JavaParser's `MethodReferenceExpr`s have no corresponding handler method mapped to them yet, replacing the `LambdaExpression` in line 6 by `this::square` results in a `ProblemExpression`.

I am very sorry to keep bugging you but I keep stumbling over these things, so I thought I might as well. Still, this project aids me greatly in my work, so thank you!

konradweiss:
Adds a getter for the closest enclosing function scope that is associated with a `MethodDeclaration` and therefore can be a target for `this`. The getter is then used in `ExpressionHandler.createImplicitThis`.

Fixes #1455


cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.kt
```diff
@@ -771,9 +770,7 @@ class ExpressionHandler(lang: JavaLanguageFrontend) :
     private fun createImplicitThis():
         de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression {
         val base: de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
-        val thisType =
-            (frontend.scopeManager.currentFunction as? MethodDeclaration?)?.receiver?.type
-                ?: unknownType()
+        val thisType = frontend.scopeManager.currentMethod?.receiver?.type ?: unknownType()
         base = newReference("this", thisType).implicit("this")
```
oxisto, line 774: Shouldn't this be the current class instead of using the current method and the receiver? Seems more error-prone that using the current class.

konradweiss: What about one of those C++ situations where a function definition is not inside the class definition? I could see that there, the receiver is still correct but the current class points to the wrong type for the this reference.

oxisto: > What about one of those C++ situations where a function definition is not inside the class definition? I could see that there, the receiver is still correct but the current class points to the wrong type for the this reference.

No, in this case, the "currentClass" is also the "correct" class, since we jump to the scope of the class when entering the external method declaration.

oxisto: This happens here: https://github.com/Fraunhofer-AISEC/cpg/blob/662a730234a9e16e6a69ae16caf30b06786226bf/cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/DeclarationHandler.kt#L151-L158

`holder` is either a class or a namespace.

konradweiss: Ok, than it is better to go over the `currentClass`, I would still keep the currentMethod function though for future use.

oxisto (approved):

PR #1469, opened 2024-03-22
Adds pass dependencies for `EdgeCachePass`

fwendland:
The `EdgeCachePass` acts as a cache for AST, DFG and EOG edges.

Currently, the `EdgeCachePass` does not depend on any other passes. This can cause incomplete edge caches, if passes, excecuted later, add new edges.

This change adds pass dependencies for all current passes that add AST, DFG or EOG edges.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EdgeCachePass.kt
```diff
@@ -82,6 +83,13 @@ object Edges {
  *
  * The cache itself is stored in the [Edges] object.
  */
+@DependsOn(TypeResolver::class)
+@DependsOn(TypeHierarchyResolver::class)
+@DependsOn(EvaluationOrderGraphPass::class)
+@DependsOn(SymbolResolver::class)
+@DependsOn(DFGPass::class)
+@DependsOn(DynamicInvokeResolver::class)
+@DependsOn(ControlFlowSensitiveDFGPass::class)
```
oxisto, line 92: I would only include EvaluationOrderGraphPass, SymbolResolver, and the DFG passes here. Those "should" be enough

PR #1473, opened 2024-03-23
Only follow full DFG flow in `followNextDFGEdgesUntilHit` and similar

oxisto:
We had a problem that the extension functions, such as `followNextDFGEdgesUntilHit` were following all DFG flows, including partial ones. This lead to unexpected results.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
@@ -233,13 +233,13 @@ fun Node.followPrevDFGEdgesUntilHit(predicate: (Node) -> Boolean): FulfilledAndF
 
     while (worklist.isNotEmpty()) {
         val currentPath = worklist.removeFirst()
-        if (currentPath.last().prevDFG.isEmpty()) {
+        if (currentPath.last().prevFullDFG.isEmpty()) {
             // No further nodes in the path and the path criteria are not satisfied.
             failedPaths.add(currentPath)
             continue
         }
 
-        for (prev in currentPath.last().prevDFG) {
+        for (prev in currentPath.last().prevFullDFG) {
             // Copy the path for each outgoing DFG edge and add the prev node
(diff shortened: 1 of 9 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -233,13 +233,13 @@ fun Node.followPrevDFGEdgesUntilHit(predicate: (Node) -> Boolean): FulfilledAndF
 
     while (worklist.isNotEmpty()) {
         val currentPath = worklist.removeFirst()
-        if (currentPath.last().prevDFG.isEmpty()) {
+        if (currentPath.last().prevFullDFG.isEmpty()) {
```
konradweiss, line 236: I think we either need to rename the function reflecting that we are only following full dfgs or make it configurable. if the configuration has the default value full, we may have the same problem with confusing naming, so the default should not restrict the granularity.

konradweiss (changes requested):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
oxisto: Yes that is true, ok that sounds like a potentially bigger issue/discussion. Can we put this on the agenda for the dev meeting on 4/4/24?.

konradweiss: I thought about simply renaming the function to `followPrevFullDFGEdgesUntilHit` as a minimum requirement to merge this PR as it sound rather urgent and then merge it. If there are more issues or better solutions we can put it on the agenda.

oxisto: Sounds also reasonable to me

KuechA (approved):

konradweiss (approved):

PR #1482, opened 2024-03-29
C++ translation improvements

### LINKED ISSUE #1481: C++ move semantics are not recognised
C++ has a `&&` operator, next to the single `&` reference operator.

```
class MyClass {
public:
    MyClass(const MyClass &m) {}
    MyClass(MyClass &&m) {}
};
```

But currently, they end up in the same `ReferenceType`, making both constructors indistinguishable. 

We could either add an enum or operator name to the `ReferenceType` or just make it a `ReferenceType(ReferenceType)` to at least distinguish it.

### LINKED ISSUE #1483: C++ frontend does not detect function overloading of const/non-const methods
In C++ you can overload the same function signature with a variant that is `const` (and then has a `MyClass* const` receiver) and a non-const variant. Currently, for us, we see the functions, but they have an identical function signature, since we do not take the const into account.

This might not be so trivial to solve.

### LINKED ISSUE #1489: Support typedef in class definition
In C++ it is possible to add a typedef in a class, meaning that the typedef name will be "scoped" to the class, for example:

```cpp
class SomeDataClass {
public:
    int doOperation();
};

class MyClass {
private:
    // This typedef is FQN'd to MyClass::Data, but it cannot be accessed
    // outside of the class because it is declared as private (however we
    // are not yet using visibility information).
    typedef SomeDataClass Data;

    Data *d;

public:
    MyClass() {
        d = new Data();
    }

    int myFunction() {
        return d->doOperation();
    }
};
``` 

In this case, the typedefs actual name is `MyClass::Data`.

oxisto:
This PR contains several C++ improvements, such as:

- Better support for typedefs in classes (we can resolve partial FQNs now, but a real generalised solution for this is needed, see #1533, in the long run we will handle typedefs differently anyway)
- Support for type aliases (`using myString = std::string`)
- Support for namespace aliases (`namespace estd :: std`)
- Detection of `const T&` == `T` in function overloading
- Better realisation of template types, e.g. also pointer-types of T (e.g. `T*`) are now realised to their concrete types (e.g. `int*`, if `T` is `int`)

Fixes #1481 
Fixes #1489 
Fixes #1483


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
```diff
@@ -115,6 +118,20 @@ private constructor(
             }
         }
 
+        // Print some graph statistics
```
oxisto, line 121: This should be moved to our internal application

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationContext.kt
```diff
@@ -54,4 +54,9 @@ class TranslationContext(
      * the [TranslationResult.finalCtx] this may either be null or the last component analyzed.
      */
     var currentComponent: Component? = null,
+
+    // TODO: move somewhere better or maybe make translation context the statistics holder instead
+    // of the result?
+    var inferredFunctions: Int = 0,
```
oxisto, line 60: not needed because we can just count inferred nodes at the end, should be removed

konradweiss (changes requested):

konradweiss (approved):

Issue #1045, opened 2023-01-17
CPG infers new function declaration when integer argument is castable


keremc:
# Unexpected behavior

When you call a function that takes a non-builtin integer type like `size_t` as a parameter, if the parameter value isn't already the expected type, CPG infers a new function declaration.

This is problematic when performing an analysis on the call graph, since I cannot inspect the definition of the called function in all cases.

# Code of interest

```c
// foo.h

#include <stdlib.h>
void foo(size_t);

// foo.c

void foo(size_t n) { }

// bar.c

#include "foo.h"

int main(int argc, char **argv)
{
        foo(argc);         // Infers a new function.
        foo((size_t)argc); // References the right function in the graph.
}
```

# Configuration

In cpg-console:

```kotlin
import java.io.File
import de.fraunhofer.aisec.cpg.frontends.CompilationDatabase
import de.fraunhofer.aisec.cpg.TranslationConfiguration
import de.fraunhofer.aisec.cpg.TranslationManager
val db = CompilationDatabase.fromFile(File("compile_commands.json"))
val translationConfiguration = TranslationConfiguration.builder()
    .defaultLanguages()
    .useCompilationDatabase(db)
    .sourceLocations(db.sourceFiles)
    .defaultPasses()
    .loadIncludes(true)
    .build()
val translationManager = TranslationManager.builder().config(translationConfiguration).build()
val translationResult = translationManager.analyze().get()
```

oxisto:
Thanks for reporting this! Looks like an issue with typedefs (which are a still a mess). After tackling naming in v6 we are now having a closer look at call resolving and typing in the next version.

Issue #1484, opened 2024-03-29
Move connecting of declaration and definition to C++ extra pass


oxisto:
Currently, connecting declarations and definitions is done in the frontend, which might have a limited view and is not really a clean approach. Instead, this should be moved to the already existing `CXXExtraPass`, potentially even into a completely language neutral pass, but we are currently only filling the `definition` field for C++.

It also seems to be somewhat untested, so test coverage should be improved as well.

oxisto:
Closed by #1505

Issue #194, opened 2020-08-06
Resolve definitions of function declarations


oxisto:
We currently only resolve definitions of *method* declarations, we should do the same for all functions. It might be best to do this in a pass, since the definition or a function might be in a different translation unit.

PR #1496, opened 2024-04-02
Introduction of concepts for next-generation call resolving

### LINKED ISSUE #406: Refactor CallResolver to allow for more precise resolutions
The CallResolver is getting increasingly hard to maintain, since we need to distinguish between the call resolution for C++ and for Java. This will probably become worse as we want to resolve function calls of experimental languages. @konradweiss suggested to incorporate different CallResolvers based on the language with a strategy pattern, so that we can provide more precise call resolution. @konradweiss  also noted that it is probably not sufficient to check for the LanguageFrontend as we are doing currently, but instead check the language on a lower level (nodes?). The reason for this would be that there are ways to mix multiple languages, even in the same TUD. If we discuss changes in the architecture of the passes, maybe we should also take #310 into consideration.

### LINKED ISSUE #1497: Migrate regular function call resolving to new resolver

### LINKED ISSUE #1498: Provide an initial implementation and concepts for the next-generation resolving
We should probably start with the implementation of the necessary concepts and classes and fill in the necessary placeholders step by step.

oxisto:
This PR tries to implement the initial concepts described in #704 and also migrates the resolving of regular function calls to the new system.

Fixes #406
Fixes #1498 
Fixes #1497


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -680,20 +681,101 @@ class ScopeManager : ScopeProvider {
...
+        /**
+         * The actual start scope of the resolution, after [extractScope] is called on the callee.
+         * This can differ from the original start scope parameter handed to [resolveCall] if the
+         * callee contains an FQN.
+         */
+        var actualStartScope: Scope?
+    ) {
+
+        enum class Type {
+            OK,
```
KuechA, line 715: Doc at least for PROBLEMATIC (= a state that should never happen) and AMBIGUOUS (= can happen but there are multiple options e.g. due to overriding. What's the issue here that prevents us from saying "OK"?)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -680,20 +681,101 @@ class ScopeManager : ScopeProvider {
...
+                it.name.lastPartsMatch(name) && it.hasSignature(call.signature, call.arguments)
             }
 
         return func
     }
 
+    class CallResolutionResult(
+        /** The original call expression. */
+        val call: CallExpression,
+        var candidateFunctions: List<FunctionDeclaration>,
```
KuechA, line 701: It's not quite clear why we need the `candidateFunctions` or the `call` in the result. It's either the input or an intermediate step, so it's not super relevant? Some doc for all the lists, sets, maps would be useful. Why is `candidateFunctions` a list and not a set (same for `bestViable`)?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -40,11 +39,32 @@ import de.fraunhofer.aisec.cpg.graph.statements.expressions.CallExpression
...
+     * super class in the hierarchy.
+     */
+    var depthDistance: Int
+)
+
+data object CastNotPossible : CastResult(-1)
+
+data object DirectMatchNoCastNeeded : CastResult(0)
+
+data object ImplicitCastNeeded : CastResult(1)
```
KuechA, line 64: Maybe merge with `CastNeeded`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -40,11 +39,32 @@ import de.fraunhofer.aisec.cpg.graph.statements.expressions.CallExpression
...
+     * be cast. For example, if the types are the same, the distance is 0. An implicit cast is 1.
+     * The distance from a derived class to its direct super class is also 1. It increments for each
+     * super class in the hierarchy.
+     */
+    var depthDistance: Int
+)
+
+data object CastNotPossible : CastResult(-1)
+
+data object DirectMatchNoCastNeeded : CastResult(0)
```
KuechA, line 62: Just `DirectMatch`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -273,6 +356,60 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
...
+        // result
+        val directMatches = result.signatureResults.entries.filter { it.value.isDirectMatch }
+        return if (directMatches.size > 1) {
+            // This is an ambiguous result
+            result.type = ScopeManager.CallResolutionResult.Type.AMBIGUOUS
+            // Let's return all direct matches
+            directMatches.map { it.key }
+        } else if (directMatches.size == 1) {
+            // Let's return the single direct match
+            listOf(directMatches.first().key)
```
KuechA, line 395: I'd somehow prefer to set the `result.type` here instead of having the default `OK` set in the `ScopeManager`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -273,6 +356,60 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
...
+        if (single != null) {
+            return listOf(single)
+        }
+
+        // Check for direct matches. Let's hope there is only one, otherwise we have an ambiguous
+        // result
+        val directMatches = result.signatureResults.entries.filter { it.value.isDirectMatch }
+        return if (directMatches.size > 1) {
+            // This is an ambiguous result
+            result.type = ScopeManager.CallResolutionResult.Type.AMBIGUOUS
```
KuechA, line 390: This should be fine if we have languages where we can overload or override functions/methods. Otherwise it would be PROBLEMATIC (e.g. in C), right? Can we use our LanguageTraits here to make this choice?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/FunctionDeclaration.kt
```diff
@@ -106,12 +106,8 @@ open class FunctionDeclaration : ValueDeclaration(), DeclarationHolder, EOGStart
                     }
                 })
 
-    /**
-     * This function checks, if the supplied [CallExpression] has the same signature as the current
-     * [FunctionDeclaration].
-     */
-    fun hasSignature(call: CallExpression): Boolean {
-        return hasSignature(call.signature, call.arguments)
+    fun hasSignatureWithImplicitCast(call: CallExpression): Boolean {
```
KuechA, line 109: Necessary/Ever used? Remove from this PR or implement it ;)

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CLanguage.kt
```diff
@@ -161,4 +161,63 @@ open class CLanguage :
...
+        targetHint: HasType?
+    ): CastResult {
+        val match = super.tryCast(type, targetType, hint, targetHint)
+        if (match != CastNotPossible) {
+            return match
+        }
+
+        // Numeric types can be cast implicitly
+        if (type is NumericType && targetType is NumericType) {
+            return ImplicitCastNeeded
```
KuechA, line 198: Double-check spec if all numbers are on the same hierarchy level

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -479,6 +480,85 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
+
+        when (result.type) {
+            ScopeManager.CallResolutionResult.Type.PROBLEMATIC -> {
+                log.error(
+                    "Resolution of ${call.name} returned an problematic result and we cannot decide correctly, the invokes edge will contain all possible candidates"
+                )
+                call.invokes = result.bestViable.toList()
+            }
+            ScopeManager.CallResolutionResult.Type.AMBIGUOUS -> {
+                log.error(
```
KuechA, line 525: Is `log.error` appropriate here or is it only a warning?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -479,6 +480,85 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
+                call.invokes = result.bestViable.toList()
+            }
+            ScopeManager.CallResolutionResult.Type.OK -> {
+                call.invokes = result.bestViable.toList()
+            }
+            ScopeManager.CallResolutionResult.Type.UNRESOLVED -> {
+                // Resolution has provided no result, we can forward this to the inference system,
+                // if we want. While this is definitely a function, it could still be a function
+                // inside a namespace. We therefore have two possible start points, a namespace
+                // declaration or a translation unit. Nothing else is allowed (fow now). We can
```
KuechA, line 537:
```suggestion
                // declaration or a translation unit. Nothing else is allowed (for now). We can
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
oxisto: A warning is probably better.

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CLanguage.kt
oxisto: It seems so, because if you have something like this:
```cpp
void f(int x) {}
void f(float x) {}

int main() {
   f(2.0); // 2.0 is a double
}
```

Then clang says its ambiguous, so it seems that double to float has the same ranking as double to int.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
oxisto: A would keep an error for "PROBLEMATIC" though

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
oxisto: Do you have a better naming idea for `Type`?

KuechA:
Did you check if this PR fixes #1045?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
oxisto: I would keep it to give the language refinement method as much information as possible, if a language decides to implement something very very accurate. Also its nice to have from a debugging perspective if you look at a finished result and see what the original candidates were.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
oxisto: I renamed it to `SuccessKind` for now

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
oxisto: We probably need to differentiate between overloading and overriding. Even if you overload a function, there can be a point, in which we cannot differentiate between two function candidates, which are equally ranked and thus ambiguous. A compiler will return an error. I described this now in the doc.

I am not 100 % sure how we will deal with overridden functions, in this case they could still both match but the result is probably "ok". Not sure if we will only encounter this once I try to implement method resolution.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/FunctionDeclaration.kt
oxisto: Not used. can be removed

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
oxisto: I can try, the only issue is that we need to set the PROBLEMATIC already in resolveCall when we have a language that does not allow function overloading but more than one candidate function was returned. We are basically not allowed to "recover" from the problematic state, so that's why I handle it in resolveCall.

Open to suggestions :)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
KuechA: I agree that we have to differentiate between these two issues and that it's probably a future topic.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
KuechA: ok, so we keep them. Remaining points:
* Documentation of `candidateFunctions`, `viableFunctions`, `signatureResults`, `bestViable`
* `List` vs `Set`: There's no ordering and no duplicates, so a `Set` should work in all cases

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
KuechA: Hmm but couldn't we currently recover from it if there were multiple direct matches (which would be very weird) or if there is no direct match but several equally-ranked matches? Then, we should probably stop the matching after l. 796 in ScopeManager and set the remaining fields there or at least, we should not enter this function (which we do at the moment)?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
oxisto: Added more documentation and made all of them a set.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
oxisto: > Hmm but couldn't we currently recover from it if there were multiple direct matches (which would be very weird) or if there is no direct match but several equally-ranked matches? 

Ah you are correct, in this case we would "upgrade" from problematic to ambiguous, which is not intended.

> Then, we should probably stop the matching after l. 796 in ScopeManager and set the remaining fields there or at least, we should not enter this function (which we do at the moment)?

I think it's probably easiest if we do not enter bestViableResolution in this case then. That's how I implemented it now.

oxisto:
> Did you check if this PR fixes #1045?

I theory yes, in practice this still fails because of #194, which we probably should fix first then.

KuechA (approved):

oxisto:
> > Did you check if this PR fixes #1045?
> 
> I theory yes, in practice this still fails because of #194 because the function definition and declaration are not connected and both end up as candidates and we therefore think this is a problematic result :(

After #1484, this now also fixes #1045, at least for the case where the stdlib is included. Otherwise we cannot know that `size_t` is an integer and it will not match.

oxisto:
@KuechA I will wait until next week to merge this after @maximiliankaul  and @konradweiss are back for some more discussions. It is important to get the concepts correct this time.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -898,3 +976,170 @@ class ScopeManager : ScopeProvider {
...
+    /**
+     * A set of candidate symbols we discovered based on the [CallExpression.callee] (using
+     * [SymbolResolver.findSymbols]), more specifically a list of [FunctionDeclaration] nodes.
+     */
+    var candidateFunctions: Set<FunctionDeclaration>,
+
+    /**
+     * A set of functions, that restrict the [candidateFunctions] to those whose signature match.
+     */
+    var viableFunctions: Set<FunctionDeclaration>,
```
konradweiss, line 1078: Maybe matchingFunctions would be a better name, signatureMatchingFunctions is a bit lengthy, but to me the difference between candidateFunctions and viable functions is not clear enough from the name alone.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -246,54 +263,116 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
...
+    ): CastResult {
+        // We can take a shortcut if it is the same type
+        if (type == targetType) {
+            return DirectMatch
+        }
+
+        // We can also take a shortcut: if they are not of the same subclass, they will never
+        // match
+        if (type::class != targetType::class) {
+            return CastNotPossible
```
konradweiss, line 283: Are there no casts where our type classes can differ? I intuitively think of wrappers for primitive types in java but I don't know if our modeling currently has an issue with the casting here.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -188,6 +188,8 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
             // now).
             wouldResolveTo = scopeManager.resolveReference(current)
             if (wouldResolveTo !is VariableDeclaration && wouldResolveTo !is ParameterDeclaration) {
+                // TODO: not the best place for it and it is somewhat redundant
```
konradweiss, line 191: I would like a small explanation here why we do need it to find better places for it in future refactoring.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -254,6 +256,41 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
+        startScope: Scope? = scopeManager.currentScope
+    ): Set<Declaration> {
+        val (scope, name) = scopeManager.extractScope(nodeWithName, startScope)
+        val list =
+            scopeManager
+                .resolve<Declaration>(scope, true) { it.name.lastPartsMatch(name) }
+                .toMutableSet()
+        // If we have both the definition and the declaration of a function declaration in our list,
+        // we chose only the definition
+        val it = list.iterator()
```
konradweiss, line 280: Is this faster than using a filter like `list.filter {...}` ?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -479,6 +516,77 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
             return
         }
 
+        // Let's do some next generation stuff here, but currently with some restrictions
```
konradweiss, line 519: I would prefere a comment detaching this from the next gen such that it is more understandable in the future and renaming the variable to useNonLegacyResolution or negating the bools and call it `useLegacyResolution`

konradweiss (changes requested):
The functionality looks like what we want and I have only some minor remarks on documentation and naming.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
oxisto: I took the naming from the C++ overload resolution algorithm: https://en.cppreference.com/w/cpp/language/overload_resolution#Viable_functions

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
oxisto: Unboxing of primitive types in Java is probably something that needs to be specified in the Java language

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
oxisto: Hmm it is probably a little bit faster since we are removing elements from the list that we need to check, so if we have a large list of function declarations, using the mutable iterator should be faster since we are removing those declarations step by step. Using the filter on the original list would mean that we need to loop through the original list items even though we know that the element was already removed.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
oxisto: I already moved it to the "better place"

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -898,3 +976,170 @@ class ScopeManager : ScopeProvider {
...
+        if (param.isVariadic) {
+            remainingArguments = 0
+            break
+        }
+
+        // Try to find a matching call argument by index
+        val type =
+            signature.getOrNull(i)
+                ?: // If the argument is null, we could still have a default argument
+                return if (useDefaultArguments) {
```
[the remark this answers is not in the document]
oxisto: Good point, it was a bit too much kotlin magic. I restructed the function for better readability.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
konradweiss: Ah there is the check if its definition is already in the list so we cannot merge the entire conditional removing into the filter or predicate of the scope resolution above. At least not trivially

konradweiss (approved):

PR #1499, opened 2024-04-03
Extracted call/cast replacement into separate pass

### LINKED ISSUE #1487: Extract handling of "functional style casts" into separate pass
Multiple languages, such as Go and C++ support "functional style casts", in the form of `int(5)`. During the frontend parsing, they are indistinguishable from function calls. Therefore, we need to do a cleanup after all types are known but before other passes are invoked that replace those calls with casts. The proposed solution is to include a new language trait `HasFuntionalCasts` and to move the logic from the `GoExtraPass` into a separate, language-neutral one.

oxisto:
Multiple languages, such as Go and C++ support "functional style casts", in the form of `int(5)`. During the frontend parsing, they are indistinguishable from function calls. Therefore, we need to do a cleanup after all types are known but before other passes are invoked that replace those calls with casts. The proposed solution is to include a new language trait `HasFuntionalCasts` and to move the logic from the `GoExtraPass` into a separate, language-neutral one.

Fixes #1487

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ReplaceCallCastPass.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ReplaceCallCastPass.kt
@@ -26,27 +26,37 @@
 package de.fraunhofer.aisec.cpg.passes
 
 import de.fraunhofer.aisec.cpg.TranslationContext
+import de.fraunhofer.aisec.cpg.frontends.Handler
 import de.fraunhofer.aisec.cpg.frontends.HasFunctionalCasts
+import de.fraunhofer.aisec.cpg.frontends.Language
+import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend
 import de.fraunhofer.aisec.cpg.graph.*
 import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.CallExpression
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.CastExpression
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.ConstructExpression
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.UnaryOperator
+import de.fraunhofer.aisec.cpg.graph.types.Type
 import de.fraunhofer.aisec.cpg.helpers.SubgraphWalker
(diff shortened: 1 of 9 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ReplaceCallCastPass.kt
```diff
@@ -0,0 +1,144 @@
...
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.CallExpression
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.ConstructExpression
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.UnaryOperator
+import de.fraunhofer.aisec.cpg.helpers.SubgraphWalker
+import de.fraunhofer.aisec.cpg.passes.order.DependsOn
+import de.fraunhofer.aisec.cpg.passes.order.ExecuteBefore
+
+@ExecuteBefore(EvaluationOrderGraphPass::class)
+@DependsOn(TypeResolver::class)
+class ReplaceCallCastPass(ctx: TranslationContext) : TranslationUnitPass(ctx) {
```
KuechA, line 41: Documentation missing

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ReplaceCallCastPass.kt
```diff
@@ -0,0 +1,144 @@
...
+            }
+        }
+    }
+
+    override fun cleanup() {
+        // Nothing to do
+    }
+}
+
+context(ContextProvider)
```
KuechA, line 106: Documentation

KuechA (approved):

oxisto:
@KuechA thanks for the initial check. I did some additional cleanup and introduced a new annotation that can be put on passes, that they are only executed if the target (e.g. a translation unit) of the pass has a specific language trait.

KuechA (approved):
We could rename the package `de.fraunhofer.aisec.cpg.passes.order` to something like `executionConfig` since the annotations do not only concern the order of passes but also if they are executed at all.

oxisto:
> We could rename the package `de.fraunhofer.aisec.cpg.passes.order` to something like `executionConfig` since the annotations do not only concern the order of passes but also if they are executed at all.

I had a similar thought, although I will do this in an extra PR.

PR #1510, opened 2024-04-10
Fixing javaparser resultion error

konradweiss:
When javaparser tries to resolve the type of `this` it failed in a case where it was used in an anomymouse inner class. In this case we now take the name of the current record from the scope manager.

### DIFF
--- a/cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.kt
+++ b/cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.kt
@@ -470,8 +470,13 @@ class ExpressionHandler(lang: JavaLanguageFrontend) :
 
     private fun handleThisExpression(expr: Expression): Reference {
         val thisExpr = expr.asThisExpr()
-        val resolvedValueDeclaration = thisExpr.resolve()
-        val type = this.objectType(resolvedValueDeclaration.qualifiedName)
+        var qualifiedName: String
+        try {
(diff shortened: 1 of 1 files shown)


cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/ExpressionHandler.kt
```diff
@@ -470,8 +470,13 @@ class ExpressionHandler(lang: JavaLanguageFrontend) :
 
     private fun handleThisExpression(expr: Expression): Reference {
         val thisExpr = expr.asThisExpr()
-        val resolvedValueDeclaration = thisExpr.resolve()
-        val type = this.objectType(resolvedValueDeclaration.qualifiedName)
+        var qualifiedName: String
+        try {
+            qualifiedName = thisExpr.resolve().qualifiedName
+        } catch (use: UnsolvedSymbolException) {
+            qualifiedName = frontend.scopeManager.currentRecord?.name.toString()
```
oxisto, line 477: Can't we just use our scope manager function in all cases?

KuechA: At least our tests work when we always use our scope manager.

konradweiss: ok we kann do that

oxisto (approved):

PR #1513, opened 2024-04-10
Consider definitions of `FunctionDeclarations` if there is no body

KuechA:


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -78,7 +78,7 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : EOGStarterPass
 
         // Calculate the complexity of the function and see, if it exceeds our threshold
         val max = passConfig<Configuration>()?.maxComplexity
-        val c = node.body?.cyclomaticComplexity ?: 0
+        val c = node.definition?.body?.cyclomaticComplexity ?: 0
```
oxisto, line 81: Why? If the body is not there, the pass will not do anything?

KuechA: Good point, probably it's not necessary

PR #1514, opened 2024-04-10
Use the max complexity config for ControlFlowSensitiveDFGPass in neoj and console

KuechA:
* Configures a max. complexity of 200 for the `ControlFlowSensitiveDFGPass`
* Turns off `parallelPasses` since it causes some errors.


cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -485,7 +485,12 @@ class Application : Callable<Int> {
                 .addIncludesToGraph(loadIncludes)
                 .debugParser(DEBUG_PARSER)
                 .useUnityBuild(useUnityBuild)
-                .useParallelPasses(true)
+                .useParallelPasses(false)
```
oxisto, line 488: at least for the application I think it would be good to have this stored in a variable that we can (potentially) change from the outside, for the cog-console I think its ok.

oxisto (approved):

PR #1515, opened 2024-04-10
Fixing placement of inferred record declarations

oxisto:
Before this PR, we inferred all record declarations into the global scope, even though we knew (because of the FQN) that the record is part of another "name" (e.g. a namespace). This PR fixes that and correctly places the inferred record in the extracted scope of its FQN.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -447,7 +452,35 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
         }
-        return member ?: handleUnknownField(containingClass, reference)
+
+        if (member != null) {
+            return member
+        }
+
+        // This is a little bit of a workaround, but at least this makes sure we are not inferring a
+        // record, where a namespace already exist
+        val (scope, _) = scopeManager.extractScope(reference, null)
```
KuechA, line 462: This part is not really intuitive/understandable to me. Will we implement something for NameScope and else or does this have to remain as it is right now?

oxisto: Yes we will do that in the future

KuechA (approved):

PR #1517, opened 2024-04-12
Move `de.fraunhofer.aisec.cpg.passes.order` to `de.fraunhofer.aisec.cpg.passes.executionConfiguration`

### LINKED ISSUE #1503: Rename package `de.fraunhofer.aisec.cpg.passes.order` to something more fitting
We could rename the package `de.fraunhofer.aisec.cpg.passes.order` to something like `executionConfig` since the annotations do not only concern the order of passes but also if they are executed at all.

_Originally posted by @KuechA in https://github.com/Fraunhofer-AISEC/cpg/pull/1499#pullrequestreview-1981979427_

KuechA:
Closes #1503

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
@@ -35,7 +35,7 @@ import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdge
 import de.fraunhofer.aisec.cpg.graph.statements.IfStatement
 import de.fraunhofer.aisec.cpg.graph.statements.WhileStatement
 import de.fraunhofer.aisec.cpg.helpers.*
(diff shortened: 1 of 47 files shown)


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
```diff
@@ -35,7 +35,7 @@ import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdge
 import de.fraunhofer.aisec.cpg.graph.statements.IfStatement
 import de.fraunhofer.aisec.cpg.graph.statements.WhileStatement
 import de.fraunhofer.aisec.cpg.helpers.*
-import de.fraunhofer.aisec.cpg.passes.order.DependsOn
+import de.fraunhofer.aisec.cpg.passes.executionConfiguration.DependsOn
```
oxisto, line 38: I am not a big fan of such a long package name (including the camelCase). Just `configuration`?

oxisto (approved):

PR #1521, opened 2024-04-12
Using `io.joern:eclipse-cdt-core` release of CDT

oxisto:
It seems that joern provides releases of CDT on maven central. We could use this instead of having this rather ugly workaround with the ivy repository, which we need to change with every CDT release.

I am not sure how often they will update it. Any opinions on this?


oxisto:
Alternatively, I found a way to make the Ivy repository at least release-independent, so while a user still needs to configure it, at least it will not need to be changed per CDT release. See https://github.com/Fraunhofer-AISEC/cpg/pull/1522

PR #1522, opened 2024-04-12
Trying to use a release-independent repository for CDT

oxisto:
This is some rather ugly workaround in our ivy repository layout, but it should work for all future releases.


konradweiss (commented):
Could you provide a description where the benefits (or drawbacks) are compared to the previous way we included the CDT? I only have a slight idea that we will have to update version in the new repo but users are now always getting the one we specify there.

oxisto:
> Could you provide a description where the benefits (or drawbacks) are compared to the previous way we included the CDT? I only have a slight idea that we will have to update version in the new repo but users are now always getting the one we specify there.

Previously, users needed to adjust their "repository" in their gradle configuration if we updated CDT because the repository URL contained the CDT version (see https://github.com/Fraunhofer-AISEC/codyze/pull/791/files). We only reserved this for major versions of the CPG for now

With this new approach the repository URL stays the same and we can update the CDT version also in a minor version.

konradweiss (approved):

oxisto:
We should probably reserve this for release v9

PR #1525, opened 2024-04-13
Remove slow `toString` implementation

oxisto:
This PR limits ourselves to a very limited implementation of `toString`. The old approach using a to-string-builder and including a lot of fields was actually costing some performance, for example in the (debug) output of the inference system.

I would argue that toString should only include the very basic things to quickly identify a node, which is in most cases its name and location and then sparsley adding additional information. In one particular case the execution time of the symbol resolver was reduced from 3000ms to 500ms.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/IncludeDeclaration.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/IncludeDeclaration.kt
@@ -31,7 +31,6 @@ import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdge
 import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdge.Companion.propertyEqualsList
 import de.fraunhofer.aisec.cpg.graph.edge.PropertyEdgeDelegate
 import java.util.Objects
-import org.apache.commons.lang3.builder.ToStringBuilder
 import org.neo4j.ogm.annotation.Relationship
 
 /** This declaration represents either an include or an import, depending on the language. */
@@ -71,16 +70,6 @@ class IncludeDeclaration : Declaration() {
         val propertyEdge = PropertyEdge(this, problemDeclaration)
         propertyEdge.addProperty(Properties.INDEX, problemEdges.size)
         problemEdges.add(propertyEdge)
-    }
-
-    override fun toString(): String {
-        return ToStringBuilder(this, TO_STRING_STYLE)
-            .appendSuper(super.toString())
-            .append("name", name)
-            .append("filename", filename)
-            .append("includes", includes)
-            .append("problems", problems)
-            .toString()
     }
 
     override fun equals(other: Any?): Boolean {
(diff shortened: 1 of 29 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/IncludeDeclaration.kt
```diff
@@ -73,16 +72,6 @@ class IncludeDeclaration : Declaration() {
         problemEdges.add(propertyEdge)
     }
 
-    override fun toString(): String {
-        return ToStringBuilder(this, TO_STRING_STYLE)
-            .appendSuper(super.toString())
-            .append("name", name)
-            .append("filename", filename)
```
KuechA, line 80: Can we identify the filename from the `name`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/ProblemDeclaration.kt
```diff
@@ -39,10 +38,7 @@ class ProblemDeclaration(
 ) : ValueDeclaration(), ProblemNode {
 
     override fun toString(): String {
-        return ToStringBuilder(this, TO_STRING_STYLE)
-            .appendSuper(super.toString())
-            .append("problem", problem)
-            .toString()
+        return "${this::class.simpleName}(problem=$problem,type=$type,location=$location)"
```
KuechA, line 41: `type` or `problemType`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/DeclarationStatement.kt
```diff
@@ -77,10 +76,7 @@ open class DeclarationStatement : Statement() {
     }
 
     override fun toString(): String {
-        return ToStringBuilder(this, TO_STRING_STYLE)
-            .appendSuper(super.toString())
-            .append("declarations", declarations)
-            .toString()
+        return "${this::class.simpleName}(label=$declarations,location=$location)"
```
KuechA, line 79:
```suggestion
        return "${this::class.simpleName}(declarations=$declarations,location=$location)"
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/Literal.kt
```diff
@@ -41,10 +41,7 @@ class Literal<T> : Expression() {
     @Convert(ValueConverter::class) var value: T? = null
 
     override fun toString(): String {
-        return ToStringBuilder(this, TO_STRING_STYLE)
-            .appendSuper(super.toString())
-            .append("value", value)
-            .toString()
+        return "${this::class.simpleName}(value=$name,type=$type,location=$location)"
```
KuechA, line 44:
```suggestion
        return "${this::class.simpleName}(value=$value,type=$type,location=$location)"
```

PR #1528, opened 2024-04-14
Take operator into account when evaluating conditional expressions

### LINKED ISSUE #1527: There is a logic error in ValueEvaluator.
There is a logic error in `de.fraunhofer.aisec.cpg.analysis.ValueEvaluator#handleConditionalExpression`.

# testcase

```java
public void test() {
        int a = 1;
        a = a != 1 ? 2 : a;
        System.out.println(a);
}
```

When outputting the value of reference 'a', it should equal 1, but the result is 2.

oxisto:
Previously, the `ValueEvaluator` was assuming that a conditional expression was only using the `==` operator.

Fixes #1527

### DIFF
--- a/cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/ValueEvaluatorTest.kt
+++ b/cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/ValueEvaluatorTest.kt
@@ -772,7 +772,7 @@ class ValueEvaluatorTest {
     }
 
     @Test
-    fun handleConditionalExpression() {
+    fun testHandleConditionalExpression() {
         with(TestLanguageFrontend()) {
             val a = newVariableDeclaration("a")
             a.initializer = newLiteral(1)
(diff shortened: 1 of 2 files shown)


cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/ValueEvaluatorTest.kt
```diff
@@ -770,4 +770,23 @@ class ValueEvaluatorTest {
...
+
+            val aRef = newReference("a")
+            aRef.refersTo = a
+            aRef.prevDFG = mutableSetOf(a)
+
+            val comparison = newBinaryOperator("!=")
+            comparison.lhs = aRef
+            comparison.rhs = newLiteral(1)
+
+            val cond = newConditionalExpression(comparison, newLiteral(2), aRef)
```
KuechA, line 788: Not sure why this test ever worked because we didn't handle the `!=` operator if I see this correctly

KuechA (approved):

cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/analysis/ValueEvaluatorTest.kt
oxisto: This is a new test :)

oxisto: Hmm interesting ^^

Issue #1531, opened 2024-04-15
`DynamicInvokeResolver` runs in parallel to `ControlFlowSensitiveDFGPass`


konradweiss:
`DynamicInvokeResolver` runs in parallel with `ControlFlowSensitiveDFGPass`, but uses DFG edges and should IMO run either before or after (i think before). However, currently we don't have a way to specify an optional dependency, i.e. a dependency saying that the pass should run after the other, only if it was specified to run @RunAfter(ControlFlowSensitiveDFGPass::class) or a grouping of passes by semantic @RunAfter(DFG) and DFG = [`DFGPass`, `ControlFlowSensitiveDFGPass`].

KuechA:
The normal `DFGPass` also adds DFG edges and the `DynamicInvokeResolver` depends on this one already. The `ControlFlowSensitiveDFGPass` basically makes the edges better, so it could make sense to add a soft dependency on this pass too? There's a difference between using the Annotation as `@DependsOn(ControlFlowSensitiveDFGPass, false)` (default and would always register the `ControlFlowSensitiveDFGPass`) and `@DependsOn(DynamicInvokeResolver, true)` (wouldn't register the pass on its own but only enforces the order if the user added this pass). So, I think using the second version to annotate the `DynamicInvokeResolver` would do what you want.

konradweiss:
ah perfect, I did forget that we had that functionality already implemented somewhere.

PR #1532, opened 2024-04-15
Adding soft from DynamicInvokeResolver to ControlFlowSensitiveDFGPass

konradweiss:
Solves #1531

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DynamicInvokeResolver.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DynamicInvokeResolver.kt
@@ -58,7 +58,7 @@ import java.util.function.Consumer
  */
 @DependsOn(SymbolResolver::class)
 @DependsOn(DFGPass::class)
-@DependsOn(ControlFlowSensitiveDFGPass::class, true)
+@DependsOn(ControlFlowSensitiveDFGPass::class, softDependency = true)
 class DynamicInvokeResolver(ctx: TranslationContext) : ComponentPass(ctx) {
     private lateinit var walker: ScopedWalker
     private var inferDfgForUnresolvedCalls = false
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DynamicInvokeResolver.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DynamicInvokeResolver.kt
@@ -58,6 +58,7 @@ import java.util.function.Consumer
  */
 @DependsOn(SymbolResolver::class)
 @DependsOn(DFGPass::class)
+@DependsOn(ControlFlowSensitiveDFGPass::class, true)
 class DynamicInvokeResolver(ctx: TranslationContext) : ComponentPass(ctx) {
     private lateinit var walker: ScopedWalker
     private var inferDfgForUnresolvedCalls = false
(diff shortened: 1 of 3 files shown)


konradweiss:
Apparently there is now a problem with dependency ordering between passes. We have to solve that before this can be merged.

oxisto:
> Apparently there is now a problem with dependency ordering between passes. We have to solve that before this can be merged.

Hm yes it looks like the dependency matching algorithm is aborting because the `ControlFlowSensitiveDFGPass` is not registered in the test; but you explicitly added it as a "soft" dependency, so looks like this is broken :(

oxisto:
Can you try with the following patch?

```patch
diff --git a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/configuration/PassWithDependencies.kt b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/configuration/PassWithDependencies.kt
index 4ee5f627d..50ff80b02 100644
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/configuration/PassWithDependencies.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/configuration/PassWithDependencies.kt
@@ -64,4 +64,32 @@ data class PassWithDependencies(
         }
         return builder.toString()
     }
+
+    /**
+     * Checks whether the [dependencies] of this pass are met. The list of [softDependencies] and
+     * [hardDependencies] is removed step-by-step in
+     * [PassWithDepsContainer.getAndRemoveFirstPassWithoutDependencies].
+     */
+    fun dependenciesMet(workingList: MutableList<PassWithDependencies>): Boolean {
+        // In the simplest case all our dependencies are empty since they were already removed by
+        // the selecting algorithm.
+        if (this.dependencies.isEmpty() && !this.isLastPass) {
+            return true
+        }
+
+        // We also need to check, whether we still "soft" depend on passes that are just not
+        // there (after all hard dependencies are met), in this case we can select the pass
+        // as well
+        val remainingClasses = workingList.map { it.pass }
+        if (
+            this.hardDependencies.isEmpty() &&
+                this.softDependencies.all { !remainingClasses.contains(it) } &&
+                !this.isLastPass
+        ) {
+            return true
+        }
+
+        // Otherwise, we still depend on an unselected pass
+        return false
+    }
 }
diff --git a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/configuration/PassWithDepsContainer.kt b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/configuration/PassWithDepsContainer.kt
index e6b5988fb..c3219e55c 100644
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/configuration/PassWithDepsContainer.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/configuration/PassWithDepsContainer.kt
@@ -160,9 +160,8 @@ class PassWithDepsContainer {
 
             // Keep going until our dependencies are met, this will collect passes that can run in
             // parallel in results
-            if (
-                currentElement.dependencies.isEmpty() && !currentElement.isLastPass
-            ) { // no unsatisfied dependencies
+            if (currentElement.dependenciesMet(workingList)) {
+                // no unsatisfied dependencies
                 val result = currentElement.pass
                 results.add(result)
 ```

File: cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DynamicInvokeResolver.kt [hunk 1]
```diff
@@ -58,6 +58,7 @@ import java.util.function.Consumer
  */
 @DependsOn(SymbolResolver::class)
 @DependsOn(DFGPass::class)
+@DependsOn(ControlFlowSensitiveDFGPass::class, true)
```
oxisto, line 61:
```suggestion
@DependsOn(ControlFlowSensitiveDFGPass::class, softDependency = true)
```

For better readability, otherwise its hard to know what "true" is

oxisto (approved):

PR #1536, opened 2024-04-16
Initial implementation for symbols and imports

### LINKED ISSUE #1539: Implement concepts for symbols

### LINKED ISSUE #1540: Handle imports correctly

oxisto:
Tries to implement the concepts in #1533. This is not yet implemented for all language frontends. I am using the new concepts in the C/C++ and Go frontend as a show case.

- [x] Fixes #1539
- [x] Fixes #1540
- [x] Removing `aliases` and replacing them with imports

### DIFF
--- a/cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontend.kt
+++ b/cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontend.kt
@@ -34,6 +34,7 @@ import de.fraunhofer.aisec.cpg.frontends.golang.GoStandardLibrary.Modfile
 import de.fraunhofer.aisec.cpg.frontends.golang.GoStandardLibrary.Parser
 import de.fraunhofer.aisec.cpg.graph.*
 import de.fraunhofer.aisec.cpg.graph.declarations.DeclarationSequence
+import de.fraunhofer.aisec.cpg.graph.declarations.MethodDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
 import de.fraunhofer.aisec.cpg.graph.newNamespaceDeclaration
@@ -170,6 +171,11 @@ class GoLanguageFrontend(language: Language<GoLanguageFrontend>, ctx: Translatio
         scopeManager.resetToGlobal(tu)
         currentTU = tu
 
+        // We need to keep imports on a special file scope. We can simulate this by "entering" the
+        // translation unit
+        scopeManager.enterScope(tu)
+
+        // We parse the imports specifically and not as part of the handler later
         for (spec in f.imports) {
             val import = specificationHandler.handle(spec)
             scopeManager.addDeclaration(import)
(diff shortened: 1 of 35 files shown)


oxisto:
Putting this back to draft since I need to fix some test failures first.

cpg-language-llvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/llvm/DeclarationHandler.kt
```diff
@@ -219,15 +217,17 @@ class DeclarationHandler(lang: LLVMIRLanguageFrontend) :
...
 
+            frontend.scopeManager.enterScope(record)
+
             val field = newFieldDeclaration(fieldName, fieldType, listOf(), null, false)
 
             frontend.scopeManager.addDeclaration(field)
-        }
 
-        frontend.scopeManager.leaveScope(record)
+            frontend.scopeManager.leaveScope(record)
```
KuechA, line 226: Why do we enter and leave the record's scope for each of the fields separately now?

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontend.kt
```diff
@@ -205,11 +211,28 @@ class GoLanguageFrontend(language: Language<GoLanguageFrontend>, ctx: Translatio
             if (declaration is DeclarationSequence) {
                 declaration.declarations.forEach { scopeManager.addDeclaration(it) }
             } else {
-                scopeManager.addDeclaration(declaration)
+                // We need to be careful with method declarations. We need to put them in the
+                // respective name scope of the record and NOT on the global scope / namespace scope
+                // TODO: this is broken if we see the declaration of the method before the class :(
```
KuechA, line 216: Can this happen? If so, doesn't this indicate that it's more likely to be a job for a pass?

cpg-language-go/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontendTest.kt
```diff
@@ -1017,17 +1017,27 @@ class GoLanguageFrontendTest : BaseTest() {
     fun testBuildTags() {
         val stdLib = Path.of("src", "test", "resources", "golang-std")
         val topLevel = Path.of("src", "test", "resources", "golang", "buildtags")
+
+        // make sure we parse main.go as the last
```
KuechA, line 1021: Will the test fail if we change this ordering? Doesn't this suggest that we implemented some feature that is ultimately not reliably available?

cpg-language-llvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/llvm/DeclarationHandler.kt
oxisto: Because of the way parsing of "classes" work in LLVM-IR. We handle the "type" of each field and if the type is a structure, we create a new class. If we would just enter the record once, we would create the class-type of the field within the scope of the fields' record instead of the global scope. I am not sure why this worked before, but if I leave it like it was the name of the class was wrong and it was created in the wrong scope.

cpg-language-go/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontendTest.kt
oxisto: We always had this limitation, since we do not have a proper ordering of dependencies in the `SymbolResolver`. For Go, we somewhat minimise this problem because we have special functions that print out the package dependency tree if Go is installed and feed that as an input to the translation manager. But in the long run we should figure out the dependency tree based on our imports (that we have now). But this is probably work for an extra PR.

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontend.kt
oxisto: See https://github.com/Fraunhofer-AISEC/cpg/issues/1561

KuechA (approved):

Issue #1542, opened 2024-04-22
Python ExpressionHandler does not handle ASTSubscript


immqu:
I'm using the Cloud Property Graph (using cpg v8.1.2) to analyze Python code. The code is contained in this directory: https://github.com/clouditor/cloud-property-graph/tree/main/ppg-testing-library/Detectability/D4-detectable-at-storage/Python

The analysis fails with the stack trace below, since the ExpressionHandler cannot handle a Python.ASTSubscript

# Stack trace

```
kotlin.NotImplementedError: An operation is not implemented.
java.util.concurrent.ExecutionException: kotlin.NotImplementedError: An operation is not implemented.
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
	at io.clouditor.graph.App.doTranslate(App.kt:179)
	at io.clouditor.graph.App.call(App.kt:91)
	at io.clouditor.graph.TestUtilKt.executePPGAndQuery(TestUtil.kt:24)
	at io.clouditor.graph.DetectabilityTest.testD4Python(DetectabilityTest.kt:128)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
[...]

PR #1545, opened 2024-04-23
Implement ast.Subscript and ast.Slice

maximiliankaul:
#1542


immqu:
Thanks, I have tested the changes using the PPG and it works now!

KuechA (approved):

PR #1552, opened 2024-05-01
Model values of Enum entries in Java with constructor call

### LINKED ISSUE #1546: Java Frontend: EnumConstantDeclarations do not save their arguments
Currently, enum constants only save their name, so their arguments get lost.
The arguments are delivered by the Java Parser API, so I hope it is easy to add a property for them.

I am not sure how you decide whether to safe AST children as fields or create edges for them, and also I would not know how to integrate it into the DeclarationBuilder so that other language modules are not affected, so I'd rather leave this one up to you.

(I'm sorry that I did not get around to give feedback on your last proposed fix, but I did not run into any more trouble so far. Thanks!)

Example:
```
public enum ChessPiece {
      PAWN("Bauer",1),
      KNIGHT("Springer", 3),
      BISHOP("Läufer", 3),
      ROOK("Turm", 5),
      QUEEN("Dame", 9),
      KING("König", -1);

      private final String name;
      private final int value;

      ChessPiece(String name, int value) {...}
}
```

KuechA:
Hopefully fixes #1546

Feel free to comment whether the solution helps to do so.

### DIFF
--- a/cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/DeclarationHandler.kt
+++ b/cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/DeclarationHandler.kt
@@ -207,12 +207,10 @@ open class DeclarationHandler(lang: JavaLanguageFrontend) :
 
         if (frontend.scopeManager.currentScope is RecordScope) {
             // We need special handling if this is a so called "inner class". In this case we need
-            // to
-            // store
+            // to store
             // a "this" reference to the outer class, so methods can use a "qualified this"
             // (OuterClass.this.someFunction()). This is the same as the java compiler does. The
-            // reference
-            // is stored as an implicit field.
+            // reference is stored as an implicit field.
             processInnerRecord(recordDeclaration)
         }
         return recordDeclaration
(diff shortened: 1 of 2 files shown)


cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/DeclarationHandler.kt
```diff
@@ -405,7 +406,27 @@ open class DeclarationHandler(lang: JavaLanguageFrontend) :
...
+        val currentEnum = frontend.scopeManager.currentRecord
+        val result =
+            this.newEnumConstantDeclaration(enumConstDecl.nameAsString, rawNode = enumConstDecl)
+        if (enumConstDecl.arguments.isNotEmpty()) {
+            val arguments =
+                enumConstDecl.arguments.mapNotNull {
+                    frontend.expressionHandler.handle(it)
+                        as? de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
+                }
+            val matchingConstructor =
```
oxisto, line 418: Shouldn't this be done by some kind of call resolving in a later stage? But since the java frontend is in dire need of a complete overhaul, I won't block it :)

oxisto (approved):

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/DeclarationHandler.kt
KuechA: Since the entries of an enum aren't connected by the EOG, they won't be reached by the SymbolResolver if I remember this correctly. That's why I decided to add the target constructor here manually. But this is indeed problematic because the constructor might not have been parsed yet

KuechA: I added a comment, so we remember this in the future

PR #1558, opened 2024-05-17
Java frontend: Handle multi-field-declaration statements

### LINKED ISSUE #1548: Java Frontend: Multi-Field declaration statements omit all but their first declaration
In `de.fraunhofer.aisec.cpg.frontends.java.DeclarationHandler::242`:
> // TODO: can  field have more than one variable?

The answer is yes :)
Multi-variable declarations are wrapped in a DeclarationStatement, but statements only live inside of methods, so I guess multi-field declarations should be mapped to multiple independent FieldDeclarations. However, the most important thing is that the field declaration appears in the target RecordDeclaration.

Example: 
```java
public class Point {
    int x, y;
    ...
}
```
results in RecordDeclaration with only one field `x`.
Any access to `y` will be mapped to a placeholder field declaration.

KuechA:
Fixes #1548 by implementing the proposed measures. However, `handleFieldDeclaration` now only returns the last field/variable which is declared in a multi-field-declaration statement but since we never use the return value in this case, it should be fine.

### DIFF
--- a/cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/DeclarationHandler.kt
+++ b/cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/DeclarationHandler.kt
@@ -235,8 +235,8 @@ open class DeclarationHandler(lang: JavaLanguageFrontend) :
 
     fun handleFieldDeclaration(
         fieldDecl: com.github.javaparser.ast.body.FieldDeclaration
-    ): FieldDeclaration? {
-        var fieldDeclaration: FieldDeclaration? = null
+    ): DeclarationSequence {
+        val declarationSequence = DeclarationSequence()
 
         val modifiers = fieldDecl.modifiers.map { modifier -> modifier.keyword.asString() }
 
@@ -282,7 +282,7 @@ open class DeclarationHandler(lang: JavaLanguageFrontend) :
                     type.typeOrigin = Type.Origin.GUESSED
                 }
             }
-            fieldDeclaration =
+            val fieldDeclaration =
                 this.newFieldDeclaration(
                     variable.name.asString(),
                     type,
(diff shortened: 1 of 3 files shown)


cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/DeclarationHandler.kt
```diff
@@ -235,61 +235,64 @@ open class DeclarationHandler(lang: JavaLanguageFrontend) :
...
             }
+            fieldDeclaration =
+                this.newFieldDeclaration(
+                    variable.name.asString(),
+                    type,
+                    modifiers,
+                    initializer,
+                    rawNode = fieldDecl
+                )
+            frontend.scopeManager.addDeclaration(fieldDeclaration)
```
oxisto, line 293: You can use the magical `DeclarationSequence` instead of the single field declaration. Not sure if we need to handle it at the caller. But that is how we do it in C++ if it can return more than one declaration

cpg-language-java/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguageFrontendTest.kt
```diff
@@ -297,8 +296,11 @@ internal class JavaLanguageFrontendTest : BaseTest() {
             namespaceDeclaration?.getDeclarationAs(0, RecordDeclaration::class.java)
         assertNotNull(recordDeclaration)
 
-        val fields = recordDeclaration.fields.map { it.name.localName }
-        assertTrue(fields.contains("field"))
+        val fields = recordDeclaration.fields.map { it.name.localName }.toSet()
+        assertTrue(fields.containsAll(setOf("field", "field1", "field2")))
+
+        assertEquals(1, (recordDeclaration.fields["field1"]?.initializer as? Literal<*>)?.value)
```
oxisto, line 302: You can use the very cool `assertLiteralValue` function here :)

oxisto (approved):

PR #1564, opened 2024-05-22
Made build K2 ready and remove Gradle warnings

oxisto:
Please make sure to enable K2 mode in IntelliJ. Also removes some gradle warning


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -322,7 +322,7 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : EOGStarterPass
             val lhs = currentNode.lhs.singleOrNull()
             writtenDeclaration = (lhs as? Reference)?.refersTo
 
-            if (writtenDeclaration != null && lhs != null) {
```
KuechA, line 325: Not sure why this change is here

KuechA (changes requested):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
konradweiss: lhs != null can be inferred now

KuechA (approved):

PR #1573, opened 2024-06-03
Trying to fix switch EOG

### LINKED ISSUE #1568: Switch without default breaks EOG
# Unexpected behavior

The EOG does not consider the default case in switch statements if it is not explicitly stated. This is critical when all cases return. In the example code, the EOG does not contain the printf-call as long as the Line `//default:` is commented out. 

Screenshot from 2024-05-29 16-37-51

# Code of interest

```
  1 #include <stdio.h>                                                                                                                                        
  2                                                                                                                                                           
  3 int main() {                                                                                                                                              
  4   int a=0;                                                                                                                                                
  5   switch (a) {                                                                                                                                            
  6     case 1:                                                                                                                                               
  7       return 0;                                                                                                                                           
  8     //default:                                                                                                                                            
  9   }                                                                                                                                                       
 10   printf("%d\n", a);                                                                                                                                      
 11 }      
```

oxisto:
Not sure, if this really the "correct" solution. @konradweiss should probably have another look at this.

Fixes #1568

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
@@ -968,6 +968,14 @@ open class EvaluationOrderGraphPass(ctx: TranslationContext) : TranslationUnitPa
             }
             createEOG(subStatement)
         }
+
+        // If we do not have default statement, we also need to put the switch statement into the
+        // currentPredecessors, otherwise we will completely ignore everything that is "beyond" the
+        // switch statement
+        if (compound.statements.filter { it is DefaultStatement }.isEmpty()) {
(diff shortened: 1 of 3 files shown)


konradweiss:
Yes, this should be the correct fix. When implementing the EOG for the switch statement I thought the assumption that one case always has to match, therefore enforcing the presence of a default if the switches do not map exhaustively. 

This means with this PR we accept having an incorrect EOG edge if the cases alone handle all possible values of the expression in the switch but now support switch statements that don't enforce a `default` if the cases do not cover the expressions value exhaustively.

konradweiss (approved):
This should be an adequate improvement although it adds in incorrect edge if there is no `default` case but all `case:` cases exhaustively handle the selector expression.

oxisto:
> This should be an adequate improvement although it adds in incorrect edge if there is no `default` case but all `case:` cases exhaustively handle the selector expression.

Hmm any idea how we can do that?

PR #1586, opened 2024-06-11
Improvements to function and record inference

oxisto:
This PR achieves several things:

- We try to infer the return types of function based on their context
- A simple heuristic to better decide whether a function should be inferred in a class or in a global scope
- A simple heuristic to decide whether a record or a namespace should be inferred
- We recursively infer nested namespaces, e.g. `A::B::C`


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/InferenceConfiguration.kt
```diff
@@ -49,6 +49,12 @@ private constructor(
     /** Enables the inference of variables, such as global variables. */
     val inferVariables: Boolean,
 
+    /**
+     * A very EXPERIMENTAL feature. If this is enabled, we will try to infer return types of
+     * functions based on the context of the call it originated out of. This is disabled by default.
+     */
+    val inferReturnTypes: Boolean,
```
oxisto, line 56: I am somewhat satisfied with this feature, so we could also enable this by default

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/ArgumentHolder.kt
```diff
@@ -68,4 +68,6 @@ interface ArgumentHolder : Holder<Expression> {
     operator fun minusAssign(node: Expression) {
         removeArgument(node)
     }
+
+    fun hasArgument(expression: Expression): Boolean
```
oxisto, line 72: @konradweiss this is needed in order to check the context of a call expression in order to derive the function type. But adding this "polutes" the PR somewhat because I need to change all files that implement `ArgumentHolder`. I guess it would make sense to have this in a separate PR?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -473,6 +474,11 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
             is Reference -> handleReference(currClass, node)
             is ConstructExpression -> handleConstructExpression(node)
             is CallExpression -> handleCallExpression(scopeManager.currentRecord, node)
+
+            // the following function are more for type-resolving, but they are placed inside the
+            // symbol resolver, because this allows us to have the necessary location of nodes where
+            // we infer from
+            is RecordDeclaration -> handleRecordDeclaration(node)
```
oxisto, line 481: I need to check, I think this is not necessary anymore, this should already be handled by the type resolver now.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -754,34 +760,29 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
-        possibleContainingTypes: Set<Type>,
-        bestGuess: Type?,
-        call: CallExpression
-    ): List<FunctionDeclaration> {
-        var records =
-            possibleContainingTypes.mapNotNull {
-                val root = it.root as? ObjectType
-                root?.recordDeclaration
-            }
+    fun handleRecordDeclaration(record: RecordDeclaration) {
```
oxisto, line 763: see above, I think this is not needed anymore

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -528,6 +544,52 @@ class Inference internal constructor(val start: Node, override val ctx: Translat
         this.scopeManager = ctx.scopeManager
         this.typeManager = ctx.typeManager
     }
+
+    fun inferReturnType(call: CallExpression?): Type {
```
oxisto, line 548: kdoc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -605,3 +667,197 @@ fun RecordDeclaration.inferMethod(
             call
         ) as? MethodDeclaration
 }
+
+fun TranslationContext.tryNamespaceInference(
```
oxisto, line 671: kdoc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -605,3 +667,197 @@ fun RecordDeclaration.inferMethod(
...
+    return (holder ?: scopeManager.globalScope?.astNode)
+        ?.startInference(this)
+        ?.inferNamespaceDeclaration(name, null, locationHint)
+}
+
+/**
+ * Tries to infer a [RecordDeclaration] from an unresolved [Type]. This will return `null`, if
+ * inference was not possible, or if it was turned off in the [InferenceConfiguration].
+ */
+fun TranslationContext.tryRecordInference(
```
oxisto, line 700: add a quick comment that we are also inferring the (nested) scope of the record, if it does not exist

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -605,3 +667,197 @@ fun RecordDeclaration.inferMethod(
...
+    }
+
+    var types = mutableListOf(type)
+    types.addAll(type.superTypes)
+
+    var methods = types.map { it.recordDeclaration }.flatMap { it.methods }
+    return methods.any { it.name.localName == name }
+}
+
+fun TranslationContext.tryParentScopeInference(
```
oxisto, line 850: kdoc

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXAmbiguitiesTest.kt
```diff
@@ -58,7 +58,7 @@ class CXXAmbiguitiesTest {
         assertNotNull(tu)
 
         // make sure we still have only one declaration in the file (the record)
-        assertEquals(1, tu.declarations.size)
+        assertEquals(3, tu.declarations.size)
```
oxisto, line 61: need to double-check why it is 3 now

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -528,6 +539,52 @@ class Inference internal constructor(val start: Node, override val ctx: Translat
...
+
+    fun inferReturnType(call: CallExpression?): Type {
+        if (call == null) {
+            return unknownType()
+        }
+
+        // Try to find out, if the supplied hint is part of an assignment.  If yes, we can use their
+        // type as the return type of the function
+        var targetType =
+            ctx.currentComponent.assignments.firstOrNull { it.value == call }?.target?.type
```
KuechA, line 551: `firstOrNull` is not really what I'd do here. Since most languages cannot differentiate between the methods to invoke based on the return type, I'd collect everything and use the most specific common supertype

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -528,6 +539,52 @@ class Inference internal constructor(val start: Node, override val ctx: Translat
...
+            ctx.currentComponent.allChildren<ArgumentHolder> { it.hasArgument(call) }.singleOrNull()
+        when (holder) {
+            is UnaryOperator -> {
+                // If it's a boolean operator, the return type is probably a boolean
+                if (holder.operatorCode == "!") {
+                    return call.language?.builtInTypes?.values?.firstOrNull { it is BooleanType }
+                        ?: unknownType()
+                }
+                // If it's a numeric operator, return the fist numeric type that we have
+                if (holder.operatorCode in listOf("+", "-", "++", "--")) {
```
KuechA, line 567: This is extremely language-specific.

At least for `+`, it could also be a string in a good amount of languages...

For languages with operator overloading, it could be anything.

maximiliankaul (approved):
I'm a bit undecided with this PR. I'm approving it, because it overall seems to be doing the job when looking at the tests. I like the positive examples demonstrated in the test cases. However, I'm not 100% sure on all the mechanics involved here. Maybe a short (in house) introduction to the inference logic would be beneficial...
I don't like all the tests being implemented for the CXX language. This inference logic is used for all languages, isn't it? I understand that there are plans of getting rid of the fluent dsl stuff for fronted-independent tests, but having everything only tested with CXX is a big negative point in my view.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/InferenceConfiguration.kt
konradweiss: I would agree, at least from a very uninformed perspective, to keep it disabled by default as it is described as very EXPERIMENTAL

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
konradweiss: If the idea is to implement an inference heuristic that "maaaay" still be ok here, hmm do we have a better alternative?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -237,7 +234,7 @@ class ScopeManager : ScopeProvider {
                     is Block -> BlockScope(nodeToScope)
                     is WhileStatement,
                     is DoStatement,
-                    is AssertStatement -> LoopScope(nodeToScope as Statement)
+                    is AssertStatement -> LoopScope(nodeToScope)
                     is ForStatement,
                     is ForEachStatement -> LoopScope(nodeToScope as Statement)
```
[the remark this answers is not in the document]
oxisto: Will be gone with https://github.com/Fraunhofer-AISEC/cpg/pull/1856

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -528,6 +539,52 @@ class Inference internal constructor(val start: Node, override val ctx: Translat
...
+        // type as the return type of the function
+        var targetType =
+            ctx.currentComponent.assignments.firstOrNull { it.value == call }?.target?.type
+        if (targetType != null && targetType !is UnknownType) {
+            return targetType
+        }
+
+        // Look for an "argument holder". These can be different kind of nodes
+        val holder =
+            ctx.currentComponent.allChildren<ArgumentHolder> { it.hasArgument(call) }.singleOrNull()
```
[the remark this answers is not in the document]
oxisto: I would say that we should keep the `ArgumentHolder` interface, but only implement it for nodes that have "real" arguments, such as operators and calls.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
oxisto: The `hint` can only be part of one assignment (if it is part of any) in the AST. Therefore `firstOrNull` is the correct thing to do here. To make it a bit more clearer, I will change it to `singleOrNull`.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -528,6 +539,52 @@ class Inference internal constructor(val start: Node, override val ctx: Translat
...
+                if (holder.operatorCode in listOf("+", "-", "++", "--")) {
+                    return call.language?.builtInTypes?.values?.firstOrNull { it is NumericType }
+                        ?: unknownType()
+                }
+            }
+            is ConstructExpression -> {
+                return holder.type
+            }
+            is BinaryOperator -> {
+                // If it is on the right side, it's probably the same as on the left-side (and
```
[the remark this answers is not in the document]
oxisto: Hm I wonder if we somehow could leverage `Language.propagateTypeOfBinaryOperation` here.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/PassHelper.kt
```diff
@@ -328,5 +383,50 @@ internal fun Pass<*>.tryMethodInference(
...
+private fun methodExists(
+    type: Type?,
+    name: String,
+): Boolean {
+    if (type == null) {
+        return false
+    }
+
+    var types = mutableListOf(type)
+    types.addAll(type.superTypes)
```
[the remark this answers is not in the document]
oxisto: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -80,7 +81,7 @@ class Inference internal constructor(val start: Node, override val ctx: Translat
         code: String?,
```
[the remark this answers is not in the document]
oxisto: Agree, but I didn't really change anything in there - expect calling the return type inference ;)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -111,10 +112,11 @@ class Inference internal constructor(val start: Node, override val ctx: Translat
             debugWithFileLocation(
                 hint,
                 log,
-                "Inferred a new {} declaration {} with parameter types {} in $it",
+                "Inferred a new {} declaration {} with parameter types {} and return types {} in $it",
```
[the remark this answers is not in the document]
oxisto: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
oxisto: > This is extremely language-specific.
> 
> At least for `+`, it could also be a string in a good amount of languages...
> 
> For languages with operator overloading, it could be anything.

Note that this is a *unary* operator with a numeric operation, so the assumption that it is any numeric type is probably ok. We could extend this heuristic of course to first look for integer types, with the highest bidwidth, to over-approximate and then if none are found, use a float type. Not sure if this is really better though.

As for operator overloading, yes of course, there will be cases where we are completely wrong, but on the other hand, there is no possibility at all to deduce the correct type. Return types are not taken into account when resolving, so the "worst" thing we can do is to propagate a wrong type into the `assignedTypes` of the data flows that follow and that is ok-ish.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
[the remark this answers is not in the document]
oxisto: > I'd prefer some logic here to catch the obvious cases `&&`, `||`, ... and then probably a new abstract function in Language, where each language has to define what happens for cases like `float + int`

This already exists in the form of `Language.propagateTypeOfBinaryOperation`, but it does not really help us here, because what we see is something like `Y = float + X`, where `Y` depends on the outcome of `Language.propagateTypeOfBinaryOperation`, but `X` is unknown. So in my opinion the best guess here is to assume that `X` is also a float.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/PassHelper.kt
```diff
@@ -294,23 +315,57 @@ internal fun Pass<*>.tryFunctionInference(
...
+    // inference.
+    // 1a) If the language does not even support functions at a global level, it's easy
+    // 1b) If this is a member call expression, it's also easy
+    var inferGlobalFunction =
+        if (call.language !is HasGlobalFunctions || call is MemberCallExpression) {
+            false
+        } else if (methodExists(bestGuess, call.name.localName)) {
+            // 2) We do a quick check, whether we would have a method with our name in the "best
+            // guess" class. Because if we do, we most likely ended up here because of an
+            // argument type mismatch. Once we use the new call resolution also for member
```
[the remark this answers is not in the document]
oxisto: Beyond the scope currently I would say.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -528,6 +539,52 @@ class Inference internal constructor(val start: Node, override val ctx: Translat
         this.scopeManager = ctx.scopeManager
         this.typeManager = ctx.typeManager
     }
+
+    fun inferReturnType(call: CallExpression?): Type {
```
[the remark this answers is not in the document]
oxisto: Removed

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/inference/Inference.kt
```diff
@@ -528,6 +539,52 @@ class Inference internal constructor(val start: Node, override val ctx: Translat
...
+        when (holder) {
+            is UnaryOperator -> {
+                // If it's a boolean operator, the return type is probably a boolean
+                if (holder.operatorCode == "!") {
+                    return call.language?.builtInTypes?.values?.firstOrNull { it is BooleanType }
+                        ?: unknownType()
+                }
+                // If it's a numeric operator, return the fist numeric type that we have
+                if (holder.operatorCode in listOf("+", "-", "++", "--")) {
+                    return call.language?.builtInTypes?.values?.firstOrNull { it is NumericType }
```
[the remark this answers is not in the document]
oxisto: there is no base, all we have is something like `+foo()`

PR #1589, opened 2024-06-14
Fixed Neo4J import of `ImportDeclaration`

oxisto:
The `ImportDeclaration` was trying to use the `Name` class without a converter, but unfortunately, we cannot add the prefix to the converter, so having three converters in the same file would override all properties.

Therefore, we are simply converting `alias` and `import` to a string.

### DIFF
--- a/cpg-neo4j/build.gradle.kts
+++ b/cpg-neo4j/build.gradle.kts
@@ -54,4 +54,6 @@ dependencies {
     // Command line interface support
     api(libs.picocli)
     annotationProcessor(libs.picocli.codegen)
+
+    testImplementation(testFixtures(projects.cpgCore))
 }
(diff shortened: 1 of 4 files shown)


cpg-neo4j/build.gradle.kts
```diff
@@ -54,4 +54,6 @@ dependencies {
     // Command line interface support
     api(libs.picocli)
     annotationProcessor(libs.picocli.codegen)
+
+    testImplementation(testFixtures(projects.cpgCore))
```
peckto, line 58: What is the effect of this change?

oxisto: That I can use the test utility classes in the neo4j tests. This could probably be moved into the common buildSrc convention because it makes sense for all projects.

peckto: Ok, I see. Fine for me.

peckto (approved):
Looks good to me and works as expected.

PR #1606, opened 2024-06-29
Added basic implementation of operator overloading

oxisto:
This PR adds basic functionality for resolving operator overloading for languages that support it (via the trait `HasOperatorOverloading` - which is currently python and C++). The respective frontend looks for function declarations that match entries in `overloadedOperatorNames` and creates an `OperatorDeclaration`.

The `SymbolResolver` is then looking for specific expressions (that implement `HasOverloadedOperation`) to see whether the expression is really an overloaded call to an operator declaration. In this case, the expression is **replaced** with an `OperatorCallExpression`, pointing to the respective  `OperatorDeclaration`. This happens only if a declaration was found, currently, no `OperatorDeclaration` is inferred. This might be something for the future.

Currently the following nodes support `HasOverloadedOperation` and should work:
- `BinaryOperator`: Tested in C++ and Python
- `UnaryOperator`: Tested in C++ and Python
- `MemberExpression`: Only really possible in C++, also tested in C++. This is basically the main trick behind every smart pointer framework by overriding the `->` operator.
- `CallExpression`: Not tested

There are probably more expressions, such as array indices, etc. that currently do not implement `HasOverloadedOperation` yet.

This PR depends on the following PRs which need to be resolved first:

- [x] #1605 
- [x] #1607 
- [x] #1608 
- [x] #1646
- [x] #1684

PR #1607, opened 2024-07-01
Added more basic interfaces

oxisto:
This PR adds more basic interfaces. It solves the issue that we currently either have a direct node in our utility functions (for example symbol resolving), e.g. a `CallExpression` or some very specific interfaces `HasBase`. But as we for example need a `name` for the `HasBase` interface/node, we need to do a very ugly `as Node` cast. 

Therefore, in this PR we are adding some more basic interfaces that "describe" nodes, e.g., whether they have a name, location (actually all of the nodes have that) and a mixture of that.

This is prep work for #1606, for example where we need to share a common interface between calls and operator expressions which are overloadable. Something which is currently not yet possible with the current interfaces.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Interfaces.kt
```diff
@@ -0,0 +1,141 @@
...
+    val scope: Scope?
+}
+
+/**
+ * This interface can be used to describe something similar to a "call", in a way that this node
+ * implements [HasBase], although [HasBase.base] is nullable and therefore optional and the node has
+ * some kind of [arguments]. The most prominent implementer of this interface is a [CallExpression],
+ * but other nodes might implement this as well, if suitable.
+ */
+interface HasArgumentsAndOptionalBase : HasLanguage, HasBase, HasNameAndLocation, HasScope {
```
KuechA, line 133: I don't really like this optional thing. Do we expect a base or not? I'd prefer to remove some null-checks instead of introducing more ;)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/CallExpression.kt
```diff
@@ -45,7 +45,8 @@ import org.neo4j.ogm.annotation.Relationship
  * An expression, which calls another function. It has a list of arguments (list of [Expression]s)
  * and is connected via the INVOKES edge to its [FunctionDeclaration].
  */
-open class CallExpression : Expression(), HasType.TypeObserver, ArgumentHolder {
+open class CallExpression :
+    Expression(), HasType.TypeObserver, HasArgumentsAndOptionalBase, ArgumentHolder {
```
KuechA, line 49: I don't know why the `CallExpression` should suddenly have a `base`. If I understood it correctly, this type of node is used for function calls, i.e., calls which * never* have a base. If a base is required, I thought we used the `MemberCallExpression`. Similarly, an operator doesn't make sense then

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
```diff
@@ -60,7 +60,7 @@ open class ControlDependenceGraphPass(ctx: TranslationContext) : EOGStarterPass(
     }
 
     /**
-     * Computes the CDG for the given [functionDeclaration]. It performs the following steps:
+     * Computes the CDG for the given [FunctionDeclaration]. It performs the following steps:
```
KuechA, line 63: Should probably be `[startNode]` (and the condition in line 76 will have to be updated to an EOG Starter without prevEOG, etc. in the future)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -98,7 +102,7 @@ open class Node : IVisitable<Node>, Persistable, LanguageProvider, ScopeProvider
     var comment: String? = null
 
     /** Location of the finding in source code. */
-    @Convert(LocationConverter::class) var location: PhysicalLocation? = null
+    @Convert(LocationConverter::class) override var location: PhysicalLocation? = null
```
KuechA, line 105: I'd move the doc to the interface.

Can we add the `@Convert` annotation to the interface? It doesn't make any difference right now but if we would use the interfaces for more things, it could make a difference

KuechA (commented):
I don't see how this solves the problem of casts between different interfaces. Shouldn't the "specific" interfaces extend the "basic" ones (e.g. `HasBase: HasLanguage, HasNameAndLocation`) to accomplish this?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Interfaces.kt
oxisto: Removed it and renamed it to `HasCallableArguments`

oxisto: alternatively just `IsCallable`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
oxisto: Moved the doc. Unfortunately, the convert annotation needs toe specified at the field and therefore at the implemented class.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Interfaces.kt
oxisto: I got rid of if completely

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/CallExpression.kt
```diff
@@ -87,12 +87,18 @@ open class CallExpression : Expression(), HasType.TypeObserver, ArgumentHolder {
...
+            return argumentEdges.map { it.end.type }
+        }
+
     /**
      * The expression that is being "called". This is currently not yet used in the [SymbolResolver]
      * but will be in the future. In most cases, this is a [Reference] and its [Reference.refersTo]
      * is intentionally left empty. It is not filled by the [SymbolResolver].
      */
-    @AST var callee: Expression? = null
+    @AST var callee: Expression = ProblemExpression("could not parse callee")
```
KuechA, line 101: Is it a problem of parsing or translation? I think, we should set the `ProblemNode.ProblemType` to something different (currently it's `TRANSLATION` by default). Might require another option in the enum

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/CallExpression.kt
oxisto: It should be set by the frontend so I guess `TRANSLATION` is ok.

PR #1608, opened 2024-07-01
Better integration of legacy call resolution with the new system and more cleanup

oxisto:
Lot of cleanups and removal of deprecated / legacy functions:

- Better integration of legacy call resolving by removing some duplicated code from the old call resolving functionality and instead providing a `legacyCandidates` set that can be supplied to `ScopeManager::resolveCall`
- Removed `ScopeManager::resolveFunctionLegacy`, as it was only used in one test and not used by the code base anymore
- Removed `SymbolResolver::resolveReferenceCallee` and `SymbolResolver::resolveMemberCallee` because (almost) all functionality was already implemented in `SymbolResolver::resolveCalleeByName` (which will stay for now)
- Removed `SymbolResolver::getInvocationCandidatesFromRecord` as it was only used by C++ and its functionality is also replaced by `SymbolResolver::resolveCalleeByName`
- Removed `ScopeManager::resolveFunctionStopScopeTraversalOnDefinition` and `SymbolResolver::shouldSearchForInvokesInParent`
- Removed the language trait `HasComplexCallResolution` and its associated function `refineMethodCallResolution` completely as it is not needed anymore because of better integration of the legacy call resolver (see above). It is still possible for a language to fine-tune call resolution decision by overriding `Language::bestViableResolution`. We might (re)-introduce an interface like this if the need arrives
- `ScopeManager::extractScope` now returns `null` in the pair, if no scope was found instead of the `startScope`. This makes it easier for callers to differentiate if the supplied name does not contain a scope
- Made `ScopeManager::resolve` internal, as it should have only been called by the scope manager / symbol resolver anyway. I replaced the few calls that the frontends made to it by `ScopeManager::getRecordByName`. In the future, we should also get rid of those calls
- Renamed `HasSuperClasses::handleSuperCall` to `HasSuperClasses::handleSuperExpression` because it was already capable of handling super calls and super field accesses. Remove all code that did extra handling for fields

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
@@ -981,14 +981,11 @@ class ScopeManager : ScopeProvider {
         // We need to differentiate between a qualified and unqualified lookup. We have a qualified
         // lookup, if the scope is not null. In this case we need to stay within the specified scope
         val list =
-            // TODO(oxisto): extractScope does not return null in all cases, so we need to make sure
-            //  that the returned scope is NOT our startscope
-            if (scope != null && scope is NameScope && scope != startScope) {
+            if (scope != null) {
                     scope.lookupSymbol(n.localName, thisScopeOnly = true, predicate = predicate)
-                }
-                // Otherwise, we can look up the symbol alone (without any FQN) starting from the
-                // startScope
-                else {
+                } else {
+                    // Otherwise, we can look up the symbol alone (without any FQN) starting from
+                    // the startScope
                     startScope?.lookupSymbol(n.localName, predicate = predicate)
                 }
                 ?.toMutableList() ?: return listOf()
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
@@ -35,6 +35,7 @@ import de.fraunhofer.aisec.cpg.graph.types.FunctionPointerType
 import de.fraunhofer.aisec.cpg.graph.types.IncompleteType
 import de.fraunhofer.aisec.cpg.graph.types.Type
 import de.fraunhofer.aisec.cpg.helpers.Util
+import de.fraunhofer.aisec.cpg.passes.SymbolResolver
 import de.fraunhofer.aisec.cpg.sarif.PhysicalLocation
 import java.util.*
 import java.util.function.Predicate
@@ -596,8 +597,6 @@ class ScopeManager : ScopeProvider {
      *
      * @param ref
      * @return
-     *
-     * TODO: We should merge this function with [.resolveFunction]
      */
     fun resolveReference(ref: Reference): ValueDeclaration? {
         val startScope = ref.scope
@@ -614,7 +613,10 @@ class ScopeManager : ScopeProvider {
             return pair.second
         }
 
-        val (scope, name) = extractScope(ref, startScope)
+        var (scope, name) = extractScope(ref, startScope)
+        if (scope == null) {
+            scope = startScope
+        }
 
         // Try to resolve value declarations according to our criteria
         val decl =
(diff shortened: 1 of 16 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -992,6 +826,42 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
                 ?.createInferredConstructor(constructExpression.signature)
     }
 
+    fun tryFunctionInference(
+        call: CallExpression,
+        result: CallResolutionResult,
+    ): List<FunctionDeclaration> {
+        // We need to see, whether we have any suitable base (e.g. a class) or not; but only if
```
KuechA, line 833: Does this comment actually reflect the code? "but only if the call itself is not already scoped" is not immediately visible to me here

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXAmbiguitiesTest.kt
```diff
@@ -57,8 +57,10 @@ class CXXAmbiguitiesTest {
             }
         assertNotNull(tu)
 
-        // make sure we still have only one declaration in the file (the record)
-        assertEquals(1, tu.declarations.size)
+        // we have 3 (record) declarations in our TU now. 1 of the original MyClass and two because
+        // CDT thinks that "call" is the return type and "crazy" the type of the parameter. We infer
+        // record declarations for all types, so we end up with 3 declarations here.
```
KuechA, line 62: And we haven't inferred the record declarations until now? Do we actually have those types in the final graph too? Shouldn't we replace these types since we don't really know?

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/CallResolverTest.kt
```diff
@@ -717,6 +717,7 @@ class CallResolverTest : BaseTest() {
                 true
             ) {
                 it.registerLanguage<CPPLanguage>()
+                it.inferenceConfiguration(InferenceConfiguration.builder().enabled(false).build())
```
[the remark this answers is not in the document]
oxisto: inference must be enable because otherwise the invokes edge wont be empty. Personally, I have no idea why it was empty before, I think that actually before the test or this functionality was broken and therefore the invokes edge was empty.

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/java/JavaLanguage.kt
```diff
@@ -108,11 +109,11 @@ open class JavaLanguage :
         } else super.propagateTypeOfBinaryOperation(operation)
     }
 
-    override fun handleSuperCall(
+    override fun handleSuperExpression(
         callee: MemberExpression,
```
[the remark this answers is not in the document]
oxisto: Yes

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/JavaCallResolverHelper.kt
```diff
@@ -42,64 +41,63 @@ class JavaCallResolverHelper {
...
+         * resolve the [MemberExpression].
          *
-         * @param callee The callee of the call expression that needs to be adjusted
+         * @param me The member expression that needs to be adjusted
          * @param curClass The class containing the call
          */
-        fun handleSuperCall(
-            callee: MemberExpression,
+        fun handleSuperExpression(
+            me: MemberExpression,
```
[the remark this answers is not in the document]
oxisto: Yeah sure.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -1084,9 +977,21 @@ class ScopeManager : ScopeProvider {
...
     ): List<Declaration> {
         val (scope, n) = extractScope(name, location, startScope)
+
+        // We need to differentiate between a qualified and unqualified lookup. We have a qualified
+        // lookup, if the scope is not null. In this case we need to stay within the specified scope
         val list =
-            scope?.lookupSymbol(n.localName, predicate = predicate)?.toMutableList()
-                ?: mutableListOf()
+            // TODO(oxisto): extractScope does not return null in all cases, so we need to make sure
+            //  that the returned scope is NOT our startscope
```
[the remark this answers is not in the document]
oxisto: I have to re-check if this is actually still a problem, because I am actually now returning null if the scope is not successfully extracted, so my first guess is that this behaviour is fixed.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -307,57 +322,15 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
-                // The base should have been resolved by now. Maybe we have some other clue about
-                // this base from the type system, so we can set the declaration accordingly.
-                if (base.refersTo == null) {
-                    base.refersTo = base.type.recordDeclaration
-                }
+        if (base is Reference) {
+            // The base should have been resolved by now. Maybe we have some other clue about
+            // this base from the type system, so we can set the declaration accordingly.
+            if (base.refersTo == null) {
+                base.refersTo = base.type.recordDeclaration
```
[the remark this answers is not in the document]
oxisto: I just moved things around and did not change this code. When I remove this, things start to break in the Java frontend, but I suspect that the java frontend does things wrong here and this is a crude way to fix it. I will add a TODO here.

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXAmbiguitiesTest.kt
oxisto: Yes exactly, we did not infer record declaration of types until now. I am not sure if there is anything more I can do at this point 🤷

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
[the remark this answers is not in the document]
oxisto: This should be solved, removing this comment

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
oxisto: It's the `!call.callee.name.isQualified()` part

oxisto: Ok you are correct, the comment is wrong or misleading. I will change the comment. This behaviour will anyway be redesigned as part of https://github.com/Fraunhofer-AISEC/cpg/pull/1586

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -531,224 +539,116 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
-                (!call.language.isCPP || shouldSearchForInvokesInParent(call))
-        ) {
+        if (invocationCandidates.isEmpty() && symbol.isNotEmpty()) {
             val records = possibleContainingTypes.mapNotNull { it.root.recordDeclaration }.toSet()
             invocationCandidates =
-                getInvocationCandidatesFromParents(localName, call, records).toMutableList()
+                getInvocationCandidatesFromParents(symbol, records).toMutableSet()
         }
 
         // Add overridden invokes
```
konradweiss, line 646: Rename `invocationCandidates`

konradweiss (changes requested):
One small naming change

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
konradweiss: Ah this was mapped to the wrong lines

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -531,224 +539,116 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
         invocationCandidates.addAll(
```
konradweiss, line 647: Here we should rename `invocationCadidates`

konradweiss (approved):

PR #1618, opened 2024-07-15
Clean up old call to byName

peckto:


oxisto (approved):

cpg-console/src/main/kotlin/de/fraunhofer/aisec/cpg/console/CompilationDatabase.kt
```diff
@@ -59,7 +59,6 @@ class CompilationDatabase : Plugin {
                     "import de.fraunhofer.aisec.cpg.graph.allChildren",
                     "import de.fraunhofer.aisec.cpg.graph.ast",
```
oxisto, line 60: I think we need to include `import de.fraunhofer.aisec.cpg.graph.*` here I think  and it looks like something is still using `byName`.

peckto: Its already there:
https://github.com/Fraunhofer-AISEC/cpg/pull/1618/files#diff-2ceb1018c52a5beec418b3ab986e10abe7a5ebda5469c0900af841e9b0ef050aL53

peckto: Does `byName` still exist? I cant find the symbol anymore.

oxisto: Not sure, but the test fails now :D

oxisto: Ah! you removed one line and the test checks the number of lines. So we need to adjust the assert from 19 to 18.

peckto: ah, should be fixed now

PR #1621, opened 2024-07-18
Converting LLVM frontend to`withChildren` style

oxisto:


KuechA:
I guess this PR is obsoleted by #1646 and #1642 and should be closed?

PR #1625, opened 2024-07-22
Python operators

maximiliankaul:
This PR implements the new operator overloading functionality for the Python language.

### DIFF
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
@@ -96,6 +96,9 @@ class PythonLanguage :
                 "+" to "__pos__", // TODO __add__
                 // "ads()" to "__abs__",
                 "~" to "__invert__",
+                "()" to
+                    "__call__", // ... x(arg1, arg2, ...) roughly translates to type(x).__call__(x,
+                // arg1, ...)
                 // TODO walrus operator (x := 42) not included in documentation
             )
 
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
@@ -25,16 +25,19 @@
  */
 package de.fraunhofer.aisec.cpg.frontends.python
 
+import de.fraunhofer.aisec.cpg.frontends.HasOperatorOverloading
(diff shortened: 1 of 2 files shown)


maximiliankaul:
There are still some problems to discuss:
- some operators like `+` can map to different functions depending on them being used as a unary `__pos__` or binary `__add__` operator
- do we want to include "operators" like `abs()`?
- the documentation say, that `[]` or `()` "roughly" translates to ... -> I'd still hard code them as an exact match

oxisto:
> There are still some problems to discuss:
> 
> * some operators like `+` can map to different functions depending on them being used as a unary `__pos__` or binary `__add__` operator

Unfortunately, I have no good idea how to handle this :( The only "correct" way to do that would be to use a Pair of an class and operator code as a key, something like
```kotlin
Pair(BinaryOperator::class, "+") to "__add__",
Pair(UnaryOperator::class, "+") to "__pos__",
```

which is not too bad. If that makes sense I can prepare this in the base PR

> * do we want to include "operators" like `abs()`?

No

> * the documentation say, that `[]` or `()` "roughly" translates to ... -> I'd still hard code them as an exact match

Yes, that's ok

maximiliankaul:
> > There are still some problems to discuss:
> > 
> > * some operators like `+` can map to different functions depending on them being used as a unary `__pos__` or binary `__add__` operator
> 
> Unfortunately, I have no good idea how to handle this :( The only "correct" way to do that would be to use a Pair of an class and operator code as a key, something like
> 
> ```kotlin
> Pair(BinaryOperator::class, "+") to "__add__",
> Pair(UnaryOperator::class, "+") to "__pos__",
> ```
> 
> which is not too bad. If that makes sense I can prepare this in the base PR

I'm in favor of this approach.

> > * do we want to include "operators" like `abs()`?
> 
> No

I'll remove the comments.
 
> > * the documentation say, that `[]` or `()` "roughly" translates to ... -> I'd still hard code them as an exact match
> 
> Yes, that's ok

:+1:

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
```diff
@@ -49,6 +55,54 @@ class PythonLanguage : Language<PythonLanguageFrontend>(), HasShortCircuitOperat
     override val compoundAssignmentOperators =
         setOf("+=", "-=", "*=", "**=", "/=", "//=", "%=", "<<=", ">>=", "&=", "|=", "^=", "@=")
 
+    // https://docs.python.org/3/reference/datamodel.html#special-method-names
+    override val overloadedOperatorNames:
```
oxisto, line 59: Need to add `@Transient`

oxisto (approved):

PR #1626, opened 2024-07-22
Declarative cli tool to make use of the query capabilities

MariusAlbrecht:
# Declarative CLI Tool to Make Use of the Query Capabilities

This pull request introduces a CLI to use Queries outside of the interactive console in a declarative way.
Should work similiar to regular security checkers, insofar it should be able to just be run on a codebase and report something. To do this predefined `Rule`s are neccesary.
Should also be extensible / customizable. To do this the `Rule` and `Reporter` interface can be manually implemented to add new rules and output formats.

## Usage

Right now the code lives in cpg-analysis (which I acknowledge should probably change).
To actually use 
1. add a Rule to the `RuleRunner` class (see below for an example)
2. do :installDist and run the binary which is then located at `cpg/cpg-analysis/build/install/cpg-analysis/bin/cpg-analysis`

## Notes

This is still absolutely WIP, just wanted to get this mostly working thing in a PR before being away for holiday. There's a number of actual "TODO" comments in the code, additionally the following should probably be addressed as well
- logging
- tests
- actual Rules
- at least in sarif, dynamic messages (e.g. adding variable names into a predefined message) are possible. The `Rule` interface supports this, it's up to the Rule to supply the arguments. This could be explored to improve ux

I plan to adddress all of this once I'm back in 3 weeks.

---

Rule exmaple:

```Kotlin
/*
   <copyright> 
   package
*/

import de.fraunhofer.aisec.cpg.TranslationResult
import de.fraunhofer.aisec.cpg.graph.statements.expressions.CallExpression
import de.fraunhofer.aisec.cpg.query.*

class BufferOverreadMemcpy : Rule {
    override var queryResult: QueryTree<*>? = null // set by run() method of Rule interface
    override val id = "cpg-0000"
    override val name = "memcpy src smaller than size"
    override val shortDescription =
        "This rule detects memcpy calls where the size of the source is smaller than the size argument, which can " +
            "overread the src buffer"
    override val cweId: String = "787"
    override val level = Rule.Level.Error
    override val message = "memcpy call with source size smaller than size argument detected"

    override fun run(result: TranslationResult) {
        queryResult =
            result.allExtended<CallExpression>(
                { it.name.localName == "memcpy" },
                //          src                     n
                { sizeof(it.arguments[1]) ge min(it.arguments[2]) }
            )
    }
}

```

note that this only gives pretty output and evaluation steps when the used `sizeof` and `min` functions are adjusted to include nodes in their children. Just do `mutableListOf(QueryTree(n))` instead of `mutableListOf()` in these 2 locations:

https://github.com/Fraunhofer-AISEC/cpg/blob/bf8b37cbc25361b3517b9c454388c4f2f2bbb137/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/Query.kt#L123
https://github.com/Fraunhofer-AISEC/cpg/blob/bf8b37cbc25361b3517b9c454388c4f2f2bbb137/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/Query.kt#L137


KuechA:
* [ ] @oxisto checks if we can run scripts from the `cpg-console`
* [ ] Extract rules/queries to a kotlin script
* [ ] The `SarifReporter` and `Rule` can be kept in the `cpg-analysis` module but the CLI/application functionality should be moved to the `cpg-console`
* [ ] Run these scripts from the `cpg-console`

PR #1630, opened 2024-07-23
Added a `withParent` helper function

oxisto:
Slightly inspired by discussion with @konradweiss. This adds a helper function to set the ast parent in cases where the node cannot be created within the `withChildren` block.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
@@ -420,6 +420,31 @@ fun <T : Node> T.withChildren(
     return this
 }
 
+/**
+ * This function can be used to set the [Node.astParent] of this node to the current node on the
+ * [AstStackProvider]'s stack. This is particularly useful if the node was created outside of the
+ * [withChildren] lambda. This is a usual pattern if the node is optionally wrapped in something
+ * else.
+ *
+ * Example:
+ * ```kotlin
+ * val expr = newReference("p")
+ *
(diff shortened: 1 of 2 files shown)


maximiliankaul (approved):
I'm neutral on this one. If it helps your use case, then I have no reservations merging it.
Some remarks:
- `.withParent()` could be a bit misleading. Which parent? I'd favor `.withParent(parentNode)` or renaming this to `.withLatestParent()`
- `node.withParent()` vs `node.astParent = parentNode` do we gain anything with this new `withParent()` (besides people not using `astParent` directly)?

oxisto:
> I'm neutral on this one. If it helps your use case, then I have no reservations merging it. Some remarks:
> 
> * `.withParent()` could be a bit misleading. Which parent? I'd favor `.withParent(parentNode)` or renaming this to `.withLatestParent()`

I already thought of this aswell. `withCurrentParent()`? `withParent` has the advantage of being very short :)

> * `node.withParent()` vs `node.astParent = parentNode` do we gain anything with this new `withParent()` (besides people not using `astParent` directly)?

Yes because you can directly use the `withParent` expression as mentioned in the example
```kotlin
it.input = expr.withParent()``
```

vs.

```kotlin
it.input = expr
expr.astParent = it
```

maximiliankaul:
> > I'm neutral on this one. If it helps your use case, then I have no reservations merging it. Some remarks:
> > 
> > * `.withParent()` could be a bit misleading. Which parent? I'd favor `.withParent(parentNode)` or renaming this to `.withLatestParent()`
> 
> I already thought of this aswell. `withCurrentParent()`? `withParent` has the advantage of being very short :)

I'm ok with any of the options.

PR #1636, opened 2024-07-26
Workaround to use `astParent` in DFG passes

oxisto:
This workaround removes the DFG/EOG nodes from the `EdgeCachPass`, which will now only cache AST nodes. In the future we want to get rid of this pass alltogether.


oxisto:
Obsolete once we merge in https://github.com/Fraunhofer-AISEC/cpg/pull/1646

PR #1642, opened 2024-08-01
Improved property edges

oxisto:
## TL;DR

This PR adds major improvements to property edges. We got rid of the `Properties` map and replaced them with individual Kotlin properties. Furthermore lists of edges are now dedicated container classes with magic.

## Introduction of new specific edge classes

Next to the `Dataflow` edge class, which already exists, we added new edge classes. The `PropertyEdge` class itself is now abstract and thus no "generic" property edges (without any properties) can be created. There were a few uses of these kind of edges, this PR reverts them back to regular node lists (until we decide that this edge actually needs properties).

### `Usage`

This edge class denotes the usage of a declaration in a reference.

### `Invoke`

Invocation of a function declaration by a call expression.

### `AstEdge`

Part of the AST.

#### `TemplateArgument`

AST edge for template invocation arguments.

### `EvaluationOrder`

Part of the EOG.

### `ControlDependence`

Part of the CDG.

## Conversion of properties in `Properties` map to individual properties

This is basically the consequence of the above:
* `INDEX`: `index` to `Edge`
* `NAME`: `name` to `Edge`
* `DEPENDENCE`: `dependence` to `Edge`
* `UNREACHABLE`: `unreachable` to `EvaluationOrder`
* `BRANCH`: `branch` to `EvaluationOrder`
* `BRANCH`: `branches` to `ControlDependence`
* `INSTANTIATION`: `instantiation` to `TemplateArgument`

## Simplified creation of edges

Previously we had a lot of `addXXX` wrapper functions around property edges that took care of things like "mirroring" prev/next as well as inserting indices. This is all now done automagically by appropriate containers of property edges (and under the `PropertyEdgeCollection` interface under the hood.

This makes it easy to add new edges to the collection based on "target" nodes and it also supports setting appropriate properties using a builder pattern.

```Kotlin
node1.nextEOGEdges.add(node2) {
  branch = false
  unreachable = true
}
```

For all "mirrored" properties, this will also add the edge to the corresponding mirror property (next vs. prev). In order to use this feature the mirror property needs to be specified.

```Kotlin
class Node /* ... */ {
  var nextDFGEdges: Dataflows<Node> = Dataflows<Node>(this, Node::prevDFGEdges, outgoing = true)
  var prevDFGEdges: Dataflows<Node> = Dataflows<Node>(this, Node::nextDFGEdges, outgoing = false)
}
```

```Kotlin
node1.nextDFGEdges.add(node2)
// node2 will also now have node1 in its prevDFGEdges
```
## Open Issues

- [x] Use mirror collection in EOG nodes
- [x] Remove all remaining add/remove functions
- [x] Implement preRemove hook for removal functions
- [x] Final cleanups
- [x] Rename `PropertyEdge` to `Edge` to be more consistent with `Node`?
- [x] Decide on the type returned by `unwrapping` (`List` vs. `MutableList` vs `UnwrappedEdgeList`)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/SubgraphWalker.kt
```diff
@@ -164,19 +162,22 @@ object SubgraphWalker {
                         (field.getAnnotation(Relationship::class.java).direction ==
                             Relationship.Direction.OUTGOING)
                 }
-                if (checkForPropertyEdge(field, obj) && obj is Collection<*>) {
+                /*if (checkForPropertyEdge(field, obj) && obj is Collection<*>) {
```
KuechA, line 165: Remove completely

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
```diff
@@ -103,17 +105,16 @@ object Util {
                 val border = SubgraphWalker.getEOGPathEdges(n)
                 if (en == Edge.ENTRIES) {
                     val pe = border.entries.flatMap { it.prevEOGEdges }
-                    if (Quantifier.ALL == q && pe.any { !it.containsProperties(props) })
-                        return false
-                    pe.filter { it.containsProperties(props) }.map { it.start }
+                    if (Quantifier.ALL == q && pe.any { false && it.branch != branch }) return false
+                    pe.filter { true || it.branch == branch }.map { it.start }
```
KuechA, line 109: `true ||` (always true) makes the filter useless. Same for `false &&` (always false) and the other occurrences below. What's the purpose of all this? Is this function useless/not needed?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
```diff
@@ -183,30 +182,32 @@ open class ControlDependenceGraphPass(ctx: TranslationContext) : EOGStarterPass(
...
-                            // The if statement has only a then branch but there's a way to "jump
-                            // out" of this branch. In this case, we want to set the false property
-                            // here.
-                            properties[Properties.BRANCH] = setOf(false)
-                        }
+                    node.prevCDGEdges.add(k) {
+                        branches =
+                            when {
+                                branchesSet.size == 1 -> {
+                                    branchesSet
```
KuechA, line 195: This branch and `branchesSet.isNotEmpty()` can be merged.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/InitializerListExpression.kt
```diff
@@ -47,12 +47,7 @@ class InitializerListExpression : Expression(), ArgumentHolder, HasType.TypeObse
     /** The list of initializers. */
     @Relationship(value = "INITIALIZERS", direction = Relationship.Direction.OUTGOING)
     @AST
-    var initializerEdges = mutableListOf<PropertyEdge<Expression>>()
-        set(value) {
-            field.forEach { it.end.unregisterTypeObserver(this) }
-            field = value
-            value.forEach { it.end.registerTypeObserver(this) }
-        }
+    var initializerEdges = astEdgesOf<Expression>(postAdd = { it.end.registerTypeObserver(this) })
```
KuechA, line 50: And unregistering the TypeObserver happens automatically?

oxisto: No, still pending. The "remove" hook is not implemented yet.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
```diff
@@ -126,13 +126,13 @@ private fun handleIfStatement(
     val (unreachableEdge, remainingEdges) =
         if (evalResult is Boolean && evalResult == true) {
             Pair(
-                n.nextEOGEdges.firstOrNull { e -> e.getProperty(Properties.INDEX) == 1 },
-                n.nextEOGEdges.filter { e -> e.getProperty(Properties.INDEX) != 1 }
+                n.nextEOGEdges.firstOrNull { e -> e.index == 1 },
```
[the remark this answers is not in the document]
oxisto: isn't this encoded in the `branch` property?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -113,13 +119,15 @@ open class Node :
...
-    var prevEOGEdges: MutableList<PropertyEdge<Node>> = ArrayList()
+    @JsonIgnore
+    var prevEOGEdges = EvaluationOrders<Node>(this, outgoing = false)
         protected set
 
     /** Outgoing control flow edges. */
     @Relationship(value = "EOG", direction = Relationship.Direction.OUTGOING)
     @PopulatedByPass(EvaluationOrderGraphPass::class)
-    var nextEOGEdges: MutableList<PropertyEdge<Node>> = ArrayList()
+    @JsonManagedReference
```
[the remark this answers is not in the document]
oxisto: This can mostly be removed again I think, this was early attempts to silence neo4j errors.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -250,54 +264,25 @@ open class Node :
...
-            }
-        nextDFGEdges.add(edge)
-        next.prevDFGEdges.add(edge)
-    }
-
     fun removeNextDFG(next: Node?) {
         if (next != null) {
-            val thisRemove =
-                PropertyEdge.findPropertyEdgesByPredicate(nextDFGEdges) { it.end === next }
+            val thisRemove = nextDFGEdges.filter { it.end === next }
```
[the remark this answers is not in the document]
oxisto: Yes, this is still pending, see open issues.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -250,54 +264,25 @@ open class Node :
...
-    fun addNextEOG(propertyEdge: PropertyEdge<Node>) {
-        nextEOGEdges.add(propertyEdge)
-    }
-
-    fun clearNextEOG() {
-        nextEOGEdges.clear()
-    }
-
-    /** Adds a [Dataflow] edge from this node to [next], with the given [Granularity]. */
-    fun addNextDFG(
```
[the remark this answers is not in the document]
oxisto: Still pending, after the cleanup all of them will be removed

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edge/PropertyEdge.kt
```diff
@@ -54,271 +47,76 @@ import org.slf4j.LoggerFactory
...
-                }
-            }
-            return false
+    /**
+     * This special hash-code function takes into account whether the caller put this edge into an
+     * outgoing or incoming edge. This is primarily used by [PropertyEdgeList.hashCode].
+     */
+    fun directionalHashCode(outgoing: Boolean = true): Int {
+        return if (outgoing) {
+            Objects.hash(end)
```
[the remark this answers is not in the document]
oxisto: We need to exclude the node where we "start" because otherwise we would go into the node when we calculate the hashcode, this would then include the hashcode of the property edge list, which will then loop.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edge/flows/ControlDependence.kt
```diff
@@ -0,0 +1,80 @@
...
+ */
+@RelationshipEntity
+class ControlDependence(
+    start: Node,
+    end: Node,
+    /** A set of [EvaluationOrder.branch] values. */
+    var branches: Set<Boolean> = setOf(),
+) : PropertyEdge<Node>(start, end), ProgramDependence {
+    /** All control dependence edges exercise control dependence. */
+    override var dependence: DependenceType? = DependenceType.CONTROL
```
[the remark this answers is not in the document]
oxisto: the parent `dependence` type needs to be nullable, because not all DFG edges end up in the PDG.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edge/flows/ProgramDependence.kt
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * This edge can represent some kind of dependence (see [dependence]). This is special in a way that
+ * this is not an edge class, but only a interface with two known implementors: [Dataflow] and
+ * [ControlDependence].
+ */
+sealed interface ProgramDependence {
+
+    /**
+     * The type of dependence (e.g. control or data or none). This field is intentionally nullable,
+     * because not all [Dataflow] edges are selected. This selection is performed in the
```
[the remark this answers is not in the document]
oxisto: This is a bit tricky. The PDG does not have a distinct type, but it is an interface of all nodes which implement `dependence`. All CDG nodes implement dependence, but for the DFG only certain nodes are selected. We cannot partially implement an interface, therefore `dependence` is nullable. I am open for suggestions for a better approach.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edge/PropertyEdge.kt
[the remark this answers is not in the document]
KuechA: But we do use `start` in `hashCode`, isn't this the same problem then?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edge/PropertyEdge.kt
```diff
@@ -54,271 +47,76 @@ import org.slf4j.LoggerFactory
...
-     * Checks if the properties of the edge contain the given properties with the specified values.
-     */
-    fun containsProperties(props: Map<Properties, Any?>): Boolean {
-        return properties.entries.containsAll(props.entries)
+        return true
     }
 
     override fun hashCode(): Int {
-        return Objects.hash(end, properties)
+        return Objects.hash(start, end)
```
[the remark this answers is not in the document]
oxisto: Fixed now

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edge/PropertyEdge.kt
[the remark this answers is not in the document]
oxisto: Moved the function outside of property edge because it is only used in edge lists to avoid loops.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
oxisto: I have no idea how these functions work but they test only seem to parse if I do it like this :)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edge/Extensions.kt
```diff
@@ -0,0 +1,60 @@
...
+package de.fraunhofer.aisec.cpg.graph.edge
+
+import de.fraunhofer.aisec.cpg.graph.Node
+
+/**
+ * If the node list is backed by a [PropertyEdgeList] container (such as
+ * [de.fraunhofer.aisec.cpg.graph.edge.ast.AstEdges]), this returns the first [PropertyEdge] that
+ * fulfills the [predicate].
+ */
+fun <T : Node> List<T>.edge(predicate: (PropertyEdge<T>) -> Boolean): PropertyEdge<T>? {
```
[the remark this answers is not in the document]
oxisto: I removed these extensions for now.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
KuechA: @konradweiss any ideas what this function is doing? Do we need it?

In the current form, you can simplify several things here:
* `if (Quantifier.ALL == q && pe.any { false && it.branch != branch }) return false` can be removed since the condition is always `false`. Same holds for the other conditions where `false` is returned (l. 106, 115, 128, 137)
* `pe.filter { true || it.branch == branch }.map { it.start }` can be written as `pe.map { it.start }` (probably just `border.entries.flatMap{it.prevEOG}` in l.105?). Similar observations hold for the other filters (l. 117, 130, 139)

While this doesn't explain anything, it would simplify the method and probably not check anything in the end ;)

konradweiss: Yes, I know what it is doing, yes we need it ;) but thats the short answer. It is used in the EOG tests so check if a node (or its subtree) connects to another node(or its subtree) with some(or all) of the outgoing EOG edges. The filters are there to ensure that the need properties are set on the compared edges. In the Tests we probably use only the branch property but the changes in this PR do limit the functionality of the function: allowing to specify any property on the edge to be true for them to match in the test. Why there now is a `true` and `false` in the condition is new to me. I would guess the properties are not set properly on the edges and therefore instead of fixing the missing property the filters here were changed to not fail when they are missing.

konradweiss: @oxisto I will look into the reasons why the tests were failing when considering the branch property on the edge.

oxisto: I am trying to replicate the old behaviour with a predicate but still 1 tests tails.. Unfortunately this function is quite complex and not easily comprehensible :(

oxisto: Works now.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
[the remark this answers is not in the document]
oxisto: I tried using the `branch` property and the tests seem to work 🤷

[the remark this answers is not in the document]
KuechA: Well, that's great and much more understandable than this :D

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/collections/EdgeList.kt
```diff
@@ -0,0 +1,137 @@
...
+        edges.forEach { handlePostRemove(it) }
+    }
+
+    override fun add(index: Int, element: EdgeType) {
+        // Make sure, the index is always set
+        element.index = this.size
+
+        super<ArrayList>.add(index, element)
+
+        // TODO: actually we need to re-compute all other index properties
```
konradweiss, line 101: Why, isn't this append, so all other indices stay the same?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/EvaluationOrder.kt
```diff
@@ -0,0 +1,85 @@
...
+
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.edges.Edge
+import de.fraunhofer.aisec.cpg.graph.edges.collections.EdgeList
+import de.fraunhofer.aisec.cpg.graph.edges.collections.MirroredEdgeCollection
+import kotlin.reflect.KProperty
+import org.neo4j.ogm.annotation.RelationshipEntity
+
+@RelationshipEntity
+class EvaluationOrder(
```
konradweiss, line 36: I got the Idea here, but it applies to most of the edges, e.g. `ControlDependence`. When someone looks at this Class he might not know what we mean with evaluation order. What that semantic means should be contained in the documentation of the Pass that adds the edge or the member in the node class that holds these edges. Do we want to duplicate that comment here, or add a shorter description with a reference where to find more extensive documentation?

konradweiss (changes requested):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
```diff
@@ -103,17 +106,17 @@ object Util {
                 val border = SubgraphWalker.getEOGPathEdges(n)
                 if (en == Edge.ENTRIES) {
                     val pe = border.entries.flatMap { it.prevEOGEdges }
-                    if (Quantifier.ALL == q && pe.any { !it.containsProperties(props) })
+                    if (Quantifier.ALL == q && pe.any { predicate?.invoke(it) == false })
                         return false
-                    pe.filter { it.containsProperties(props) }.map { it.start }
+                    pe.filter { predicate?.invoke(it) != false }.map { it.start }
```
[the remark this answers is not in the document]
oxisto: It's not possible, because predicate can be nullable and you can only invoke `!` on non-nullable objects.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/collections/EdgeList.kt
oxisto: No, it's added at this particular index. All items before the index stay the same, everthing behind that needs to be shifted + 1

konradweiss: ok

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/ExpressionBuilder.kt
```diff
@@ -567,18 +565,26 @@ fun <T> Literal<T>.duplicate(implicit: Boolean): Literal<T> {
...
+            duplicate.prevDFGEdges.addContextSensitive(
+                next.start,
+                next.granularity,
+                next.callingContext
+            )
+        } else {
+            duplicate.prevDFGEdges += next
+        }
     }
     // TODO: This loses the properties of the edges.
```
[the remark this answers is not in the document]
oxisto: I added a new loop that keeps the properties

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/collections/EdgeList.kt
oxisto: implemented

konradweiss (approved):

PR #1647, opened 2024-08-07
Implementing `astParent` for single nodes using a setter

oxisto:
An alternative implementation. Alternative to #1646


maximiliankaul (changes requested):
No. Don't do this :( Let's go with #1646

oxisto:
Closed per @maximiliankaul's request

PR #1663, opened 2024-08-27
Replace `TODO()` with `ProblemDeclaration` nodes

maximiliankaul:
- Change to `Node`: add `addtionalProblems` field.
- Replace `TODO`s with `ProblemDeclaration`s.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Problem.kt
```diff
@@ -0,0 +1,32 @@
...
+ *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph
+
+/** A problem node documenting that something failed. */
+data class Problem(
```
maximiliankaul, line 29: @oxisto: @KuechA questions the need for the new class. I created it, because we cannot instantiate a `ProblemNode` as it is an interface. What's your preferred solution?

oxisto: Can’t we just use the existing problem expression and declarations? I would assume that we can categorize our problems into the two existing classes. Having another „Problem“ class sounds confusing to me.

oxisto: Having the additional problems property is a good idea though.

maximiliankaul: Ok, so we would then have a `ProblemExpression` if there's a problem with a statement? Works for me - I just want to make sure that I understand you correctly.

oxisto: Since expression derives from statement, yes.

maximiliankaul: https://github.com/Fraunhofer-AISEC/cpg/pull/1663/commits/af5e3d0f3c9fb26ae95b25ae91d92e22670468f0

oxisto (approved):
Looks good to me. Since this changes node properties... Maybe a second opinion @konradweiss @KuechA

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -257,6 +257,12 @@ abstract class Node :
     @Relationship("ANNOTATIONS") var annotationEdges = astEdgesOf<Annotation>()
     var annotations by unwrapping(Node::annotationEdges)
 
+    /**
+     * Additional problem nodes. These nodes represent problems which occurred during processing of
+     * a node (i.e. only partially processed).
+     */
+    val additionalProblems: MutableSet<ProblemNode> = mutableSetOf()
```
oxisto, line 264: I wonder if it would make sense to make this an AST edge. But in this case we need to either use `ProblemDeclaration` or `ProblemExpression` since our AST edges only work with a real `Node` and `ProblemNode` is an interface (although misleadingly named).

oxisto: The advantage of making this an AST edge would be that we could gather all problems using `allChildren` at some point.

KuechA:
> Looks good to me. Since this changes node properties... Maybe a second opinion @konradweiss @KuechA

I'm also fine with the changes but we should fix the comment above

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
oxisto: > But it's not really part of the AST, so I wouldn't use this hack. At some point, we'll just remove it and we'll be surprised why things are missing ;) But you have a point: There there's a shortcut to collect problems (https://github.com/Fraunhofer-AISEC/cpg/blob/main/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt#L608), and I guess we have to add this set to the result there. I'm wondering if we have to do this recursively though...

We probably can't, since it uses `allChildren` and that only uses AST edges.

maximiliankaul: > > But it's not really part of the AST, so I wouldn't use this hack. At some point, we'll just remove it and we'll be surprised why things are missing ;) But you have a point: There there's a shortcut to collect problems (https://github.com/Fraunhofer-AISEC/cpg/blob/main/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt#L608), and I guess we have to add this set to the result there. I'm wondering if we have to do this recursively though...
> 
> We probably can't, since it uses `allChildren` and that only uses AST edges.

Still, I'm with @KuechA on this. It's a stretch to say theses nodes are part of the AST. If we have them as AST nodes now, then we should at least try to come up with a cleaner solution asap.

oxisto: After internal discussion, we keep those problems out of the AST, but regards this as a *temporary* solution and have a look at a cleaner design that encompasses all problems (AST translation, parsing, passes).

PR #1669, opened 2024-09-03
Handle arguments in Python frontend

### LINKED ISSUE #1653: Python frontend: better argument handling
The Python frontend is missing some functionality when handling arguments:
- [x] `vararg` https://github.com/Fraunhofer-AISEC/cpg/blob/06995d44afb21212c72fe18320324d76d562575a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt#L438
- [ ] `kwonlyargs` https://github.com/Fraunhofer-AISEC/cpg/blob/06995d44afb21212c72fe18320324d76d562575a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt#L448
- [ ] `kw_defaults` https://github.com/Fraunhofer-AISEC/cpg/blob/06995d44afb21212c72fe18320324d76d562575a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt#L458
- [ ] `kwarg` https://github.com/Fraunhofer-AISEC/cpg/blob/06995d44afb21212c72fe18320324d76d562575a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt#L468
- [ ] `defaults` https://github.com/Fraunhofer-AISEC/cpg/blob/06995d44afb21212c72fe18320324d76d562575a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt#L478

lshala:
This partially fixes #1653


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -394,95 +395,102 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+        val recvPythonNode = positionalArguments.firstOrNull()
+        if (recvPythonNode == null) {
             val problem =
                 newProblemDeclaration(
-                    "`kw_defaults` are not yet supported",
+                    "Expected a receiver",
                     problemType = ProblemNode.ProblemType.TRANSLATION,
                     rawNode = args
                 )
             frontend.scopeManager.addDeclaration(problem)
```
maximiliankaul, line 428: Change to `result.additionalProblems += ...` once https://github.com/Fraunhofer-AISEC/cpg/pull/1663/ is merged.

oxisto:
Can you please merge/rebase main, thanks!

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -394,95 +395,102 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+        } else {
+            val tpe = recordDeclaration.toType()
+            val defaultValue =
+                args.defaults.getOrNull(0)?.let { frontend.expressionHandler.handle(it) }
```
maximiliankaul, line 432: I don't think this is correct: assume there are some default values, then this will always get a default value and you'll create an error a few lines below -> this needs some fixing.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -394,95 +395,102 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+                        problemType = ProblemNode.ProblemType.TRANSLATION,
+                        rawNode = args
+                    )
+                frontend.scopeManager.addDeclaration(problem)
+            }
+            frontend.scopeManager.addDeclaration(recvNode)
+            when (result) {
+                is ConstructorDeclaration -> result.receiver = recvNode
+                is MethodDeclaration -> result.receiver = recvNode
+                else -> TODO()
```
maximiliankaul, line 453: Please also update to a Problem node, once the other PR is in main.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -394,95 +395,102 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
-            val problem =
-                newProblemDeclaration(
-                    "`vararg` is not yet supported",
-                    problemType = ProblemNode.ProblemType.TRANSLATION,
-                    rawNode = it
-                )
-            frontend.scopeManager.addDeclaration(problem)
-        }
+        args.vararg?.let { handleArgument(it, isPosOnly = false, isVariadic = true) }
+        args.kwarg?.let { handleArgument(it, isPosOnly = false, isVariadic = true) }
```
maximiliankaul, line 408: is a keyword arg really variadic?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -394,95 +395,102 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
-                    problemType = ProblemNode.ProblemType.TRANSLATION,
-                    rawNode = args
-                )
-            frontend.scopeManager.addDeclaration(problem)
-        }
+        handleKeywordOnlyArguments(args)
+    }
 
-        if (args.kw_defaults.isNotEmpty()) {
+    private fun handleReceiverArgument(
```
maximiliankaul, line 413: doc string

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -394,95 +395,102 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
 
-        args.kwarg?.let {
-            val problem =
-                newProblemDeclaration(
-                    "`kwarg` is not yet supported",
-                    problemType = ProblemNode.ProblemType.TRANSLATION,
-                    rawNode = it
-                )
-            frontend.scopeManager.addDeclaration(problem)
+    private fun handlePositionalArguments(
```
maximiliankaul, line 458: doc string

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -394,95 +395,102 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
 
-        if (args.defaults.isNotEmpty()) {
-            val problem =
-                newProblemDeclaration(
-                    "`defaults` are not yet supported",
-                    problemType = ProblemNode.ProblemType.TRANSLATION,
-                    rawNode = args
-                )
-            frontend.scopeManager.addDeclaration(problem)
+    private fun handleKeywordOnlyArguments(args: Python.ASTarguments) {
```
maximiliankaul, line 484: doc string

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -556,13 +564,30 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         return result
     }
 
-    internal fun handleArgument(node: Python.ASTarg, isPosOnly: Boolean = false) {
+    internal fun handleArgument(
```
maximiliankaul, line 567: doc string

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -394,95 +395,102 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
-            frontend.scopeManager.addDeclaration(problem)
+    private fun handlePositionalArguments(
+        positionalArguments: List<Python.ASTarg>,
+        args: Python.ASTarguments,
+        startIdx: Int
+    ) {
+        val nonDefaultArgsCount = positionalArguments.size - args.defaults.size
+
+        for ((idx, arg) in positionalArguments.drop(startIdx).withIndex()) {
+            val defaultIndex = idx - nonDefaultArgsCount
```
maximiliankaul, line 466: This indexing math looks off to me. Assume `foo(self, a, b = 2, c = 3):` then
- `nonDefaultArgsCount = 4 - 2 = 2`
- `positionalArguments.drop(startIdx)` => this will drop the receiver => a list of 3 elements with indices 0,1,2
- `defaultIndex = idx - nonDefaultArgsCount = 0 - 2 = -2` ???

Either I'm confused or this is off.

maximiliankaul (approved):
I like this PR :+1: as it significantly improves the code quality / readability of my messy argument handling code :see_no_evil: 
There are some nitpicking points to be taken care of and the indexing needs another look (maybe I'm wrong - you're tests seem to work)... I'm confused -> will look at this after my vacation if no one else approves by then.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -394,95 +395,102 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+                    args.defaults.getOrNull(defaultIndex)?.let {
+                        frontend.expressionHandler.handle(it)
+                    }
+                } else {
+                    null
+                }
+            handleArgument(
+                arg,
+                isPosOnly = arg in args.posonlyargs,
+                isKwoOnly = defaultValue != null,
```
maximiliankaul, line 478: is this `isKwoOnly = defaultValue != null,` obvious? otherwise: please some doc

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: I just tried this with the debugger -> there is an error as this code does not find the default for `b`.
Please also include a test case for:
```Python
class Foo:
  def foo(self, a, b = 1, c = 2):
    pass
```

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Since the property “isKwoOnly” is a boolean expression, the result i.e. if the condition is true, returns true otherwise false

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: At the moment we are not handling the case that a receiver can have a default value. As discussed, the receiver (if available) will be created as a variable. I added this information to the doc string.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: The indexing math ensures that default arguments are applied correctly by first dropping the receiver. The defaultIndex is calculated by subtracting the number of non-default arguments, which might result in negative values for arguments without a default. However, this is intentional because we only use defaultIndex when it's greater than or equal to 0, meaning it only applies to arguments with defaults.

I added another test with self, which like the others also passes.

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler/ArgumentsHandlerTest.kt
```diff
@@ -0,0 +1,202 @@
...
+import org.junit.jupiter.api.BeforeAll
+
+class ArgumentsHandlerTest {
+
+    companion object {
+        private lateinit var topLevel: Path
+        private lateinit var result: TranslationResult
+
+        @JvmStatic
+        @BeforeAll
```
KuechA, line 51: Isn't there `kotlin.test.BeforeClass`? This could make it a bit nicer and get rid of the companion and JvmStatic annotation

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler/ArgumentsHandlerTest.kt
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+
+    @Test
+    fun testDefaultsArguments() {
+        var func = result.functions["defaults"]
+        assertNotNull(func, "Function 'defaults' should be found")
+
+        var defaults = mapOf("b" to 1, "c" to 2)
+        for ((paramName, expectedDefaultValue) in defaults) {
```
KuechA, line 137: Does it make sense to add a check that there are no additional arguments (e.g. by checking the size of `func.parameters` is as expected)?

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler/ArgumentsHandlerTest.kt
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    @Test
+    fun testKwDefaultArguments() {
+        var func = result.functions["kw_defaults"]
+        assertNotNull(func, "Function 'kw_defaults' should be found")
+
+        var kwOnlyParams = mapOf("c" to 2, "d" to null, "e" to 3)
+
+        for ((paramName, expectedDefaultValue) in kwOnlyParams) {
```
KuechA, line 112: Does it make sense to add a check that there are no additional arguments (e.g. by checking the size of `func.parameters` is as expected)?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: Please see this test https://github.com/Fraunhofer-AISEC/cpg/pull/1669/commits/8b15388ed69913a33abbe6e48d37efe5b09cecc2 (feel free to edit / update it as I currently do not have access to a Java IDE).

maximiliankaul: Also: it would be great to have your answer in the code as to get a better understanding what and why things are happening here. Maybe a URL to the Python docs where the argument handling is introduced?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: I'm still confused. The method is called `handlePositionalArguments` and this code reads like the _positional_ argument can be _keyword only_? I need to read up on this.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: I fail to see it. Is this some Java / Intellij magic again?

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler/ArgumentsHandlerTest.kt
```diff
@@ -0,0 +1,217 @@
...
+                analyze(listOf(topLevel.resolve(fileName).toFile()), topLevel, true) {
+                    it.registerLanguage<PythonLanguage>()
+                }
+            assertNotNull(result)
+        }
+    }
+
+    @Test
+    fun testPosOnlyArguments() {
+        var func = result.functions["pos_only_and_args"]
```
maximiliankaul, line 68: Some nitpicking: can we make constant stuff `val` instead of `var`?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
KuechA: Thinking more about it: Don't the keyword and positional arguments affect the function calls more than the declarations? Does this "only" require fixing the SymbolResolver or also an update of the frontend?

I think that I can always use the keyword syntax even for positional arguments. Maybe the "only" makes the difference here which is somewhat confusing and it should be "isKeyword" and "isPositional" and then, there might be a combination where everything can be used in the function call

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler/ArgumentsHandlerTest.kt
lshala: good point. Done

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler/ArgumentsHandlerTest.kt
lshala: Done

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler/ArgumentsHandlerTest.kt
lshala: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Sorry, my bad. 
Done.

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler/ArgumentsHandlerTest.kt
lshala: I've changed it to the suggested one.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: As we are handling not only the posonlyargs but also the "normal" args in this method and these can either be kwonlyargs or not, I removed the property isKwOnly here.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: I refactored the method and also added a doc string. Thank you for adding an additional test. Now this test also passes.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
KuechA: Ok, but I think if you have something like `def foo(self, a, b=1)` you would currently get the default value `1` here. I think you have to fix the "offset" of the first default value like in the method below (`positionalArguments.size - args.defaults.size`) or so.

lshala: Done

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/StatementHandlerTest.kt
```diff
@@ -25,21 +25,42 @@
...
 
+    private lateinit var topLevel: Path
+    private lateinit var result: TranslationResult
+
+    @BeforeAll
+    fun setup() {
+        topLevel = Path.of("src", "test", "resources", "python")
+    }
+
+    fun analyzeFile(fileName: String) {
```
oxisto, line 54: Personally I don't like this pattern, but I won't block this PR any further. What is the advantage of storing the result in the class? I only see this pattern usually if multiple tests execute on the same file, but each test is executed on another file and the `result` property gets overwritten. This feels very error prone. I get that the tests contain a lot of boilerplate but then we should see how we can rewrite the `analyze` functions to be leaner and still return the result/tu instead of assigning it to a class member.

lshala: I totally agree with you. I used this pattern for the tests in ArgumentsHandlerTest.kt, in which we are using the same file for every test. 
I will improve it over time.

PR #1675, opened 2024-09-06
Restructure Python.kt -> ast is now encapsulated in an interface.

maximiliankaul:


maximiliankaul:
@oxisto can you please test this on your Mac? I recall some naming issues that we resolved by prefixing everything with `AST`.

PR #1680, opened 2024-09-06
Extending functionality of `ReplaceCallCastPass` to `ResolveCallExpressionAmbiguityPass`

oxisto:
This PR extends the functionality of the `ReplaceCallCastPass` to also replace calls into construct expressions for languages, where we cannot differentiate between them in a frontend.

This also removes a symbol resolver check from the python frontend.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
@@ -194,23 +194,31 @@ interface HasAnonymousIdentifier : LanguageTrait {
 interface HasGlobalVariables : LanguageTrait
 
 /**
- * A language trait, that specifies that the language has so-called functional style casts, meaning
- * that they look like regular call expressions. Since we can therefore not distinguish between a
- * [CallExpression] and a [CastExpression], we need to employ an additional pass ([ReplaceCallPass])
- * after the initial language frontends are done.
+ * A common super-class for all language traits that arise because they are an ambiguity of a
+ * function call, e.g., function-style casts. This means that we cannot differentiate between a
+ * [CallExpression] and other expressions during the frontend and we need to invoke the
+ * [ResolveCallExpressionAmbiguityPass] to resolve this.
  */
-interface HasFunctionalCasts : LanguageTrait
+sealed interface HasCallExpressionAmbiguity : LanguageTrait
 
 /**
- * A language trait, that specifies that the language has functional style constructs, meaning that
- * constructor calls look like regular call expressions (usually meaning that the language has no
- * dedicated `new` keyword).
+ * A language trait, that specifies that the language has so-called functional style casts, meaning
+ * that they look like regular call expressions. Since we can therefore not distinguish between a
+ * [CallExpression] and a [CastExpression], we need to employ an additional pass
+ * ([ResolveCallExpressionAmbiguityPass]) after the initial language frontends are done.
+ */
+interface HasFunctionStyleCasts : HasCallExpressionAmbiguity
+
+/**
+ * A language trait, that specifies that the language has functional style (object) construction,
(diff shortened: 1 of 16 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
```diff
@@ -196,11 +196,22 @@ interface HasGlobalVariables : LanguageTrait
...
+/**
+ * A language trait, that specifies that the language has functional style constructs, meaning that
+ * constructor calls look like regular call expressions (usually meaning that the language has no
+ * dedicated `new` keyword).
+ *
+ * Since we can therefore not distinguish between a [CallExpression] and a [ConstructExpression] in
+ * the frontend, we need to employ an additional pass ([ReplaceCallPass]) after the initial language
+ * frontends are done.
+ */
+interface HasFunctionalConstructs : LanguageTrait
```
konradweiss, line 213: Can we rename this to HasFunctionalConstructors? Constructs sounds to generic to me and I do not intuitively thought that the call of a constructor was meant.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ReplaceCallPass.kt
```diff
@@ -28,29 +28,30 @@ package de.fraunhofer.aisec.cpg.passes
...
+ * If a [Language] has the trait [HasFunctionalCasts] or [HasFunctionalConstructs], we cannot
+ * distinguish between [CallExpression], [CastExpression] or [ConstructExpression] during the
+ * initial translation. This stems from the fact that we might not know all the types yet. We
+ * therefore need to handle them as regular call expression in a [LanguageFrontend] or [Handler] and
+ * then later replace them with a [CastExpression] or [ConstructExpression], if the
+ * [CallExpression.callee] refers to name of a [Type] / [RecordDeclaration] rather than a function.
  */
 @ExecuteBefore(EvaluationOrderGraphPass::class)
 @DependsOn(TypeResolver::class)
-@RequiresLanguageTrait(HasFunctionalCasts::class)
```
konradweiss, line 52: Are we now omitting this because one of two language traits is necessary now and we have no way to express that?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
```diff
@@ -352,7 +350,7 @@ class ExpressionHandler(frontend: PythonLanguageFrontend) :
             } else {
                 // try to resolve -> [ConstructExpression]
                 val currentScope = frontend.scopeManager.currentScope
-                val record =
+                /*val record =
```
konradweiss, line 353: I would delete commented out code. At least so far our unwritten policy was to delete it even if we were not certain to not need it in the future

konradweiss (changes requested):
Some code cleaning is still necessary IMO and we could rename the trait to something more unambiguous.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ReplaceCallPass.kt
konradweiss: So as agreed we can use a base class named `HasFunctionstyleAmbiguity` here.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
konradweiss: or `HasFunctionStyleCast` and `HasFunctionStyleConstruction`

konradweiss (approved):

PR #1683, opened 2024-09-09
Add handleAssert method and test

lshala:


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
```diff
@@ -85,6 +85,15 @@ class ExpressionHandler(frontend: PythonLanguageFrontend) :
         }
     }
 
+    private fun handleNamedExpr(node: Python.AST.NamedExpr): Expression {
```
oxisto, line 88: For new functions (and at some time older) I would prefer a short documentation of what this original node is and into what it is translated. See https://github.com/Fraunhofer-AISEC/cpg/blob/6b9d605ca5e81a8ce3adabb9e91bb21e8886eb5a/cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/DeclarationHandler.kt#L114-L119 for an example

In this case I am not 100 % what exactly this "namedexpr" actually is

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
```diff
@@ -85,6 +85,15 @@ class ExpressionHandler(frontend: PythonLanguageFrontend) :
...
+        return newAssignExpression(
```
oxisto, line 89: You will probably need to set `usedAsExpression` to true. This is a switch whether the `AssignExpression` is used as a statement (false) or an as expression (true)

oxisto (commented):
A short comment what this PR actually does would be appreciated.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
KuechA: @oxisto then, we should note this down somewhere. not sure what would be a suitable place for this though

oxisto (approved):

PR #1684, opened 2024-09-09
Only set unqualified name in python frontend for types

oxisto:
This PR makes sure that Python now properly uses our `TypeResolver` by only setting the unqualified name for a type and let the type resolver handle the rest. Otherwise we will get incorrect results, e.g., when retrieving name of a type within a class.


oxisto:
Was already included in #1606

PR #1694, opened 2024-09-18
Helper functions to access additionalProblems

maximiliankaul:


maximiliankaul:
Should we rename the functions so that `problems` returns everything? Would make more sense to me. And then hide the other two as internal.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -612,6 +612,27 @@ val Node?.assigns: List<AssignExpression>
 val Node?.problems: List<ProblemNode>
     get() = this.allChildren()
 
+/**
+ * Returns all [Node] children in this graph, where the [Node.additionalProblems] field is not
+ * empty.
+ */
+val Node?.additionalProblemsNotEmpty: List<Node>
```
oxisto, line 619: nodesWithProblems?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -612,6 +612,27 @@ val Node?.assigns: List<AssignExpression>
...
+    get() = this.allChildren<Node> { node -> node.additionalProblems.isNotEmpty() }
+
+/**
+ * Returns all [ProblemNode] children - stored in a [Node.additionalProblems] field - in this graph,
+ * starting with this [Node].
+ */
+val Node?.additionalProblemsUnpacked: List<ProblemNode>
```
oxisto, line 626: allProblems?

oxisto (commented):
The naming is slightly confusing. Some tests would be good to better see the expected outcome

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
oxisto: ah. its only “all additional problems”

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
maximiliankaul: Please check the naming again. I'm quite happy with the current name for the additional problem nodes. However, the name for the problem nodes directly in the graph is a bit misleading. Suggestions?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
maximiliankaul: Yes. And I made it "private", as I don't see an external use case.

maximiliankaul:
I also have a version of this with only one instead of the two graph traversals. It is about 5% faster overall (the entire `analyze` function + `.problems` call). However, it is less readable...

```kotlin
val Node?.fastProblems: List<ProblemNode>
    get() {
        val relevantNodes =
            this.allChildren<Node> { it is ProblemNode || it.additionalProblems.isNotEmpty() }

        val result = mutableListOf<ProblemNode>()

        relevantNodes.forEach {
            if (it.additionalProblems.isNotEmpty()) {
                result += it.additionalProblems
            }
            if (it is ProblemNode) {
                result += it
            }
        }

        return result
    }
```

oxisto (approved):

Issue #1602, opened 2024-06-26
ImportDeclaration AST placement


maximiliankaul:
This is just a reminder to a discussion between @oxisto and myself: https://github.com/Fraunhofer-AISEC/cpg/pull/1555/commits/9d02f6c76f975f39119bf8e844af4c4f5b4fa9c5

oxisto:
Currently, we place a declaration both on the AST property `declarations` directly as well as indirectly through the `statements` property and a `DeclarationStatement`. This could potentially be a bad idea, since the AST should be a tree, not a graph.

konradweiss:
Yes, the ast should always be a tree. If it has more connections this can break several things, in the worst case lead to cycles in some algorithms

oxisto:
Also for this discussion: Currently we model a `a: Int` or a simple `a` (the statement, not the expression) as an assignment with no rhs. This works sort-of and the assignment expression will also add a declaration, however the question would be whether we want to directly translate this into a declaration statement / variable declaration.

oxisto:
This just came back to haunt me in #2004 and needs a solution :(

PR #1701, opened 2024-09-20
Support python type hints

### LINKED ISSUE #1649: Python `type_comments`
Investigate and make it work.

```
ast.parse = parse(source, filename='<unknown>', mode='exec', *, type_comments=False, feature_version=None)
    Parse the source into an AST node.
    Equivalent to compile(source, filename, mode, PyCF_ONLY_AST).
    Pass type_comments=True to get back type comments where the syntax allows.
```

### LINKED ISSUE #1655: PEP-484: Support all python type hints
We should support all the possibilities to annotate types in python.

Example for a type comment (This is parsed as an `Assign` with `type_comment`:
```python
a = 42 # type int
```

Example for a type annotation on variable (this is parsed as an `AnnAssign` with `annotation`:
```python
a: int = 42
```

Example for type annotation on function (this is parsed as `returns` on the `FunctionDef` as well as `annotation` on the `arg`):
```python
def bar(x: int) -> int: ...
```

```[tasklist]
### Tasks
- [ ] #1649
- [ ] #1640
- [ ] https://github.com/Fraunhofer-AISEC/cpg/issues/1656
- [x] Support type annotation on `Functiondef` and `arg`
```

### LINKED ISSUE #1656: Support type annotation on `AnnAssign`

lshala:
This PR aims to support python type hints, particularly type_comments and type annotation.

Fixes #1649 
Fixes #1656
Fixes #1655

### DIFF
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
@@ -43,6 +43,7 @@ import de.fraunhofer.aisec.cpg.graph.statements.expressions.ProblemExpression
 import de.fraunhofer.aisec.cpg.graph.types.FunctionType
 import de.fraunhofer.aisec.cpg.helpers.Util
 import kotlin.collections.plusAssign
+import kotlin.reflect.typeOf
 
 class StatementHandler(frontend: PythonLanguageFrontend) :
     PythonHandler<Statement, Python.AST.BaseStmt>(::ProblemExpression, frontend) {
@@ -199,6 +200,14 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
     private fun handleAnnAssign(node: Python.AST.AnnAssign): Statement {
         // TODO: annotations
         val lhs = frontend.expressionHandler.handle(node.target)
+        val annassign =
+            newAssignExpression(
+                lhs = listOf(lhs),
+                rhs = listOf(frontend.expressionHandler.handle(node.value!!)), // TODO !!
+                rawNode = node
+            )
+        val type = frontend.typeOf(node.annotation)
+        annassign.lhs.firstOrNull()?.type = type
         return if (node.value != null) {
             newAssignExpression(
                 lhs = listOf(lhs),
@@ -250,6 +259,9 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
                     },
                 rawNode = node
             )
+        node.type_comment?.let { typeComment ->
+            lhs.forEach { it.type = frontend.typeOf(typeComment) }
+        }
         return newAssignExpression(lhs = lhs, rhs = listOf(rhs), rawNode = node)
     }
 
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
@@ -36,6 +36,7 @@ import de.fraunhofer.aisec.cpg.graph.declarations.*
 import de.fraunhofer.aisec.cpg.graph.statements.AssertStatement
 import de.fraunhofer.aisec.cpg.graph.statements.DeclarationStatement
 import de.fraunhofer.aisec.cpg.graph.statements.Statement
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.AssignExpression
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.Block
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.MemberExpression
@@ -43,7 +44,6 @@ import de.fraunhofer.aisec.cpg.graph.statements.expressions.ProblemExpression
 import de.fraunhofer.aisec.cpg.graph.types.FunctionType
 import de.fraunhofer.aisec.cpg.helpers.Util
 import kotlin.collections.plusAssign
-import kotlin.reflect.typeOf
 
 class StatementHandler(frontend: PythonLanguageFrontend) :
     PythonHandler<Statement, Python.AST.BaseStmt>(::ProblemExpression, frontend) {
@@ -85,7 +85,7 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
     }
 
     /**
-     * Translates a Python [Assert](https://docs.python.org/3/library/ast.html#ast.Assert) into a
+     * Translates a Python [`Assert`](https://docs.python.org/3/library/ast.html#ast.Assert) into a
      * [AssertStatement].
      */
     private fun handleAssert(node: Python.AST.Assert): AssertStatement {
@@ -198,19 +198,14 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
     }
 
     /**
-     * Translates a Python [AnnAssign](https://docs.python.org/3/library/ast.html#ast.AnnAssign]
+     * Translates a Python [`AnnAssign`](https://docs.python.org/3/library/ast.html#ast.AnnAssign)
      * into a [Statement].
      */
-    private fun handleAnnAssign(node: Python.AST.AnnAssign): Statement {
+    private fun handleAnnAssign(node: Python.AST.AnnAssign): AssignExpression {
         val lhs = frontend.expressionHandler.handle(node.target)
         lhs.type = frontend.typeOf(node.annotation)
         val rhs = node.value?.let { listOf(frontend.expressionHandler.handle(it)) } ?: emptyList()
-        val annAssign = newAssignExpression(lhs = listOf(lhs), rhs = rhs, rawNode = node)
-        return if (node.value != null) {
-            annAssign
-        } else {
-            lhs
-        }
+        return newAssignExpression(lhs = listOf(lhs), rhs = rhs, rawNode = node)
     }
 
     private fun handleIf(node: Python.AST.If): Statement {
@@ -238,13 +233,14 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
     }
 
     /**
-     * Translates a Python [Assign](https://docs.python.org/3/library/ast.html#ast.Assign] into a
+     * Translates a Python [`Assign`](https://docs.python.org/3/library/ast.html#ast.Assign) into a
      * [Statement].
      */
-    private fun handleAssign(node: Python.AST.Assign): Statement {
+    private fun handleAssign(node: Python.AST.Assign): AssignExpression {
         val lhs = node.targets.map { frontend.expressionHandler.handle(it) }
         node.type_comment?.let { typeComment ->
-            lhs.forEach { it.type = frontend.typeOf(typeComment) }
+            val tpe = frontend.typeOf(typeComment)
+            lhs.forEach { it.type = tpe }
         }
         val rhs = frontend.expressionHandler.handle(node.value)
         if (rhs is List<*>)
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
@@ -198,22 +198,12 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
     }
 
     private fun handleAnnAssign(node: Python.AST.AnnAssign): Statement {
-        // TODO: annotations
         val lhs = frontend.expressionHandler.handle(node.target)
-        val annassign =
-            newAssignExpression(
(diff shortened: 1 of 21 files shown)


cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/StatementHandlerTest.kt
```diff
@@ -123,4 +119,19 @@ class StatementHandlerTest {
...
     }
+
+    @Test
+    fun testTypeHints() {
+        analyzeFile("type_hints.py")
+
+        // type comments
+        val a = result.refs["a"]
+        assertNotNull(a)
+        assertEquals(result.assertResolvedType("int"), a.type)
```
lshala, line 130: I'm not sure why I can't access assertResolvedType like in [PythonFrontendTest](https://github.com/Fraunhofer-AISEC/cpg/blob/4ee13d6156a98f877d09635c246442d7b458ffe5/cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt#L59C26-L59C44)

oxisto: assertResolvedType is an extension function of `MetadataProvider`. The reason behind this is that it needs access to the type manager (which can be accessed by the translation context). The only way to retrieve the context is through the metadata provider.

In order to use a syntax similar as in the other test is to use the Kotlin `with` function, which changes the specified object to be an implicit receiver within the block (black voodoo Kotlin magic).

```Kotlin
with(result) {
  assertResolvedType("int")
}
```

is equivalent to
```Kotlin
result.assertResolvedType("int")
```

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/StatementHandlerTest.kt
```diff
@@ -123,4 +119,19 @@ class StatementHandlerTest {
         assertNotNull(message, "Assert statement should have a message")
         assertEquals("Test message", message.value, "The assert message is incorrect")
     }
+
+    @Test
+    fun testTypeHints() {
+        analyzeFile("type_hints.py")
```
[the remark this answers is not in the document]
oxisto: I think we don't have a file with type comments (only type annotations), so that's ok.

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/StatementHandlerTest.kt
oxisto: In order to align with the other tests, I would prefer the `with` syntax here.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -43,6 +43,7 @@ import de.fraunhofer.aisec.cpg.graph.statements.expressions.ProblemExpression
 import de.fraunhofer.aisec.cpg.graph.types.FunctionType
 import de.fraunhofer.aisec.cpg.helpers.Util
 import kotlin.collections.plusAssign
+import kotlin.reflect.typeOf
```
oxisto, line 46: wrong import

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -84,7 +85,7 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
     }
 
     /**
-     * Translates a Python (https://docs.python.org/3/library/ast.html#ast.Assert] into a
+     * Translates a Python [Assert](https://docs.python.org/3/library/ast.html#ast.Assert) into a
```
oxisto, line 88: I think in the other functions we were using the syntax with backticks
```suggestion
     * Translates a Python [`Assert`](https://docs.python.org/3/library/ast.html#ast.Assert) into a
```

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -196,15 +197,17 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
     }
 
+    /**
+     * Translates a Python [AnnAssign](https://docs.python.org/3/library/ast.html#ast.AnnAssign]
+     * into a [Statement].
+     */
     private fun handleAnnAssign(node: Python.AST.AnnAssign): Statement {
-        // TODO: annotations
         val lhs = frontend.expressionHandler.handle(node.target)
+        lhs.type = frontend.typeOf(node.annotation)
```
oxisto, line 206: I think that is not necessary, since this should propagate through the assignment expression automatically. Can you see what happens if you remove it?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -196,15 +197,17 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         return frontend.expressionHandler.handle(node.value)
     }
 
+    /**
+     * Translates a Python [AnnAssign](https://docs.python.org/3/library/ast.html#ast.AnnAssign]
+     * into a [Statement].
```
oxisto, line 202:
```suggestion
     * into an [AssignExpression].
```

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -196,15 +197,17 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+     */
     private fun handleAnnAssign(node: Python.AST.AnnAssign): Statement {
```
oxisto, line 204:
```suggestion
    private fun handleAnnAssign(node: Python.AST.AnnAssign): AssignExpression {
```

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -234,8 +237,15 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         return ret
     }
 
+    /**
+     * Translates a Python [Assign](https://docs.python.org/3/library/ast.html#ast.Assign] into a
+     * [Statement].
+     */
     private fun handleAssign(node: Python.AST.Assign): Statement {
```
oxisto, line 244: we can also return an assign expression here now. I think we previously returned a statement here because of the problem expressions.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
```diff
@@ -72,6 +72,7 @@ class PythonAddDeclarationsPass(ctx: TranslationContext) : ComponentPass(ctx) {
         when (node) {
             is AssignExpression -> handleAssignExpression(node)
             is ForEachStatement -> handleForEach(node)
+            is Reference -> handleReference(node)
```
oxisto, line 75: I am pretty sure I intentionally removed this here a while ago. We only want to handle references that are inside an assign expression. Is it needed?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: I think this is a bigger discussion to have. I asked @lshala to implement it this way. Think of the following code: `a: int = "some string"` What do we do?
You are suggesting to drop the `int` by taking the type of the rhs.
I disagree. The annotation is there and should be reflected in the graph. We probably want to raise a warning?

This is the AST for the above snippet:
```python
AnnAssign(
      target=Name(id='a', ctx=Store()),
      annotation=Name(id='int', ctx=Load()),
      value=Constant(value='some string'),
      simple=1)]
```

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -196,18 +197,15 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         return frontend.expressionHandler.handle(node.value)
     }
 
-    private fun handleAnnAssign(node: Python.AST.AnnAssign): Statement {
-        // TODO: annotations
+    /**
+     * Translates a Python [`AnnAssign`](https://docs.python.org/3/library/ast.html#ast.AnnAssign)
+     * into a [Statement].
+     */
+    private fun handleAnnAssign(node: Python.AST.AnnAssign): AssignExpression {
```
maximiliankaul, line 204: Do we need the `simple` property?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
maximiliankaul: This depends on how we implement `a: int`. We now (after my latest change to this PR) have a `AssignExpression` with an empty list as rhs.
What do we do for `a`? This is currently an `Reference` and should then probably be handled as an `AssignExpression` with empty rhs, too. Once we do this, there is no need for this branch.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
oxisto: Ah. I thought the annotation was on the right side. My bad

maximiliankaul: There are two types / styles of annotations. See the test code example in this PR. One is handled as a `AnnAssign`, while the other adds a `type_comment`.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
oxisto: Well in any other language `a: int` would be just a `VariableDeclaration` or more specifically a `DeclarationStatement` with a `VariableDeclaration`.

As which construct is `a: int` parsed in the python AST? is it parsed as an assignment?

oxisto: > What do we do for `a`? This is currently an `Reference` and should then probably be handled as an `AssignExpression` with empty rhs, too. Once we do this, there is no need for this branch.

Semantically this would also be a declaration in my view - if we compare it with other, more sane languages ;)

maximiliankaul: > Well in any other language `a: int` would be just a `VariableDeclaration` or more specifically a `DeclarationStatement` with a `VariableDeclaration`.
> 
> As which construct is `a: int` parsed in the python AST? is it parsed as an assignment?

It's an `AnnAssign` with no value.

```python
AnnAssign(
      target=Name(id='a', ctx=Store()),
      annotation=Name(id='int', ctx=Load()),
      simple=1)
```

lshala: Removed the line

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/StatementHandlerTest.kt
lshala: Done.

oxisto (approved):
Should be ok for now. I added a comment to https://github.com/Fraunhofer-AISEC/cpg/issues/1602#issuecomment-2373784876 about the assignment with no RHS.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
oxisto: As far as I read the documentation, we can disregard the contents of `simple`.

PR #1706, opened 2024-09-20
Changing function signature of `parse` to accept the file content instead of a file

oxisto:
This PR changes the way `parse` works (in a backwards compatible way). Instead of parsing a `File`, we parse the file contents (and a path). The reasoning behind this is that almost all language frontends currently need to read the file contents and we can harmonize this. This will also allow us to provide more common statistics about the parsing context in the future.

### DIFF
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
@@ -28,6 +28,7 @@ package de.fraunhofer.aisec.cpg.frontends.python
 import de.fraunhofer.aisec.cpg.TranslationContext
 import de.fraunhofer.aisec.cpg.frontends.Language
 import de.fraunhofer.aisec.cpg.frontends.LanguageFrontend
+import de.fraunhofer.aisec.cpg.frontends.SupportsNewParse
 import de.fraunhofer.aisec.cpg.frontends.TranslationException
 import de.fraunhofer.aisec.cpg.graph.*
 import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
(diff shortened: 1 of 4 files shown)


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
```diff
@@ -62,21 +64,32 @@ class PythonLanguageFrontend(language: Language<PythonLanguageFrontend>, ctx: Tr
...
      */
     private lateinit var fileContent: String
-    private lateinit var uri: URI
+    private var filePath: Path? = null
 
     @Throws(TranslationException::class)
     override fun parse(file: File): TranslationUnitDeclaration {
-        fileContent = file.readText(Charsets.UTF_8)
-        uri = file.toURI()
+        return parse(file.readText(Charsets.UTF_8), file.toPath())
```
maximiliankaul, line 71: I'm quiet sure this breaks the location properties in the Python ast module.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
```diff
@@ -62,21 +64,32 @@ class PythonLanguageFrontend(language: Language<PythonLanguageFrontend>, ctx: Tr
...
         jep.getInterp().use {
-            it.set("content", fileContent)
-            it.set("filename", file.absolutePath)
+            it.set("content", content)
+            it.set(
+                "filename",
+                if (path != null) {
+                    path.absolute().toString()
+                } else {
+                    "<unknown>"
```
maximiliankaul, line 85: To we want this constant for cpg consistency? Python recommends `<string>` see https://docs.python.org/3/library/functions.html#compile

oxisto: We are using ast.Parse and the default parameter here was `<unknown>`. See https://docs.python.org/3/library/ast.html#ast.parse

oxisto: > Funny, that the ast.parse doc says:
> 
> > Equivalent to compile(source, filename, mode, PyCF_ONLY_AST).
> 
> But I'm ok with the "unknown". This should probably a sane constant for all frontends if there is no file available?

I was intentionally using the python specific value here, but I am fine with either way. Not sure if the SARIF standard defines something in this case, since we are using the location information from SARIF.

oxisto: Removed python for now

konradweiss (approved):
LGTM

PR #1723, opened 2024-09-24
Handle delete statement in python frontend

### LINKED ISSUE #1709: Python frontend: Handle delete statement

lshala:
Fixes #1709


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -83,6 +84,26 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+     */
+    private fun handleDelete(node: Python.AST.Delete): DeleteExpression {
+        val delete = newDeleteExpression(rawNode = node)
+        node.targets.forEach { target ->
+            if (target is Python.AST.Subscript) {
+                delete.targets.add(frontend.expressionHandler.handle(target))
+            } else {
+                delete.additionalProblems +=
+                    newProblemExpression(
+                        problem = "Delete is only supported for Subscript targets.",
```
maximiliankaul, line 99: Why? According to the Python doc: "`targets` is a list of nodes, such as [Name](https://docs.python.org/3/library/ast.html#ast.Name), [Attribute](https://docs.python.org/3/library/ast.html#ast.Attribute) or [Subscript](https://docs.python.org/3/library/ast.html#ast.Subscript) nodes." -> I think we should support Name and Attribute as well. Or is there a good reason not to?

oxisto: The issue is that we cannot fully support the semantics of deleting a name or an attribute, since it un-declares an established symbol (we don't even have an API for that in the scope manager). For subscript nodes, it deletes content inside a list or dict and we can potentially model this with DFG nodes (and in the future maybe do some value approximation). I would suggest to clarify this in the problem expression.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/DeleteExpression.kt
```diff
@@ -34,11 +35,14 @@ class DeleteExpression : Expression() {
     @Relationship("OPERAND") var operandEdge = astOptionalEdgeOf<Expression>()
     var operand by unwrapping(DeleteExpression::operandEdge)
 
+    @Relationship(value = "TARGETS") var targetEdges = astEdgesOf<Expression>()
+    var targets by unwrapping(DeleteExpression::targetEdges)
```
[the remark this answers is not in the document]
lshala: Done.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
lshala: Added additional information to the problem expression message.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -84,6 +85,27 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+    private fun handleDelete(node: Python.AST.Delete): DeleteExpression {
+        val delete = newDeleteExpression(rawNode = node)
+        node.targets.forEach { target ->
+            if (target is Python.AST.Subscript) {
+                delete.operands.add(frontend.expressionHandler.handle(target))
+            } else {
+                delete.additionalProblems +=
+                    newProblemExpression(
+                        problem =
+                            "'Name' and 'Attribute' deletions are not supported, as they removes them from the scope.",
```
maximiliankaul, line 101:
```suggestion
                            "handleDelete: 'Name' and 'Attribute' deletions are not supported, as they removes them from the scope.",
```

I'm trying to do a new thing here: have the function name as part of the error message to make it easier to match them later and build statistics.

oxisto (approved):
LGTM. We can ignore the test coverage here because it was because of missing C++ coverage

PR #1728, opened 2024-09-25
Name: add get a random name

maximiliankaul:
This PR adds a get-a-random-`Name` functionality to the `Name`.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Handler.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Handler.kt
@@ -168,7 +168,6 @@ abstract class Handler<ResultNode : Node?, HandlerNode, L : LanguageFrontend<in 
     }
 
     /** Creates a random name starting with a prefix plus a random UUID (version 4). */
-    @OptIn(kotlin.uuid.ExperimentalUuidApi::class)
     fun getRandomTempName(): Name {
         val prefix = "TEMP-RANDOM-NAME"
         val randomPart = Uuid.random().toString()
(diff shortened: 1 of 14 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Handler.kt
```diff
@@ -166,6 +167,13 @@ abstract class Handler<ResultNode : Node?, HandlerNode, L : LanguageFrontend<in
         return null
     }
 
+    /** Creates a random name starting with a prefix plus a random UUID (version 4). */
+    fun getRandomTempName(): Name {
+        val prefix = "TEMP-RANDOM-NAME"
```
maximiliankaul, line 172: Any chance to make this constant available for tests as well? I tried moving it to the class, but I obviously do not have a handler available during the tests. Maybe there is a better place for this?

oxisto, line 172: After some thought I would like to propose three things:
- probably move this to a companion object of `Name`
- move the prefix as a parameter (maybe with a default)
- get rid of `-` (or at least have `_`), we should make this "name" as natural as possible, so that we don't create any future problems if we ever try to parse this as a real name in the language or something like that. most languages won't accept `-` in the var name I guess?

```
var myLoopVar = Name.random("loopVar")
```

and then I end up with `loopVar1231938190238012301023213`

maximiliankaul: Thoughts about the current version?

oxisto: Looks good! How about just `random` instead of the very long `getRandomTempName` ?

maximiliankaul: I renamed it. However, I'm not a fan of having super short names -> I like names that describe things. You can just hit TAB in your idea and don't have any additional typing work to do...

oxisto (approved):

PR #1730, opened 2024-09-25
Implemented base class for `ExceptHandler`

oxisto:
This PR removes a TODO we had in the code, where our class `excepthandler` was both the base "interface" as well as the concrete AST node. This splits up both, similar to `ast.stmt` and other cases.


oxisto:
Please double-check @maximiliankaul, test coverage will be 0 % until we merge in the other PR so we can ignore the test coverage for now.

PR #1731, opened 2024-09-25
Open a block and function scope for python functions

oxisto:
This PR makes sure we open a new (block) scope for the function body. This is not a 1:1 mapping to python scopes, since python only has a "function scope", but in the CPG the function scope only comprises the function arguments, and we need a block scope to hold all local variables within the function body.


maximiliankaul (approved):
Assuming that we don't want to change CPG behavior -> I'm ok with this workaround.

oxisto:
> Assuming that we don't want to change CPG behavior -> I'm ok with this workaround.

I am still not 100 % convinced myself it this is a good idea or not.

oxisto:
> > Assuming that we don't want to change CPG behavior -> I'm ok with this workaround.
> 
> I am still not 100 % convinced myself it this is a good idea or not.

after discussion with @maximiliankaul we agree that this is the "best" possible workaround for the moment

PR #1733, opened 2024-09-26
Implement a throw or raise statement

maximiliankaul:
This PR implements a new `ThrowStatement`
- [ ] Check the new node. Does it work for all supported languages?
- [x] Implement EOG
- [x] Implement DFG
- [x] Implement all other relevant passes (are there any?)
- [x] Implement fluent tests

The language specific implementations will be handled in separate PRs:
- C++ currently uses a `UnaryOperator`
- Implement for Python #1741 
- Implement for other languages

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
@@ -42,9 +42,7 @@ class ThrowStatement : Statement(), ArgumentHolder {
 
     /**
      * Some languages (Python) can add a parent exception (or `cause`) to indicate that an exception
-     * was raised while handling another exception. This is stored in the graph, but has no further
-     * implications like EOG or DFG connections, as it is only of informational purpose, but it
-     * doesn't change the program behavior.
+     * was raised while handling another exception.
      */
     @Relationship(value = "PARENT_EXCEPTION")
     var parentExceptionEdge = astOptionalEdgeOf<Expression>()
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
@@ -30,6 +30,7 @@ import de.fraunhofer.aisec.cpg.graph.edges.ast.astOptionalEdgeOf
 import de.fraunhofer.aisec.cpg.graph.edges.unwrapping
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.Expression
 import java.util.Objects
+import org.apache.commons.lang3.builder.ToStringBuilder
 import org.neo4j.ogm.annotation.Relationship
 
 /** Represents a `throw` or `raise` statement. */
@@ -83,4 +84,12 @@ class ThrowStatement : Statement(), ArgumentHolder {
     }
 
     override fun hashCode() = Objects.hash(super.hashCode(), exception, parentException)
+
+    override fun toString(): String {
+        return ToStringBuilder(this, TO_STRING_STYLE)
+            .appendSuper(super.toString())
+            .append("exception", exception)
+            .append("parentException", parentException)
+            .toString()
+    }
 }
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
@@ -70,11 +70,7 @@ class ThrowStatement : Statement(), ArgumentHolder {
     }
 
     override fun hasArgument(expression: Expression): Boolean {
-        return when {
-            exception == expression -> true
-            cause == expression -> true
-            else -> false
-        }
+        return exception == expression || cause == expression
     }
 
     override fun equals(other: Any?): Boolean {
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
@@ -40,18 +40,19 @@ class ThrowStatement : Statement(), ArgumentHolder {
     var exception by unwrapping(ThrowStatement::exceptionEdge)
 
     /**
-     * Some languages (Python) can add a cause to indicate that an exception was raised while
-     * handling another exception. This is stored in the graph, but has no further implications like
-     * EOG or DFG connections, as it is only of informational purpose, but it doesn't change the
-     * program behavior.
+     * Some languages (Python) can add a parent exception (or `cause`) to indicate that an exception
+     * was raised while handling another exception. This is stored in the graph, but has no further
+     * implications like EOG or DFG connections, as it is only of informational purpose, but it
+     * doesn't change the program behavior.
      */
-    @Relationship(value = "CAUSE") var causeEdge = astOptionalEdgeOf<Expression>()
-    var cause by unwrapping(ThrowStatement::causeEdge)
+    @Relationship(value = "PARENT_EXCEPTION")
+    var parentExceptionEdge = astOptionalEdgeOf<Expression>()
+    var parentException by unwrapping(ThrowStatement::parentExceptionEdge)
 
     override fun addArgument(expression: Expression) {
         when {
             exception == null -> exception = expression
-            cause == null -> cause = expression
+            parentException == null -> parentException = expression
         }
     }
 
@@ -61,8 +62,8 @@ class ThrowStatement : Statement(), ArgumentHolder {
                 exception = new
                 true
             }
-            cause == old -> {
-                cause = new
+            parentException == old -> {
+                parentException = new
                 true
             }
             else -> false
@@ -70,14 +71,16 @@ class ThrowStatement : Statement(), ArgumentHolder {
     }
 
     override fun hasArgument(expression: Expression): Boolean {
-        return exception == expression || cause == expression
+        return exception == expression || parentException == expression
     }
 
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
         if (other !is ThrowStatement) return false
-        return super.equals(other) && exception == other.exception && cause == other.cause
+        return super.equals(other) &&
+            exception == other.exception &&
+            parentException == other.parentException
     }
 
-    override fun hashCode() = Objects.hash(super.hashCode(), exception, cause)
+    override fun hashCode() = Objects.hash(super.hashCode(), exception, parentException)
 }
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
@@ -33,11 +33,11 @@ import java.util.Objects
 import org.neo4j.ogm.annotation.Relationship
 
 /** Represents a `throw` or `raise` statement. */
-class RaiseStatement : Statement(), ArgumentHolder {
+class ThrowStatement : Statement(), ArgumentHolder {
 
     /** The exception object to be raised. */
     @Relationship(value = "EXCEPTION") var exceptionEdge = astOptionalEdgeOf<Expression>()
-    var exception by unwrapping(RaiseStatement::exceptionEdge)
+    var exception by unwrapping(ThrowStatement::exceptionEdge)
 
     /**
      * Some languages (Python) can add a cause to indicate that an exception was raised while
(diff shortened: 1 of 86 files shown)


cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/DataflowTest.kt
```diff
@@ -101,4 +117,35 @@ class DataflowTest {
...
+                                declare { variable("a", t("short")) { literal(42) } }
+                                `throw` { call("SomeError") { /* TODO parameter a */} }
+                            }
+                        }
+                    }
+                }
+            }
+
+        // Let's assert that we did this correctly
+        val main = result.functions[0] // TODO: why can't I use "foo" like with the other tests?
```
KuechA, line 138: You can but you have to fix your imports to `import de.fraunhofer.aisec.cpg.graph.*` to find the overloaded version of the `[]` operator

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/DataflowTest.kt
```diff
@@ -101,4 +117,35 @@ class DataflowTest {
...
+        assertNotNull(main)
+        val body = main.body
+        assertIs<Block>(body)
+
+        val throwStmt = body.statements.getOrNull(0)
```
KuechA, line 143:
```suggestion
        val throwStmt = body.statements.getOrNull(1)
```
Statement 0 is the declaration statement of `a`.

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/DataflowTest.kt
```diff
@@ -101,4 +117,35 @@ class DataflowTest {
...
+    @Test
+    fun testThrow() {
+        val result =
+            TestLanguageFrontend().build {
+                translationResult {
+                    translationUnit("some.file") {
+                        function("foo", t("void")) {
+                            body {
+                                declare { variable("a", t("short")) { literal(42) } }
+                                `throw` { call("SomeError") { /* TODO parameter a */} }
```
KuechA, line 130: Just adding `ref("a")` to the lambda of the `call`-function should do the job

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
```diff
@@ -1028,6 +1029,10 @@ open class EvaluationOrderGraphPass(ctx: TranslationContext) : TranslationUnitPa
         pushToEOG(stmt)
     }
 
+    protected fun handleThrowStatement(statement: ThrowStatement) {
+        TODO("Needs implementing")
```
KuechA, line 1033:
```suggestion
        createEOG(statement.exception)
        createEOG(statement.cause)
        pushToEOG(statement)
```
Is this the right order of cause and exception?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
```diff
@@ -138,6 +139,33 @@ class DFGPass(ctx: TranslationContext) : ComponentPass(ctx) {
...
+     * - [Reference]
+     */
+    protected fun handleThrowStatement(
+        node: ThrowStatement,
+        inferDfgForUnresolvedSymbols: Boolean
+    ) {
+        node.exception?.let { exc ->
+            when (exc) {
+                is CallExpression -> {
+                    handleCallExpression(exc, inferDfgForUnresolvedSymbols)
```
KuechA, line 155: In fact, this should be much easier to handle ;) I guess what you want to do is to generate a DFG from the `node.exception` to `node` and maybe also from `node.cause` to `node`? This would be simply:
```kt
node.exception?.let{ node.prevDFGEdges += it }
node.cause?.let{ node.prevDFGEdges += it }
```
Since the exception and cause are part of the AST, we will visit and handle them automatically.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
```diff
@@ -0,0 +1,87 @@
...
+    @Relationship(value = "EXCEPTION") var exceptionEdge = astOptionalEdgeOf<Expression>()
+    var exception by unwrapping(ThrowStatement::exceptionEdge)
+
+    /**
+     * Some languages (Python) can add a cause to indicate that an exception was raised while
+     * handling another exception. This is stored in the graph, but has no further implications like
+     * EOG or DFG connections, as it is only of informational purpose, but it doesn't change the
+     * program behavior.
+     */
+    @Relationship(value = "CAUSE") var causeEdge = astOptionalEdgeOf<Expression>()
```
KuechA, line 48: Can this only be a `Reference` in all cases? Currently, it's any `Expression`. Adapt the type or it would make sense to include them in the EOG and DFG.

Anyway, I would recommend adding the `cause` to the DFG because e.g. if there's sensitive data inside the Exception in the "cause" we may also have this data in the exception raised here; this could violate some policies and the data flow would be lost without the DFG-edge.

For the EOG, it may be useful as well if we can do some more sophisticated things (and it shouldn't harm to have an edge here I guess).

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
```diff
@@ -0,0 +1,87 @@
...
+            cause == old -> {
+                cause = new
+                true
+            }
+            else -> false
+        }
+    }
+
+    override fun hasArgument(expression: Expression): Boolean {
+        return when {
```
KuechA, line 73: `return exception == expression || cause == expression`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
KuechA: However, I'm not sure how to connect the `statement.nextEOG` edge to either a `CatchStatement` (if it's surrounded by a try-catch with a matching catch) the FunctionDeclaration or if this will be subject of this PR at all. Might also be a separate PR to not block other developments.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
oxisto: The cause can be any expression, either a call / constructor to a new exception or a reference to an already created one. So, yes it would make sense to have a DFG from cause to throw.

oxisto: Ah, I confused exception with cause... But same applies, in python this can be any expression.

oxisto: Can we maybe rename this field to `parentException`, which is a bit more generic?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
oxisto: In theory we might actually have all the necessary information because the EOG runs after the type resolver and we can the get the type information out of the `CatchStatement` and then match it. If there is no match, we should probably connect the EOG to the function declaration

oxisto: Ah lol we already have a (working?) implementation in handleThrowOperator! This handles the C++ implementation with a `throw` unary operator. But we have a problem here... This relies on a `TryScope` to find a matching TryStatement. But in python we do not open a scope for `try` since it does not open a new (variable) scoping. This is now the problem that we are mixing scopes for variable visibility and other stuff :(

Additionally, the implementation  is not yet handling nested try statements.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
maximiliankaul: This is copy & paste from other nodes to be in line with the style.
Changed.

KuechA:
We should keep in mind to enable throwing/raising the exception when handling the `with` statement once this feature is available and #1757 is merged.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
maximiliankaul: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
maximiliankaul: Renaming done.

KuechA:
The changes itself look ok to me, but the test coverage is a bit low. Easy fixes could be: Try different throws (with and without parentException, with and without exception). It would also make sense to cover the EOG by having some try/catch surrounding the throw

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
```diff
@@ -0,0 +1,95 @@
...
+class ThrowStatement : Statement(), ArgumentHolder {
+
+    /** The exception object to be raised. */
+    @Relationship(value = "EXCEPTION") var exceptionEdge = astOptionalEdgeOf<Expression>()
+    var exception by unwrapping(ThrowStatement::exceptionEdge)
+
+    /**
+     * Some languages (Python) can add a parent exception (or `cause`) to indicate that an exception
+     * was raised while handling another exception. This is stored in the graph, but has no further
+     * implications like EOG or DFG connections, as it is only of informational purpose, but it
```
oxisto, line 46: is that true? are we NOT connecting it to the EOG/DFG and if yes - why? We should do that in order to at least resolve possible symbols

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
```diff
@@ -138,6 +139,17 @@ class DFGPass(ctx: TranslationContext) : ComponentPass(ctx) {
         }
     }
 
+    /**
+     * Handle a [ThrowStatement]. Currently, we support two types of [ThrowStatement.exception],
+     * which are then forwarded to the appropriate handlers:
```
oxisto, line 144: Why is this mentioned here explicitly? For me it seems that this is completely opaque. We store whatever the expression handler gives us, so no need to restrict it.

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/DataflowTest.kt
```diff
@@ -101,4 +106,25 @@ class DataflowTest {
             assertEquals(0, node2.prevDFGEdges.size)
         }
     }
+
+    @Test
+    fun testThrow() {
+        val result = prepareThrowDFGTest()
+
+        // Let's assert that we did this correctly
+        val main = result.functions[0]
```
oxisto, line 115: I think the function is called "foo" and we should probably also retrieve it like `functions["foo"]` in case we extend the test and move things around.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
KuechA: Everything is connected. I'm currently updating the documentation

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
KuechA: Removed

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/DataflowTest.kt
KuechA: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
KuechA: I also added the expected behavior to the specification pages

oxisto (approved):

PR #1735, opened 2024-09-26
Implementing python's `global`

oxisto:
This PR tries to implement python's `global` keyword and tries to replicate python variable resolution as good as possible.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
```diff
@@ -90,6 +90,9 @@ abstract class Scope(
      */
     @Transient var wildcardImports: MutableSet<ImportDeclaration> = mutableSetOf()
 
+    /** A list of scope modifications. */
+    @Transient var scopeModifications: MutableMap<Name, Scope> = mutableMapOf()
```
oxisto, line 94: this would be the alternative approach to the node. Depending on what we decide, this would be removed

maximiliankaul (commented):
I'd appreciate to have two PRs:
- one for the CPG changes (with it's own tests, so that we can see how it's supposed to work in general and not only in the specific python use-case)
- one making use of the first PR by implementing the Python stuff

I get why you modify the scope of a `Reference` as this is how the Python front-end works. However, don't we need a scope modifier for the declaration (I get that this will be problematic with the Python "add declarations pass"?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/StatementBuilder.kt
```diff
@@ -329,3 +330,22 @@ fun MetadataProvider.newDefaultStatement(rawNode: Any? = null): DefaultStatement
     log(node)
     return node
 }
+
+/**
+ * Creates a new [DefaultStatement]. The [MetadataProvider] receiver will be used to fill different
```
maximiliankaul, line 335: ReferenceScopeModifierStatement

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/StatementHandlerTest.kt
```diff
@@ -232,4 +232,70 @@ class StatementHandlerTest : BaseTest() {
...
+
+    @Test
+    fun testGlobal() {
+        var file = topLevel.resolve("global.py").toFile()
+        val result = analyze(listOf(file), topLevel, true) { it.registerLanguage<PythonLanguage>() }
+        assertNotNull(result)
+
+        // There should be three variable declarations, two local and one global
+        var cVariables = result.variables("c")
+        assertEquals(3, cVariables.size)
```
maximiliankaul, line 244: I see the global one. But where is the second local one? I only see the `c = 3`

maximiliankaul: Ah, the error one.

oxisto:
> I'd appreciate to have two PRs:
> 
> * one for the CPG changes (with it's own tests, so that we can see how it's supposed to work in general and not only in the specific python use-case)
> * one making use of the first PR by implementing the Python stuff

Done. https://github.com/Fraunhofer-AISEC/cpg/pull/1742

> 
> I get why you modify the scope of a `Reference` as this is how the Python front-end works. However, don't we need a scope modifier for the declaration (I get that this will be problematic with the Python "add declarations pass"?

The declaration's scope is not modified. Only the "lookup scope" of the symbol that you are trying to resolve in the current scope is modified (e.g., from the local to the global scope).

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -537,6 +527,35 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         return wrapDeclarationToStatement(result)
     }
 
+    private fun handleGlobal(global: Python.AST.Global): LookupScopeStatement {
```
maximiliankaul, line 530: Please add a doc string similar to the other functions.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -537,6 +527,35 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+            frontend.scopeManager.globalScope?.children?.firstOrNull { it is NameScope }
+
+        return newLookupScopeStatement(
+            global.names.map { parseName(it).localName },
+            pythonGlobalScope,
+            rawNode = global
+        )
+    }
+
+    private fun handleNonLocal(global: Python.AST.Nonlocal): LookupScopeStatement {
```
maximiliankaul, line 544: Please add a doc string similar to the other functions.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -537,6 +527,35 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+        // We need to find the first outer function scope, or rather the block scope belonging to
+        // the function
+        var outerFunctionScope =
+            frontend.scopeManager.firstScopeOrNull {
+                it is BlockScope && it != frontend.scopeManager.currentScope
```
maximiliankaul, line 549: Do we need to consider nested functions here, too?

oxisto: Unfortunately, I cannot test this, as we are not parsing nested function at all at the moment (see #1736), but I am beginning to hate python even more. It seems that the nonlocal statement actually jumps to the nested function scope where the variable is "declared" and not just the next one. See example
```
def level1():
    c = 1
    def level2():
        d = 1
        def level3():
            nonlocal c, d
            c = c + 2
            d = d + 1
            print(c)
        
        level3()
    level2()
level1()
```

In this case the nonlocal c "jumps" to level 1 and nonlocal d jumps to level 2, but we cannot retrieve this information in the frontend so I would take this as a close approximation and ignore this very narrow corner case.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
```diff
@@ -79,59 +78,105 @@ class PythonAddDeclarationsPass(ctx: TranslationContext) : ComponentPass(ctx) {
         }
     }
 
-    /*
-     * Return null when not creating a new decl
+    /**
+     * This function will create a new dynamic [VariableDeclaration] if there is a write access to
```
maximiliankaul, line 82: This function does not check whether there is a write access. However, the doc implies this. Maybe rephrase to "should be called when handling write access invocations"?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
```diff
@@ -79,59 +78,105 @@ class PythonAddDeclarationsPass(ctx: TranslationContext) : ComponentPass(ctx) {
...
+
+        // First, check if we need to create a field
+        var field: FieldDeclaration? =
+            when {
+                // Check, whether we are referring to a "self.X", which would create a field
+                scopeManager.isInRecord && scopeManager.isInFunction && ref.refersToReceiver -> {
+                    // We need to temporarily jump into the scope of the current record to
+                    // add the field. These are instance attributes
+                    scopeManager.withScope(scopeManager.firstScopeIsInstanceOrNull<RecordScope>()) {
+                        newFieldDeclaration(ref.name)
```
maximiliankaul, line 119: Copy code and location from the reference?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
```diff
@@ -79,59 +78,105 @@ class PythonAddDeclarationsPass(ctx: TranslationContext) : ComponentPass(ctx) {
...
                 }
+                scopeManager.isInRecord && scopeManager.isInFunction && ref is MemberExpression -> {
+                    // If this is any other member expression, we are usually not interested in
+                    // creating fields, except if this is a receiver
+                    return null
+                }
+                scopeManager.isInRecord -> {
+                    // We end up here for fields declared directly in the class body. These are
+                    // class attributes; more or less static fields.
+                    newFieldDeclaration(ref.name)
```
maximiliankaul, line 130: Copy code and location from the reference?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
```diff
@@ -79,59 +78,105 @@ class PythonAddDeclarationsPass(ctx: TranslationContext) : ComponentPass(ctx) {
...
+                }
+            }
+
+        // If we didn't create any field up to this point and if we are still have not returned, we
+        // can create a normal variable. We need to take scope modifications into account.
+        var decl =
+            if (field == null && targetScope != null) {
+                scopeManager.withScope(targetScope) { newVariableDeclaration(ref.name) }
+            } else if (field == null) {
+                newVariableDeclaration(ref.name)
```
maximiliankaul, line 143: Copy code and location from the reference?

maximiliankaul (approved):

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
oxisto: I was actually thinking in adding another check so I can do that, should be minimal effort for the compiler.

oxisto: Then I can also get rid of the weird check if this is part of a call expression

oxisto: There is a bug with write references in a named expression. https://github.com/Fraunhofer-AISEC/cpg/pull/1785 needs to be resolved first.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: Will be addressed in #1736 -> resolved for now.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
maximiliankaul: Check implemented

PR #1750, opened 2024-09-29
Setting correct block location for `Python.AST.ExceptHandler`

### LINKED ISSUE #1748: Could not set location on wrapped block because the parent node is not a python statement
Occurs when trying to parse larger projects, such as glance.

```
14:25:16,855 WARN  StatementHandler Could not set location on wrapped block because the parent node is not a python statement
```

Still trying to figure out which nodes is causing this

oxisto:
Fixes #1748

### DIFF
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
@@ -527,8 +527,7 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
             // Make sure we open a new (block) scope for the function body. This is not a 1:1
             // mapping to python scopes, since python only has a "function scope", but in the CPG
             // the function scope only comprises the function arguments, and we need a block scope
-            // to
-            // hold all local variables within the function body.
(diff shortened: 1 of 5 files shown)


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -772,16 +771,17 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
-        return if (baseStmt != null) {
-            result.codeAndLocationFromChildren(baseStmt)
-        } else {
-            // Otherwise, continue without setting the location
-            log.warn(
-                "Could not set location on wrapped block because the parent node is not a python statement"
-            )
-            result
+        val ast = parentNode as? Python.AST.AST
+        if (ast != null) {
```
maximiliankaul, line 775: Why are we removing this warning? Is it irrelevant?

oxisto: Yes it is, this is needed because of the stupid workaround with interfaces vs. abstract classes, all nodes are ` Python.AST.AST` but we need to use the interface because of the async workaround.

PR #1757, opened 2024-10-01
Implement Python `with`

### LINKED ISSUE #1711: Implement python `With`
We should implement this like a `try`-Statement. The `with`s are declared before the try and the try contains the main body with a finally that closes the opened resources.

lshala:
This PR handles the `with` statement like a `try...finally` statement, as discussed in [#1695](https://github.com/Fraunhofer-AISEC/cpg/discussions/1695). To correctly manage the ` __enter__ ` and ` __exit__` methods of Python`s context manager ([PEP 343](https://peps.python.org/pep-0343/#specification-the-with-statement)), the whole statement is wrapped in an additional block.

Fixes #1711

### DIFF
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
@@ -418,11 +418,7 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         ret.condition = frontend.expressionHandler.handle(node.test)
         ret.statement = makeBlock(node.body, parentNode = node)
         if (node.orelse.isNotEmpty()) {
-            ret.additionalProblems +=
-                newProblemExpression(
-                    problem = "Cannot handle \"orelse\" in while loops.",
-                    rawNode = node
-                )
+            ret.elseStatement = makeBlock(node.orelse, parentNode = node)
         }
         return ret
     }
@@ -488,11 +484,7 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         }
 
         if (node.orelse.isNotEmpty()) {
-            ret.additionalProblems +=
-                newProblemExpression(
-                    problem = "handleFor: Cannot handle \"orelse\" in for loops.",
-                    rawNode = node
-                )
+            ret.elseStatement = makeBlock(node.orelse, parentNode = node)
         }
         return ret
     }
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
@@ -201,7 +201,7 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         fun generateEnterCallAndAssignment(
             managerName: Name,
             withItem: Python.AST.withitem
-        ): Pair<Expression, Name> {
+        ): Pair<AssignExpression, Name> {
             val tmpValName = Name.random(prefix = WITH_TMP_VAL)
             val enterVar = newReference(name = tmpValName).implicit()
             val enterCall =
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
@@ -117,14 +117,38 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
      * ```
      */
     private fun handleWithStatement(node: Python.AST.With): Block {
-        /** Prepares the __exit__(manager, None, None, None) call for the else-block. */
-        fun generateExitCallWithNone(
-            manager: Expression,
-            withItem: Python.AST.withitem
-        ): Statement {
+        /**
+         * Prepares the `manager = ContextManager()` and returns the random name for the "manager"
+         * as well as the assignment.
+         */
+        fun generateManagerAssignment(withItem: Python.AST.withitem): Pair<AssignExpression, Name> {
+            // Create a temporary reference for the context manager
+            val managerName = Name.random(prefix = CONTEXT_MANAGER)
+            val manager = newReference(name = managerName).implicit()
+
+            // Handle the 'context expression' (the part before 'as') and assign to tempRef
+            // Represents the line `manager = ContextManager()`
+            val contextExpr = frontend.expressionHandler.handle(withItem.context_expr)
+            val managerAssignment =
+                newAssignExpression(
+                        operatorCode = "=",
+                        lhs = listOf(manager),
+                        rhs = listOf(contextExpr)
+                    )
+                    .implicit()
+            return Pair(managerAssignment, managerName)
+        }
+
+        /** Prepares the `__exit__(manager, None, None, None)` call for the else-block. */
+        fun generateExitCallWithNone(managerName: Name, withItem: Python.AST.withitem): Statement {
             val exitCallWithNone =
                 newMemberCallExpression(
-                        callee = newMemberExpression(name = "__exit__", base = manager).implicit(),
+                        callee =
+                            newMemberExpression(
+                                    name = "__exit__",
+                                    base = newReference(name = managerName).implicit()
+                                )
+                                .implicit(),
                         rawNode = withItem
                     )
                     .implicit()
@@ -138,13 +162,18 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
          * Prepares the if-statement which is the body of the catch block. This includes the call of
          * `__exit__(manager, *sys.exc_info())`, the negation and the throw statement.
          */
-        fun generateExitCallWithSysinfo(
-            manager: Expression,
+        fun generateExitCallWithSysExcInfo(
+            managerName: Name,
             withItem: Python.AST.withitem
         ): Statement {
             val exitCallWithSysExec =
                 newMemberCallExpression(
-                        callee = newMemberExpression(name = "__exit__", base = manager).implicit(),
+                        callee =
+                            newMemberExpression(
+                                    name = "__exit__",
+                                    base = newReference(name = managerName).implicit()
+                                )
+                                .implicit(),
                         rawNode = withItem
                     )
                     .implicit()
@@ -163,18 +192,26 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         }
 
         /**
-         * calls __enter__() and assign to another random variable. Represents the line "tmpVal =
-         * manager.__enter__()"
+         * calls __enter__() and assign to another random variable. Represents the line
+         *
+         * ```python
+         * tmpVal = manager.__enter__()
+         * ```
          */
         fun generateEnterCallAndAssignment(
-            manager: Expression,
+            managerName: Name,
             withItem: Python.AST.withitem
         ): Pair<Expression, Name> {
-            val enterVarName = Name.random(prefix = CONTEXT_MANAGER + "Enter")
-            val enterVar = newReference(name = enterVarName).implicit()
+            val tmpValName = Name.random(prefix = WITH_TMP_VAL)
+            val enterVar = newReference(name = tmpValName).implicit()
             val enterCall =
                 newMemberCallExpression(
-                        callee = newMemberExpression(name = "__enter__", base = manager).implicit(),
+                        callee =
+                            newMemberExpression(
+                                    name = "__enter__",
+                                    base = newReference(name = managerName).implicit()
+                                )
+                                .implicit(),
                         rawNode = withItem
                     )
                     .implicit()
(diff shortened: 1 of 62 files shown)


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -826,4 +887,41 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+    ): MemberCallExpression {
+        return newMemberCallExpression(
+                callee = newMemberExpression(name = name, base = base),
+                rawNode = item
+            )
+            .implicit()
+    }
+
+    /** Creates a [TryStatement]. */
+    private fun createTryStatementWithBody(
```
oxisto, line 911: I would avoid these extra functions. We are currently looking into how to give the node creation a more "fluid" feeling and once that is done this will probably easier but in the mean-time I would avoid these extra functions which are basically just one-liners

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -90,6 +91,66 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+     * [PEP 343](https://peps.python.org/pep-0343/#specification-the-with-statement). The context
+     * managers `__enter__()` runs before the try block and `__exit__()` is always called later in
+     * the `finally` block to clean up, whether an exception occurs or not.
+     */
+    private fun handleWithStatement(node: Python.AST.With): Block {
+        val mainBlock = newBlock().implicit()
+        val exitStatements = mutableListOf<Statement>()
+        val tryBlock = newBlock().implicit()
+
+        node.items.forEach { withItem ->
```
oxisto, line 108: In general this is ok for me but we have to agree on this approach because it is a somewhat approximation of the "real" behavior. Basically what you modeled is (with the optional value part):
```python
tmpVar = contextExpr
var = contextExpr.__enter__()
try:
   BODY
finally:
  tmpVar.__exit__()
```

However, in reality the statement behaves like this (adopted from https://peps.python.org/pep-0343/#specification-the-with-statement):
```Python
mgr = contextExpr
tmpVar = mgr.__enter__()
exc = True
try:
    try:
        var = tmpVar
        BLOCK
    except:
        # The exceptional case is handled here
        exc = False
        if not mgr.__exit__(*sys.exc_info()):
            raise
        # The exception is swallowed if exit() returns true
finally:
    # The normal and non-local-goto cases are handled here
    if exc:
        mgr.__exit__(mgr, None, None, None)
```
(the exit calls are actually static calls to the type of the mgr, but we can forego that)

So it is a double-nested try and it seems the intention is to call the exit function with additional information when an exception occurred. Because you probably will not want to call exit again after the exception is handled, the exc check is used and the return code of exit is used to determine if the exception is thrown from the inner try to the outer try (and thus to the block, since the outer try has no catch/except).

Any opinions about how detailed we want to model this? @KuechA @konradweiss @maximiliankaul

cpg-language-python/src/test/resources/python/with.py
```diff
@@ -0,0 +1,5 @@
+with open("file.txt", "r") as file:
```
oxisto, line 1: I would prefer a test case using a class that implements the context manager protocol, similar to what we sketched together

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: I'd probably go with a single layer of `try` / `finally`

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -90,6 +91,66 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
 
+    /**
+     * Translates a Python [`With`](https://docs.python.org/3/library/ast.html#ast.With) into a
+     * [Block].
+     *
+     * We return a Block to handle the with statement, following
+     * [PEP 343](https://peps.python.org/pep-0343/#specification-the-with-statement). The context
+     * managers `__enter__()` runs before the try block and `__exit__()` is always called later in
+     * the `finally` block to clean up, whether an exception occurs or not.
+     */
```
maximiliankaul, line 102: Please add some demo example here. See https://github.com/Fraunhofer-AISEC/cpg/blob/f18321644560ea752fbbe902ed4d9b3614b9c86b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt#L254 for an example.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -90,6 +91,66 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+    private fun handleWithStatement(node: Python.AST.With): Block {
+        val mainBlock = newBlock().implicit()
```
maximiliankaul, line 104: `Block` is wrong here. Should be a try / catch, once the other PRs are merged.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -90,6 +91,66 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+        val exitStatements = mutableListOf<Statement>()
+        val tryBlock = newBlock().implicit()
```
maximiliankaul, line 106: code and loc

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -90,6 +91,66 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+
+        node.items.forEach { withItem ->
+            // Create a temporary reference for the context manager
+            val tempVarName = Name.random(prefix = CONTEXT_MANAGER)
+            val tempRef = newReference(name = tempVarName).implicit()
```
maximiliankaul, line 111: code and loc

cpg-language-python/src/test/resources/python/with.py
maximiliankaul: Yes, but please also keep this test, too. It's a very common pattern in python code.

lshala: Done.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
KuechA: Actually, we can use pretty much what is used in the documentation: https://docs.python.org/3/reference/compound_stmts.html#the-with-statement ;)

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
KuechA: What would you expect here? The whole while or only the body?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
KuechA: Which other PRs do you mean? We have to return multiple statements here and afaik this has to be wrapped in a single statement by using a `Block`. If we had a way to return multiple statements which are then all added to the surrounding block, it would be great but I don't see any PR aiming at this.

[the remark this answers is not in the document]
KuechA: I'll go for `val mainBlock = newBlock().codeAndLocationFromOtherRawNode(node).implicit()` which will work after #1784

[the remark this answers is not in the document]
KuechA: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: Given your "translation" in the doc string: only the `body` as this would be the closest match.

cpg-language-python/src/test/resources/python/with.py
```diff
@@ -0,0 +1,16 @@
...
```
maximiliankaul, line 1: Can we please also have a test with multiple resources? Something along the line
```python
with A() as a, B(a) as b, C(a, b) as c:
    doSomething(a, c)
```
See https://stackoverflow.com/a/1073814

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -80,6 +80,158 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+     *     cm = tmpVal # Doesn't exist if no variable is used
+     *     cm.doSomething()
+     * except:
+     *     if not manager.__exit__(*sys.exc_info()):
+     *         raise
+     * else:
+     *     manager.__exit__(None, None, None)
+     * ```
+     */
+    private fun handleWithStatement(node: Python.AST.With): Block {
```
maximiliankaul, line 119: This function is quiet a hand full. I think it might be easier to read, if we split it into multiple helpers. for creating all the required implicit cpg nodes. I'll create a PR to see if it's actually better...

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
KuechA: Done

cpg-language-python/src/test/resources/python/with.py
KuechA: Hm, this requires nesting the `with` statements. Currently, this is not done correctly

oxisto: I think this is already implemented as you are looping over the "node.items". The above example has 3 "items".

KuechA: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
KuechA: Since adding the multi-resource thing required a heavy refactoring, it should already be simplified a bit

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -117,14 +117,86 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+                    )
+                    .implicit()
+            val starOp = newUnaryOperator("*", false, false)
+            starOp.input =
+                newMemberExpression("exec_info", newReference("sys").implicit()).implicit()
+            exitCallWithSysExec.addArgument(starOp)
+
+            val ifStmt = newIfStatement().implicit()
+            // TODO: Needs #1733 or 1741, then add:
+            //   ifStmt.thenStatement = newThrowStatement().implicit()
```
oxisto, line 158: would it make sense to add a unary operator call to "throw" in the mean time (as its modelled for C++?) or should we just hope that we don't forget it here? Maybe add an issue?

KuechA: I'd prefer to leave a comment in the respective PRs since they should be about to be finalized too

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -117,14 +117,86 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+        }
+
         val mainBlock = newBlock().codeAndLocationFromOtherRawNode(node).implicit()
-        val exitStatementsElse = mutableListOf<Expression>()
-        val exitStatementsCatch = mutableListOf<Expression>()
-        val tryBlock =
-            newBlock() // We set it as implicit below because there we also have a code and
-        // location.
+
+        var currentBlock = mainBlock
```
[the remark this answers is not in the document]
KuechA: Done. I slightly changed the iteration

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
oxisto: ok

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -80,6 +80,221 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+                        val catchClause =
+                            newCatchClause().implicit().apply {
+                                this.body =
+                                    newBlock().implicit().apply {
+                                        this.statements.add(
+                                            generateExitCallWithSysExcInfo(managerName, withItem)
+                                        )
+                                    }
+                            }
+                        this.catchClauses.add(catchClause)
```
maximiliankaul, line 277: inline newCatchClause

maximiliankaul (approved):
Ready to merge

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: Looks pretty good now, thanks :)

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: As discussed: added at all reasonable places.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
KuechA: I inlined everything

PR #1779, opened 2024-10-03
Support for python packages

oxisto:
This PR adds (basic) support for python *packages* in addition to python *modules*.


oxisto:
This is unfortunately blocked until we have some sort of package dependency graph.

oxisto:
Depends on #1782 
Depends on #1783

oxisto:
This should work even better once the blocked PRs are merged in, but this can already be reviewed independently I think.

KuechA (approved):

PR #1781, opened 2024-10-04
Add new function `lookupUniqueTypeSymbolByName`

### LINKED ISSUE #1766: Clean up and speed up symbol and type resolution
Originated out of #1690, I decided to split up the original PR because it got too large to review.

oxisto:
This adds two new functions `ScopeManager.lookupUniqueTypeSymbolByName` and `Reference.doesReferToType`. This harmonizes a lot of boilerplate code in type resolver, cxx extra pass and resolve ambuigity pass, which were all trying to achieve the same thing.

Fixes #1766

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
@@ -366,7 +366,7 @@ val Collection<Type>.commonType: Type?
  * this can only be used in passes.
  */
 context(Pass<*>)
-fun Reference.doesReferToType(): Type? {
+fun Reference.nameIsType(): Type? {
     // First, check if it is a simple type
     var type = language?.getSimpleTypeOf(name)
     if (type != null) {
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
@@ -32,7 +32,10 @@ import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.TemplateDeclaration
 import de.fraunhofer.aisec.cpg.graph.scopes.Scope
 import de.fraunhofer.aisec.cpg.graph.scopes.TemplateScope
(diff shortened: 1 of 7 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ResolveCallExpressionAmbiguityPass.kt
```diff
@@ -114,34 +108,10 @@ class ResolveCallExpressionAmbiguityPass(ctx: TranslationContext) : TranslationU
...
-                        )
-                    } else {
-                        callee.name
-                    }
-
-                val type = typeManager.lookupResolvedType(fqn)
-                if (type != null) {
-                    walker.replaceCallWithCast(type, parent, call, pointer)
-                }
+            // Check if it is type and replace the call
```
KuechA, line 111: The unwrapping of the pointer is missing. Was there a reason why it ever existed? Looks weird but such weird things are not my main area of expertise

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ResolveCallExpressionAmbiguityPass.kt
oxisto: It is done with `callee.unwrapReference` at the beginning of the function now. It is needed because we can run into something like `(*MyStruct)(nil)` in Go, which is also parsed as call expression with a callee of a unary expression

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
```diff
@@ -353,3 +356,29 @@ val Collection<Type>.commonType: Type?
...
+/**
+ * A utility function that checks whether our [Reference] refers to a [Type]. This is used by many
+ * passes that replace certain [Reference] nodes with other nodes, e.g., the
+ * [ResolveCallExpressionAmbiguityPass].
+ *
+ * Note: This involves some symbol lookup (using [ScopeManager.lookupUniqueTypeSymbolByName]), so
+ * this can only be used in passes.
+ */
+context(Pass<*>)
+fun Reference.doesReferToType(): Type? {
```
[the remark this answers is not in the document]
oxisto: I renamed it into `nameIsType`. Is that better?

KuechA (approved):

PR #1783, opened 2024-10-07
Make sure to move `typeObservers` from old to new node when replacing nodes

oxisto:
Previously, the old node still had all the type observers, which meant that the nodes registered as observers did not get any updates. Instead, we now unregister all nodes from the old one and register them as type observers for the new node. Maybe we need to make this configurable instead of the default? We definitely need it when we replace call nodes int cast/construct expressions - but we might not need it for other cases.


maximiliankaul (approved):
Looks good to me. Two points:
- the function doc explicitly states, that *only* EOG edges are re-wired -> might be good to include the type observers there, too.
- I cannot comment on your question in your PR description. I do not see a reason not to have them rewired but I might be missing something in the big picture.

oxisto:
> Looks good to me. Two points:
> 
> * the function doc explicitly states, that _only_ EOG edges are re-wired -> might be good to include the type observers there, too.
> * I cannot comment on your question in your PR description. I do not see a reason not to have them rewired but I might be missing something in the big picture.

Good point, I added the documentation.

PR #1797, opened 2024-10-21
Handle nested functions

### LINKED ISSUE #1736: Nested functions are not added to AST
In python it is possible to nest functions in functions. For some reason, they do not end up in the AST at all. I stumbled across this once I tried to implement #1713

KuechA:
Nested functions have not been added to the AST because they occur inside a `BlockScope` but the `Block` only accepted `VariableDeclaration`s as locals. This is not correct and has been extended to `ValueDeclaration` which also covers the `FunctionDeclaration`.

Fixes #1736

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/Statement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/Statement.kt
@@ -45,9 +45,9 @@ import org.neo4j.ogm.annotation.Relationship
 @NodeEntity
 abstract class Statement : Node(), DeclarationHolder {
     /**
-     * A list of local variables associated to this statement, defined by their
-     * [VariableDeclaration] extracted from Block because for, while, if, switch can declare locals
-     * in their condition or initializers.
+     * A list of local variables (or other values) associated to this statement, defined by their
+     * [ValueDeclaration] extracted from Block because `for`, `while`, `if`, and `switch` can
+     * declare locals in their condition or initializers.
      *
      * TODO: This is actually an AST node just for a subset of nodes, i.e. initializers in for-loops
      */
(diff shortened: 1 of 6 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/Statement.kt
```diff
@@ -43,13 +45,14 @@ import org.neo4j.ogm.annotation.Relationship
 @NodeEntity
 abstract class Statement : Node(), DeclarationHolder {
     /**
-     * A list of local variables associated to this statement, defined by their [ ] extracted from
-     * Block because for, while, if, switch can declare locals in their condition or initializers.
+     * A list of local variables associated to this statement, defined by their
```
oxisto, line 48: doc still says only "variables" and variable declarations.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/Statement.kt
```diff
@@ -67,6 +70,8 @@ abstract class Statement : Node(), DeclarationHolder {
     override fun addDeclaration(declaration: Declaration) {
         if (declaration is VariableDeclaration) {
```
oxisto, line 71: can't we just add all value declarations here? do we really need to differentiate what they are?

KuechA: I have no idea. If there would be any reason to **not** support one of {`EnumConstantDeclaration`, `ParameterDeclaration`, `ProblemDeclaration`, `TypeParameterDeclaration`}, we could keep this check (but it's a bit sneaky and nobody will realize a problem). Similarly, if we would like to differentiate between different types of declarations, it could make sense. If we can't come up with anything useful, I'll just add all `ValueDeclaration`s here

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/Statement.kt
KuechA: Updated the docs

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/Statement.kt
oxisto: Let's keep it for now, I want to re design the addDeclaration bit anyway, probably as part of the NodeBuilder v2 API.

oxisto (approved):

PR #1799, opened 2024-10-21
add a "problem" warning when encountering "async"

maximiliankaul:
The recent PR #1795 does not add a warning when encountering `async with` (unlike the other `async` handlers). This PR adds the warning.

### DIFF
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
@@ -249,6 +249,8 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         val result =
             newBlock().codeAndLocationFromOtherRawNode(node as? Python.AST.BaseStmt).implicit()
 
+        addAsyncWarning(node, result)
+
         // If there are multiple elements in node.items, we have to nest the try statements.
         // We start with a generic block for the outer context manager.
         // For i > 1, we add context_manager[i] to the try-block of item[i-1]
@@ -525,14 +527,7 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
      */
     private fun handleFor(node: Python.AST.NormalOrAsyncFor): ForEachStatement {
         val ret = newForEachStatement(rawNode = node)
-        if (node is IsAsync) {
-            ret.addDeclaration(
-                newProblemDeclaration(
-                    problem = "The \"async\" keyword is not yet supported.",
-                    rawNode = node
-                )
-            )
-        }
+        addAsyncWarning(node, ret)
 
         ret.iterable = frontend.expressionHandler.handle(node.iter)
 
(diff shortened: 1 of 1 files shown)


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -1116,4 +1104,18 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
     }
+
+    /**
+     * Checks whether [mightBeAsync] has the [IsAsync] property and adds a warning to the
+     * corresponding [parentNode] stored in [Node.additionalProblems].
+     */
+    private fun addAsyncWarning(mightBeAsync: Python.AST.AsyncOrNot, parentNode: Node) {
+        if (mightBeAsync is IsAsync) {
+            parentNode.additionalProblems +=
+                newProblemDeclaration(
```
KuechA, line 1115: Should this be of type `Parsing` or `Translation`? We can parse the stuff but we can't model it adequately, so I think it should be `Translation`.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -1116,4 +1104,18 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         declStmt.addDeclaration(decl)
         return declStmt
     }
+
+    /**
+     * Checks whether [mightBeAsync] has the [IsAsync] property and adds a warning to the
```
KuechA, line 1109:
```suggestion
     * Checks whether [mightBeAsync] implements the [IsAsync] interface and adds a warning to the
```

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -1116,4 +1104,18 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+     * corresponding [parentNode] stored in [Node.additionalProblems].
+     */
+    private fun addAsyncWarning(mightBeAsync: Python.AST.AsyncOrNot, parentNode: Node) {
```
KuechA, line 1112: I'd call it simply `node` for simplicity

KuechA (approved):

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: I wanted to make it perfectly clear that this is not the node we are testing `async` on. I'll change this, if you insist, however, I prefer it this way. Or maybe `reportingNode`??

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
KuechA: Since there's only one node involved, it should be pretty clear already but I'm also ok like this

PR #1809, opened 2024-10-31
Rework extensions to follow subgraphs

KuechA:
Reworks extensions following prev/next DFG and EOG and also adds CDG- and PDG-following functions. Both new functions feature an optional interprocedural analysis which explores the callgraph.

Since there was a bunch of redundant code, the iteration was merged into a single method and only the way in which to collect the next step has to be provided by the respective specialized subgraph-following-methods.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
@@ -200,19 +200,38 @@ class FulfilledAndFailedPaths(val fulfilled: List<List<Node>>, val failed: List<
  * but not mandatory**. If the list "failed" is empty, the data flow is mandatory.
  */
 fun Node.followPrevFullDFGEdgesUntilHit(predicate: (Node) -> Boolean): FulfilledAndFailedPaths {
-    return followXUntilHit({ currentNode -> currentNode.prevFullDFG }, true, true, predicate)
+    return followXUntilHit(
+        x = { currentNode -> currentNode.prevFullDFG },
+        collectFailedPaths = true,
+        findAllPossiblePaths = true,
+        predicate = predicate
+    )
 }
 
 fun Node.collectAllPrevCDGPaths(interproceduralAnalysis: Boolean): List<List<Node>> {
     // We make everything fail to reach the end of the CDG. Then, we use the stuff collected in the
     // failed paths (everything)
-    return this.followPrevCDGUntilHit(true, true, interproceduralAnalysis) { false }.failed
+    return this.followPrevCDGUntilHit(
+            collectFailedPaths = true,
+            findAllPossiblePaths = true,
+            interproceduralAnalysis = interproceduralAnalysis
+        ) {
+            false
+        }
+        .failed
 }
 
 fun Node.collectAllNextCDGPaths(interproceduralAnalysis: Boolean): List<List<Node>> {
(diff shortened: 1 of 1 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -296,6 +342,28 @@ fun Node.followNextFullDFGEdgesUntilHit(
...
+ * Hence, if "fulfilled" is a non-empty list, a data flow from [this] to such a node is **possible
+ * but not mandatory**. If the list "failed" is empty, the data flow is mandatory.
+ */
+fun Node.followNextFullDFGEdgesUntilHit(
+    collectFailedPaths: Boolean = true,
+    findAllPossiblePaths: Boolean = true,
+    predicate: (Node) -> Boolean
+): FulfilledAndFailedPaths {
+    return followXUntilHit(
+        { currentNode -> currentNode.nextFullDFG },
```
oxisto, line 360: minor thing: you can probably specify this as a method reference here, but more or less just for style reasons.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -307,45 +375,14 @@ fun Node.followNextFullDFGEdgesUntilHit(
...
-            }
-        }
-    }
-
-    return FulfilledAndFailedPaths(fulfilledPaths, failedPaths)
+    return followXUntilHit(
+        { currentNode ->
+            currentNode.nextEOGEdges.filter { it.unreachable != true }.map { it.end }
+        },
+        true,
```
oxisto, line 382: I would use keyword args for those boolean properties

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -359,45 +396,14 @@ fun Node.followNextEOGEdgesUntilHit(predicate: (Node) -> Boolean): FulfilledAndF
...
-            }
-        }
-    }
-
-    return FulfilledAndFailedPaths(fulfilledPaths, failedPaths)
+    return followXUntilHit(
+        { currentNode ->
+            currentNode.prevEOGEdges.filter { it.unreachable != true }.map { it.start }
+        },
+        true,
```
oxisto, line 403: I would use keyword args for those boolean properties

oxisto (approved):
Approved with minor comments

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: You mean just writing `::nextFullDFG`? That doesn't work for me because it's not a method

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: Done

PR #1810, opened 2024-11-05
Simplified and more basic fixpoint iteration

KuechA:
The current fix-point iteration system is a bit tricky to understand and hard to debug. I'd try to remove some of the bloat and simplify it. In particular, it should keep less states (i.e., save memory, remove a bunch of indirections) and copy objects instead of manipulating them (which should make it easier to understand things).


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
```diff
@@ -0,0 +1,251 @@
...
+ * typically has to abstract the value for the specific purpose.
+ *
+ * This class is actually used to hold individual instances of the lattice's elements and to compute
+ * bigger elements depending on these two elements.
+ *
+ * Implementations of this class have to implement the comparator, the least upper bound of two
+ * lattices.
+ */
+interface LatticeElement<T> : Comparable<LatticeElement<T>> {
+    val elements: T
```
oxisto, line 49: The wording is still confusing to me. Is `LatticeElement` now representing one element of the set (represented as the lattice) and if yes, why is its property also called `elements`?

From what I gather this really is the complete lattice, so maybe drop the `Element` from the class name? Can we constrain `T` to a collection?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
```diff
@@ -0,0 +1,251 @@
...
+     * Computes the least upper bound of this lattice and [other]. It returns a new object and does
+     * not modify either of the objects.
+     */
+    fun lub(other: LatticeElement<out T>): LatticeElement<in T>
+
+    /** Duplicates the object, i.e., makes a deep copy. */
+    fun duplicate(): LatticeElement<in T>
+}
+
+typealias PowersetLatticeT<V> = PowersetLattice<IdentitySet<V>, V>
```
oxisto, line 61: ?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
```diff
@@ -0,0 +1,251 @@
...
+        // The call of `toSet` ensures that we don't get stuck for different types of sets.
+        return other is PowersetLattice<V, T> && this.elements.toSet() == other.elements.toSet()
+    }
+
+    override fun hashCode(): Int {
+        return super.hashCode() * 31 + elements.hashCode()
+    }
+}
+
+typealias MapLatticeT<K, V> = LatticeElement<Map<K, V>>
```
oxisto, line 98: What's with the `T`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/EOGWorklist.kt
```diff
@@ -0,0 +1,67 @@
...
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.helpers.functional
+
+import de.fraunhofer.aisec.cpg.graph.edges.flows.EvaluationOrder
+import java.util.IdentityHashMap
+
+inline fun <reified V> iterateEOGClean(
```
oxisto, line 31: Is there also a iterateEOGDirty?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
```diff
@@ -0,0 +1,251 @@
...
+        } else if (this.elements.containsAll(other.elements)) {
+            1
+        } else {
+            -1
+        }
+    }
+
+    override fun equals(other: Any?): Boolean {
+        // The call of `toSet` ensures that we don't get stuck for different types of sets.
+        return other is PowersetLattice<V, T> && this.elements.toSet() == other.elements.toSet()
```
oxisto, line 90: Is there no other way? this does not seem like a good idea

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
```diff
@@ -0,0 +1,251 @@
...
+
+typealias PowersetLatticeT<V> = PowersetLattice<IdentitySet<V>, V>
+
+inline fun <reified V> emptyPowersetLattice() = PowersetLattice<IdentitySet<V>, V>(identitySetOf())
+
+/**
+ * Implements the [LatticeElement] for a lattice over a set of nodes. The lattice itself is
+ * constructed by the powerset.
+ */
+open class PowersetLattice<V : IdentitySet<T>, T>(override val elements: V) : LatticeElement<V> {
```
oxisto, line 69: I fail to understand what this class actually makes a powerset lattice.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/EOGWorklist.kt
```diff
@@ -0,0 +1,67 @@
...
+    startEdges: List<EvaluationOrder>,
+    startState: LatticeElement<V>,
+    transformation: (EvaluationOrder, LatticeElement<V>) -> LatticeElement<V>,
+): LatticeElement<V> {
+    val globalState = IdentityHashMap<EvaluationOrder, LatticeElement<V>>()
```
oxisto, line 36: Please document what is happening here

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
oxisto: If I understand it correctly, then *if* the elements would contain a powerset, then this would be considered a lattice because then the lab is provided by set union.

BTW: why don't we need the greatest lower bound?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
KuechA: It's a single element inside the lattice. I refactored `elements` to `value` which may help a bit. I could probably also implement just individual functions for `lub` and the comparisons without having this wrapper class. However, I wouldn't store all possible elements a lattice can have because we would e.g. have to compute and (temporarily) store the powerset of all nodes in a function (for a lattice over a powerset) and, when it comes to the maps, even map all nodes to each element in the powerset. This simply requires too much memory and it's better to compute them on-demand.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
KuechA: Mistake and is updated to make sense

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/EOGWorklist.kt
KuechA: Yes :D The currently used version is a bit strange. But I re-named this one to "new", so it sounds less weird ;)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
KuechA: Since the value of both elements should be a `IdentitySet` now, this should not be required and I removed it

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLattices.kt
KuechA: I renamed it to `PowersetLatticeElement` because it implements only a single element (i.e., one set and the operations).

A greatest lower bound is required for a complete lattice which we do not necessarily need here

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/EOGWorklist.kt
KuechA: I added some documentation

oxisto (approved):
I like it. Some smaller documentation issues and maybe we can avoid some of the result order comparisons.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
```diff
@@ -0,0 +1,541 @@
...
+ * structure which only make sense in a certain context. [T] depends on the analysis and typically
+ * has to abstract the value for the specific purpose.
+ *
+ * This class provides functionality to
+ * - compute the least upper bound (also called join) of two elements.
+ * - compute the greatest lower bound (also called meet) of two elements.
+ * - compare two elements.
+ * - duplicate/clone an element.
+ *
+ * Note: We usually do not want (nor have to) store the elements spanning the lattice because it
```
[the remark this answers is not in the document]
KuechA: Depends... In the `UnreachableEOGPass`, I store all elements, in the CDG nothing at all

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/IdentitySet.kt
```diff
@@ -43,14 +43,14 @@ import java.util.concurrent.atomic.AtomicInteger
  * be very resource-intensive if nodes are very similar but not the *same*, in a work-list however
  * we only want just to avoid to place the exact node twice.
  */
-open class IdentitySet<T> : MutableSet<T> {
+open class IdentitySet<T>(expectedMaxSize: Int = 16) : MutableSet<T> {
```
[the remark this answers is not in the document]
KuechA: I documented this

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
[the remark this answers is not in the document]
KuechA: Clarified this in the documentation

PR #1811, opened 2024-11-05
 Simplified and more basic fixpoint iteration for `UnreachableEOGPass`

KuechA:
This PR adapts the `UnreachableEOGPass` to the new fixpoint iteration implementation


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
```diff
@@ -91,24 +96,34 @@ class UnreachableEOGPass(ctx: TranslationContext) : TranslationUnitPass(ctx) {
...
+    lattice: Lattice<UnreachabilityStateElement>,
+    currentEdge: EvaluationOrder,
+    currentState: UnreachabilityStateElement,
+): UnreachabilityStateElement {
+    val lattice = lattice as? UnreachabilityState ?: return currentState
+    var newState = currentState
     when (val currentNode = currentEdge.end) {
         is IfStatement -> {
-            handleIfStatement(currentEdge, currentNode, currentState)
+            newState = handleIfStatement(lattice, currentEdge, currentNode, newState)
```
konradweiss, line 107: While looking through the code I noticed that we have this explicit handlers for `IfStatement` and `WhileStatement`, but not for the other branching Constructs. is that intentional? Are they not Needed for for, foreach, do and switch?

konradweiss (commented):
Generally the changes look adequate, but I noticed that some implementation may have missed in the old and new version-

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
KuechA: I'll double-check if some of these statements can be added. I think that some of these do not exist in the current implementation because the `ValueEvaluator` is still limited (this affects the foreach) and because I wasn't sure how to implement an appropriate comparison (this is the case for the switch).

KuechA: It should generalize to other loops as well. For the switch, we cannot easily identify a matching edge/case statement which is why I only added a TODO for this one.

KuechA: Added #2058

konradweiss (approved):

PR #1813, opened 2024-11-05
Add link from EOG Pass to Specs

### LINKED ISSUE #1760: Provide link to EOG specification in each handler function of the EOG

KuechA:
Closes #1760


konradweiss (changes requested):
Add one general link to the specification in the class doc of the pass together with a reminder for future developers to extend the documentation.

One additional thing that we "could" change, but this is more a usability remark, is to change the individual link texts to "<Nodeclass>-Specification" instead of simply  "Specification" to make it clearer that they lead to different parts of the specification.

PR #1815, opened 2024-11-05
Replace UnaryOperator("throw") with ThrowStatement for CXX

KuechA:
Replaces the usage of `UnaryOperator` with operatorCode "throw" with the new node `ThrowStatement` for the CXX-frontend. We have to change the class `ThrowStatement` to inherit from `Expression` since this is required in cxx (for the handler, so probably also by the language). This means that we also rename `ThrowStatement` to `ThrowExpression`.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
@@ -33,8 +33,8 @@ import java.util.Objects
 import org.apache.commons.lang3.builder.ToStringBuilder
 import org.neo4j.ogm.annotation.Relationship
 
-/** Represents a `throw` or `raise` statement. */
-class ThrowStatement : Statement(), ArgumentHolder {
+/** Represents a `throw` or `raise` statement/expression. */
+class ThrowStatement : Expression(), ArgumentHolder {
 
     /** The exception object to be raised. */
     @Relationship(value = "EXCEPTION") var exceptionEdge = astOptionalEdgeOf<Expression>()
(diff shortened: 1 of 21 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
```diff
@@ -33,8 +33,8 @@ import java.util.Objects
 import org.apache.commons.lang3.builder.ToStringBuilder
 import org.neo4j.ogm.annotation.Relationship
 
-/** Represents a `throw` or `raise` statement. */
-class ThrowStatement : Statement(), ArgumentHolder {
+/** Represents a `throw` or `raise` statement/expression. */
+class ThrowStatement : Expression(), ArgumentHolder {
```
oxisto, line 37: Of course C++ has to be the party pooper :( Ok it seems like it make sense to make `throw` an expression (of void type?), but then I guess we should rename it to a `ThrowExpression` otherwise its very confusing.

oxisto (changes requested):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ThrowStatement.kt
KuechA: yes, wanted to do this separately/at the end of my PR session to avoid a bunch of collisions ;)

oxisto: Ok good idea. Then I'll accept this as-is under the strong promise of the follow up PR ;)

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/ThrowStatementTest.kt
```diff
@@ -29,7 +29,7 @@ import de.fraunhofer.aisec.cpg.GraphExamples.Companion.testFrontend
 import de.fraunhofer.aisec.cpg.TranslationConfiguration
```
oxisto, line 29: Rename test file

KuechA: Done

PR #1818, opened 2024-11-05
Fix some typos and unnecessary casts in LLVM frontend

KuechA:


cpg-language-llvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/llvm/StatementHandler.kt
```diff
@@ -1391,9 +1391,8 @@ class StatementHandler(lang: LLVMIRLanguageFrontend) :
         // make an assignment in each BB.
         val functionName = LLVMGetValueName(bbsFunction).string
         val functions =
```
oxisto, line 1393: can't we just use `tu.functions(functionName)` here?

oxisto (approved):

cpg-language-llvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/llvm/StatementHandler.kt
KuechA: Yes. Changed

PR #1822, opened 2024-11-06
Optimized region utils

oxisto:
It seems that region utils (whatever they do) checked for a potential new line character in every call of `codeAndLocationFromChildren`. But the new line char should stay constant within the file, so we are using the `lineSeperator` property (in the python frontend) instead.


maximiliankaul (approved):
LGTM. Only problem I see: `newLineTpye` isn't the best variable name in my book. Maybe something like `lineBreakSequence` or `fileLineSeperator`?

oxisto:
> LGTM. Only problem I see: `newLineTpye` isn't the best variable name in my book. Maybe something like `lineBreakSequence` or `fileLineSeperator`?

I will rename it to newLineChars

PR #1828, opened 2024-11-08
Add exclusion filter for directories

lshala:
This PR adds the possibility to exclude folders or files from the analysis. Specifically, it should be possible to filter either by `strings` or by `regex`.


cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt
```diff
@@ -1586,6 +1586,29 @@ class PythonFrontendTest : BaseTest() {
...
     }
 
+    @Test
+    fun test() {
+        val directoryPath = Path.of("/home", "lshala", "repos", "nova")
+        val excludedDirs = listOf("tests", "accelerator")
+        val tr =
+            analyze(".py", directoryPath, usePasses = true) {
+                it.registerLanguage<PythonLanguage>()
+                it.excludedDirs(excludedDirs)
```
lshala, line 1596: @oxisto As discussed it would be helpful to have further exclusion patterns. Would it be useful to have both a straightforward exclusion list like `excludedDirs` and an additional property (`exlusionPatterns`?) for using regex patterns, e.g.
```
val exclusionPatterns = listOf(
   Regex(".*\\.test\\..*")
   Regex(".*__tests__.*")
)
```
(Test test() will be removed)

oxisto: Yes it would be helpful :)

lshala: instead of the first proposed way of filtering via two properties, I would use only one property that allows filtering via strings or regex

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
```diff
@@ -146,6 +146,13 @@ private constructor(
 
             var sourceLocations: List<File> = ctx.config.softwareComponents[sc] ?: listOf()
 
+            sourceLocations =
```
oxisto, line 149: I think this is not enough, I think you need to move the exclusion handling down to the file walk (or after the walk), because at this point here the locations are not "expanded". For example, if I set the source location "mypackage", which has a "tests" folder in it. In this case the exclusion filter filters based on mypackage, but not on mypackage/tests which is only expanded about 10 lines down.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -453,6 +456,20 @@ private constructor(
...
+        /**
+         * Sets a list of exclusion patterns for filtering files and directories.
+         *
+         * @param patterns List of exclusion patterns: strings are treated as Regex patterns.
+         *   Example:
+         * ```
+         * exclusionPatterns(listOf(".*test(s)?", "tests"))
+         * ```
+         */
+        fun exclusionPatterns(patterns: List<String>): Builder {
```
oxisto, line 468: Any particular reason not just expose this as a List<Regex>? Seems error-prone if I expect a list that is somehow turned into a regex magically.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -453,6 +459,33 @@ private constructor(
...
+        /**
+         * Sets a list of exclusion patterns using regular expressions for filtering files and
+         * directories.
+         *
+         * @param patterns List of exclusion patterns. Example:
+         * ```
+         * exclusionPatterns(listOf(Regex(".*test(s)?")))
+         * ```
+         */
+        fun exclusionPatternsByRegex(patterns: List<Regex>): Builder {
```
lshala, line 471: Not sure if we should use two overloaded functions, e.g. “exclusionPatterns” or separate as here

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
lshala: Done, added separate functions for regex and strings

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
oxisto: You can overload them.

lshala: Done. As far as I unterstand, the `JvmName` annotation is required because the two functions only differ by a generic parameter and the JVM erases that parameter during compilation

oxisto: Can we have varargs instead of a list? Then they should be different and varargs might be nice to work with

oxisto (approved):
At least one test would be a good idea, otherwise LGTM

PR #1842, opened 2024-11-14
Fix overload operator-> for CallExpression

peckto:


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -398,6 +398,22 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
         val callee = call.callee
         val language = call.language
 
+        // Check for a possible overloaded operator->
+        if (
+            callee.language is HasOperatorOverloading &&
+                callee is MemberExpression &&
+                callee.operatorCode == "->" &&
+                callee.base.type !is PointerType
```
oxisto, line 406: We pretty much have the same code for member access. Is it possible to extract it unto a common function?

peckto:
Can someone explain to me, why the codecov/patch QA fails? I don't get it.

oxisto:
> Can someone explain to me, why the codecov/patch QA fails? I don't get it.

It is probably failing because you are not reaching all branch targets in the line (yellow).

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
peckto: yes, good point. Fixed

peckto:
> > Can someone explain to me, why the codecov/patch QA fails? I don't get it.
> 
> It is probably failing because you are not reaching all branch targets in the line (yellow).

The red one seems to be gone now. Dont know about the yellow one, but it does not block the CI ^^

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -335,13 +342,8 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
         // Check for a possible overloaded operator->
-        if (
-            reference.language is HasOperatorOverloading &&
-                reference is MemberExpression &&
-                reference.operatorCode == "->" &&
-                reference.base.type !is PointerType
-        ) {
-            val result = resolveOperator(reference)
+        if (isOverloadedOperator(reference, "->")) {
+            val result = resolveOperator(reference as MemberExpression)
```
peckto, line 346: Is it possible to get rid of the `as MemberExpression` which is ensured by the `isOverloadedOperator` function? The linter seems not to get this.

oxisto: Yes with a fancy Kotlin magic called contracts :)
```
contract { returns() implies (ex is MemberExpression) }
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -398,6 +400,17 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
         val callee = call.callee
         val language = call.language
 
+        // Check for a possible overloaded operator->
+        if (isOverloadedOperator(callee, "->")) {
```
oxisto, line 404: I actually meant extracting the whole resolveOpreator, best viable, etc. stuff. isn't this 1:1 the same in both cases?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -322,6 +322,13 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
         }
     }
 
+    private fun isOverloadedOperator(ex: Expression, operatorCode: String): Boolean {
```
oxisto, line 325: a) some small kdoc would be nice
b) I would prefer that the name of the function reflects that this only checks for overloaded member expressions.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
peckto: Yes, basically, I was just thinking about the type update in the `resolveMember` case.
I now solved it as optional return type.
I don't see where the type should be used in the `handleCallExpression` case. Do you?

oxisto: Looks good, just add some documentation now and we're done :)

peckto: Added doc for function `resolveOverloadedArrowOperator`

oxisto (approved):

PR #1846, opened 2024-11-14
CXX frontend: remove assert

maximiliankaul:
There shouldn't be an assert in a frontend.


cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/ExpressionHandler.kt
```diff
@@ -429,7 +429,12 @@ class ExpressionHandler(lang: CXXLanguageFrontend) :
         when {
             reference is MemberExpression -> {
                 val baseType = reference.base.type.root
-                assert(baseType !is SecondOrderType)
+                if (baseType is SecondOrderType) {
```
oxisto, line 432: I think you can remove these two lines alltogether, because `baseType` is not used anymore.

oxisto (approved):

PR #1849, opened 2024-11-18
EOG: `when` instead of custom dispatcher

konradweiss:
Improves debugability by using `when` instead of custom handling map and dispatching function. Users have to order the when options such that more specific `is` check does not get shadowed by the more general check for its super class. However, modern IDEs can warn of such "unrechable" cases

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
@@ -117,86 +117,6 @@ open class EvaluationOrderGraphPass(ctx: TranslationContext) : TranslationUnitPa
      */
     protected val intermediateNodes = mutableListOf<Node>()
 
-    init {
-        map[IncludeDeclaration::class.java] = { doNothing() }
-        map[TranslationUnitDeclaration::class.java] = {
-            handleTranslationUnitDeclaration(it as TranslationUnitDeclaration)
-        }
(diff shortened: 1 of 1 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
```diff
@@ -117,86 +117,6 @@ open class EvaluationOrderGraphPass(ctx: TranslationContext) : TranslationUnitPa
...
-            handleCollectionComprehension(it as CollectionComprehension)
-        }
-        map[ComprehensionExpression::class.java] = {
-            handleComprehensionExpression(it as ComprehensionExpression)
-        }
-        map[LambdaExpression::class.java] = { handleLambdaExpression(it as LambdaExpression) }
-        map[LookupScopeStatement::class.java] = {
-            handleLookupScopeStatement(it as LookupScopeStatement)
-        }
-        map[ThrowExpression::class.java] = { handleThrowExpression(it as ThrowExpression) }
```
oxisto, line 197: I think you also need to remove the `map`

oxisto (approved):
LGTM except a minor thing

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
konradweiss: Whoops, forgot that one. Will do so!

PR #1850, opened 2024-11-18
Adding pages extracted from the wiki pages

konradweiss:
After this merge we can remove the wiki pages or hold only a link to our github pages there.

I removed all wiki pages that were already contained in some form in the github pages.


konradweiss:
> I think you have to add the new files to the structure of the github-page.
> 
> I'd also suggest to think about which information is already available and which one is new. There seems to be at least some overlap with the existing documentation

I already did remove obsolete pages, and added only two new pages or subpages as mentioned in the description, Maybe there is still some information that could be condensed or is slightly contained in other pages, but for now I would keep these two.

docs/docs/GettingStarted/query_examples.md
```diff
@@ -0,0 +1,150 @@
+# Query Examples
```
oxisto, line 1: This page was more an internal brainstorming rather than a public facing documentation, I would not include this in the public documentation or if we want this needs to be polished/cleaned. This mostly consists of random thoughts for now.

docs/docs/DesignPrinciples/index.md
```diff
@@ -0,0 +1,34 @@
+# Design Principles
```
oxisto, line 1: This is not really self-explanatory for an external doc page, especially if it is a "top-level" item, maybe at least put it under the "implementation" section.

oxisto (changes requested):

docs/docs/GettingStarted/query_examples.md
konradweiss: I removed the file from navigation but kept it in the docs, we can now accept this PR as is, and later think about whether or not we want to publish this Queries. But I think we have to look over them with the original authors and ensure that they are actually working

docs/docs/DesignPrinciples/index.md
konradweiss: I moved the design principles to the implementation subpages as suggested. I would not extend that declaration further but I am open to suggestions

PR #1853, opened 2024-11-19
Add more extensions to follow edges until the end

KuechA:
More homogeneous interfaces of the `followX...` extensions, and more functions which follow the edges until the end (without condition)


maximiliankaul (approved):
Code looks good to be. However, it'd be nice to have some tests... But we already talked about this and that it'd be a lot of work to come up with all the fluent dsl code -> I get why there aren't any tests right now...

PR #1854, opened 2024-11-19
WIP! first step: import checking extensions

maximiliankaul:
Still a lot of problems and discussions we should have:

- `someCallExpressionNode.isImported`: this should probably return whether the call is imported or not. However, if we think about `foo.bar.baz` or any other `MemberExpression`, then the current approach does not work.
- Do we want the end-user to figure out the `base` of the `MemberExpression` and be user friendly when providing `isImported`, or do we make the end-user do all the work and simply provide `isImported` on `Name`s?
- Only thinking about `MemberExpressions` is not sufficient. `Reference`s need to be considered, too.
- As well as C++ `using namespace` stuff...
- What do other languages do?

(Remember to test nested member expressions, too)


oxisto:
Replaced with #1889

PR #1855, opened 2024-11-19
change jep dependency from api to implementation

maximiliankaul:
No need to expose jep's API to CPG users.

### DIFF
--- a/cpg-language-python/build.gradle.kts
+++ b/cpg-language-python/build.gradle.kts
@@ -41,7 +41,7 @@ publishing {
 
 dependencies {
     // jep for python support
-    api(libs.jep)
+    implementation(libs.jep)
 
     // to evaluate some test cases
     testImplementation(project(":cpg-analysis"))


oxisto:
Are you sure about that? If one uses the python frontend, I am pretty sure they need JEP as a dependency and it won't be registered to them as a dependency unless its an `api` dependency.

maximiliankaul:
> Are you sure about that? If one uses the python frontend, I am pretty sure they need JEP as a dependency and it won't be registered to them as a dependency unless its an `api` dependency.

No, I'm not at all sure about this entire build system.
- I don't see why the jep API being exposed to the end-user would be of any benefit.
- Test (local and on GitHub) still work just fine.
- Using the CPG stand-alone (say cpg-console) works fine.

If there are any other tests you'd like me to do, then I'm happy for any pointers :)

oxisto:
> > Are you sure about that? If one uses the python frontend, I am pretty sure they need JEP as a dependency and it won't be registered to them as a dependency unless its an `api` dependency.
> 
> No, I'm not at all sure about this entire build system.
> 
> * I don't see why the jep API being exposed to the end-user would be of any benefit.
> * Test (local and on GitHub) still work just fine.
> * Using the CPG stand-alone (say cpg-console) works fine.
> 
> If there are any other tests you'd like me to do, then I'm happy for any pointers :)

Running from an external project still works. LGTM then

oxisto (approved):

PR #1857, opened 2024-11-20
Restrict symbol resolving to language

oxisto:
Currently, the symbol resolver runs wild and considers all nodes of all languages. When we have more instances where we have multiple languages in the same component, this craates a problem.

This PR now only considers the same language as the ref we are looking for.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
@@ -881,7 +881,7 @@ class ScopeManager : ScopeProvider {
      * arguments [scope] and [predicate] are forwarded.
      */
     fun lookupSymbolByNameOfNode(
-        node: HasScope,
+        node: Node,
         scope: Scope? = node.scope,
         predicate: ((Declaration) -> Boolean)? = null,
     ): List<Declaration> {
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
@@ -876,18 +876,16 @@ class ScopeManager : ScopeProvider {
     override val scope: Scope?
         get() = currentScope
 
-    /** A convenience function to call [lookupSymbolByName] with the properties of [node]. */
-    fun lookupSymbolByNameOfNode(node: HasNameAndLocation, scope: Scope? = currentScope): List<Declaration> {
-        return lookupSymbolByName(
-            node.name,
-            node.language,
-            node.location,
-            if (node is HasScope) {
-                node.scope
-            } else {
-                currentScope
-            }
-        )
+    /**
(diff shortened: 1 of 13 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -900,6 +902,18 @@ class ScopeManager : ScopeProvider {
     override val scope: Scope?
         get() = currentScope
 
+    /**
+     * A convenience function to call [lookupSymbolByName] with the properties of [node]. The
+     * arguments [scope] and [predicate] are forwarded.
+     */
+    fun lookupSymbolByNameOfNode(
+        node: HasNameAndLocation,
+        scope: Scope? = currentScope,
```
oxisto, line 911:
```suggestion
        scope: Scope? = node.scope,
```

oxisto:
Looking for review ;) @maximiliankaul @KuechA @konradweiss @lshala

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -590,7 +590,11 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
         var candidates = mutableSetOf<Declaration>()
         val records = possibleContainingTypes.mapNotNull { it.root.recordDeclaration }.toSet()
         for (record in records) {
-            candidates.addAll(ctx.scopeManager.lookupSymbolByName(record.name.fqn(symbol)))
+            candidates.addAll(
+                ctx.scopeManager.lookupSymbolByName(record.name.fqn(symbol), record.language) {
+                    it.language == record.language
```
KuechA, line 595: Is this additional predicate really required? Shouldn't this already be ensured by the `languageOnly` checks?

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/GoExtraPass.kt
```diff
@@ -389,8 +389,10 @@ class GoExtraPass(ctx: TranslationContext) : ComponentPass(ctx) {
 
         // Try to see if we already know about this namespace somehow
         val namespace =
-            scopeManager.lookupSymbolByName(import.name, null).filter {
-                it is NamespaceDeclaration && it.path == import.importURL
+            scopeManager.lookupSymbolByNameOfNode(import).filter {
+                it is NamespaceDeclaration &&
+                    it.path == import.importURL &&
+                    it.language == import.language
```
KuechA, line 395: Again, there's this additional language check in the predicate

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
oxisto: Ah this is a leftover before I introduced the additional argument, will remove these

PR #1859, opened 2024-11-20
replace all (where stuff doesn't break) api with implementation calls

maximiliankaul:
- [x] Why do we need to expose the following dependencies (changing to implementation breaks the build)?
  - [x] `ibs.apache.commons.lang3`
  - [x] `libs.neo4j.ogm.core`
- [x] Can we restrict cpg stuff? Does the cpg-console / neo4 API have to be exposed?


oxisto:
If you add `implementation(libs.apache.commons.lang3)` to every module that uses it, then it does not need to be exposed as `api`. I will add this to the buildSrc common convention

oxisto:
For neo4j it's tricker, it seems we are using a transitive dependency from neo4j (slf4j2) and we did not declare the usage of slf4j2 as a dependency. But we are using slf4j2's Logger as part of our public API, so we should instead specify slf4j2 as public API.

oxisto:
Ah, and we have a second problem. We are using neo4j's annotation in our public API, so we need to include at least the annotation package as "api", but not the complete core.

oxisto:
Ah, the annotations are part of core, so we are stuck with exposing OGM core

cpg-core/build.gradle.kts
```diff
@@ -52,13 +52,10 @@ tasks.test {
 }
 
 dependencies {
-    api(libs.apache.commons.lang3)
-    api(libs.neo4j.ogm.core)
-    api(libs.jackson)
+    api(libs.slf4j.api)
```
oxisto, line 55: Could probably also moved to common dependencies

oxisto (approved):

PR #1860, opened 2024-11-20
Fix C style function resolution

### LINKED ISSUE #1635: Call resolving for languages without function overloading
It might be, that we are a little bit to strict with the current call resolving.
We always check the parameter types, also for languages where no function overloading exists.
This will not map function calls, where one of the parameters have an `UNKNOWN` type.

This is the relevant code part:
https://github.com/Fraunhofer-AISEC/cpg/blob/fdc642b9c2b7e890e97f59a0eba154d48c05166e/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt#L719-L731

Theoretically, for the "not HasFunctionOverloading" case, we could finish here with an `==1`:
https://github.com/Fraunhofer-AISEC/cpg/blob/fdc642b9c2b7e890e97f59a0eba154d48c05166e/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt#L723-L725

Or we only take the shortcut, when one of the types is unknown.
What do you think?

Example where current call resolving fails:
```c
//#define MY_CONST_INT 1

void foo(int i) {
}

int main() {
    foo(MY_CONST_INT);
    return 0;
}
```

peckto:
Do not use signature based symbol resolving for languages without function overloading.
Fixes: #1635

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
@@ -517,6 +517,7 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
                 source.scope,
             )
         val language = source.language
+        val sourceCall = source as? CallExpression
 
         if (language == null) {
             result.success = CallResolutionResult.SuccessKind.PROBLEMATIC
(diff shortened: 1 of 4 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -527,29 +528,43 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
-                            arguments,
-                            source.language is HasDefaultArguments,
+        // Resolution depends on language features
+        if (source.language !is HasFunctionOverloading) {
+            // If the function does not allow function overloading, and we have multiple candidate
+            // symbols, the result is "problematic"
+            if (result.candidateFunctions.size > 1) {
+                result.success = CallResolutionResult.SuccessKind.PROBLEMATIC
+            } else
+            // If we have only one candidate function and the number of arguments match, we can take
```
oxisto, line 538: This has multiple issues
1) This does not only affect C but other languages as well.
2) It does not take into account if languages have default arguments

At the very least I think we need to make this configurable. Depending on the use case you MAY or MAY NOT want to do this. For example if you actually want to check if something is resolving similar to a compiler (e.g., also inference turned off), you want to have an empty list of invokes edges here.

If you want to be very optimistic about matching types, this may be another (maybe even global?) configuration option. You can always of course override the `tryCast` in a language that depends on the C language if you want to have a "matches-everything-C-language`.

So I am very skeptical about this change.

KuechA:
Double check the issue but the PR will be closed.

PR #1871, opened 2024-11-28
C: add names to gotos and labels

### LINKED ISSUE #1869: In some cases, the `GotoStatement` does not have a `targetLabel` or `labelName` set.
# Stack trace

n/a

# Unexpected behavior

I would expect that every `GotoStatement` has a non-empty `labelName` and `targetLabel` after the translation. However, we had some C/CPP-Code for which both were empty even if the code itself had some labels written.

# Code of interest

```
codeOfInterest
```

None

None

maximiliankaul:
Fixes #1869


cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXLanguageFrontendTest.kt
```diff
@@ -1758,4 +1757,23 @@ internal class CXXLanguageFrontendTest : BaseTest() {
         assertIs<CastExpression>(cast)
         assertLocalName("mytype", cast.castType)
     }
+
+    @Test
+    @Throws(Exception::class)
```
KuechA, line 1762: The `Throws` shouldn't be required, right?

KuechA (approved):

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXLanguageFrontendTest.kt
maximiliankaul: No, it's not. I just copied the above test to keep the same pattern. -> Removed.

PR #1897, opened 2024-12-12
Adding new Overlay and Concept nodes to the graph.

maximiliankaul:
Preparation for [mk/concepts](https://github.com/Fraunhofer-AISEC/cpg/tree/mk/concepts) This are the `cpg-core` affecting changes.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/OverlayNode.kt
```diff
@@ -0,0 +1,32 @@
...
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph
+
+/**
+ * Represents an extra node added to the CPG. These nodes can live next to the CPG, typically having
+ * shared edges to extend the original CPG graph.
+ */
+interface OverlayNode
```
oxisto, line 32: Should be an abstract class extending from Node

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/Concept.kt
```diff
@@ -0,0 +1,35 @@
...
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.concepts
+
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.OverlayNode
+
+/** A node extending the CPG and storing the corresponding/responsible [cpgNode]. */
+abstract class Concept : Node(), OverlayNode {
```
oxisto, line 32: Should only inherit from OverlayNode once it is an abstract class

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/ConceptNode.kt
```diff
@@ -0,0 +1,35 @@
...
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.concepts
+
+/**
+ * Represents a new concept added to the CPG. This is intended for modelling "concepts" like
+ * logging, files, databases. The relevant operations on this concept are modeled as
+ * [OperationNode]s and stored in [ops].
+ */
+abstract class ConceptNode() : Concept() {
```
oxisto, line 33: Why Concept and ConceptNode? We said we only want "Concept"

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/OperationNode.kt
```diff
@@ -0,0 +1,34 @@
...
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.concepts
+
+/**
+ * Represents an operation executed on/with a [ConceptNode] (stored in [concept]). This is typically
+ * a `write` on a file or log object or an `execute` on a database.
+ */
+abstract class OperationNode : Concept() {
```
oxisto, line 32: Only "Operation" instead of OperationNode. Should inherit from overlay node

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/ConceptNode.kt
maximiliankaul: Changed

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/OperationNode.kt
maximiliankaul: Changed

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/Operation.kt
```diff
@@ -0,0 +1,40 @@
...
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.OverlayNode
+
+/**
+ * Represents an operation executed on/with a [Concept] (stored in [concept]). This is typically a
+ * `write` on a file or log object or an `execute` on a database.
+ */
+abstract class Operation : OverlayNode() {
+    /** All concept nodes are connected to an original cpg [Node] by this. */
+    abstract val cpgNode: Node
```
konradweiss, line 37: I would name this node `astNode`, as it specifically connects to ast nodes and I would consider overlay nodes being cpg nodes to after they are added. I am also thinking of calling them `underlayNode` but that may be a little to weird.

konradweiss (commented):
Besides some renaming and the hierarchy change suggestion this is what I had in mind.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/Operation.kt
maximiliankaul: I'm not a big fan of `astNode` because we might want to connect one of the new nodes with something that's not part of the AST.
I named it `cpgNode` to point out that this is the original/classic/raw/... CPG node.
I'm happy to change this name once we have something better. `underlayingNode` would work for me.

oxisto: `underlayingNode` sounds good. And this property should be moved to the `OverlayNode`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/Operation.kt
```diff
@@ -0,0 +1,40 @@
...
+    /** The [Concept] this operation belongs to. */
+    abstract val concept: Concept
```
oxisto, line 39: Until we do more edge-magic, we should at least annotate this with `@Relationship` for neo4j persistence, the direction needs to be "incoming".
```suggestion
    abstract val concept: Concept
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/Operation.kt
maximiliankaul: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/collections/EdgeSingletonList.kt
```diff
@@ -36,7 +36,11 @@ import org.neo4j.ogm.annotation.Transient
  *
  * Therefore, we need to wrap the edge in a list with a single element.
  */
-class EdgeSingletonList<NodeType : Node, NullableNodeType : NodeType?, EdgeType : Edge<NodeType>>(
+open class EdgeSingletonList<
```
konradweiss, line 39: Will we be able to get rid of this completely now that we have our own persistency component?

konradweiss (commented):
LGTM. The only remarks I have is that I would prefer the edges to go from the overlay node to the underlay node, but that is more because it was always like this in my imagination. And then we would need a different edge label so it is not really something we have to change for this PR.

oxisto:
> LGTM. The only remarks I have is that I would prefer the edges to go from the overlay node to the underlay node, but that is more because it was always like this in my imagination. And then we would need a different edge label so it is not really something we have to change for this PR.

I can just switch the direction, if you want

oxisto:
> > LGTM. The only remarks I have is that I would prefer the edges to go from the overlay node to the underlay node, but that is more because it was always like this in my imagination. And then we would need a different edge label so it is not really something we have to change for this PR.
> 
> I can just switch the direction, if you want

AH, no that doesn't work. It needs to be outgoing from an AST node, otherwise it will not end up in the persisted graph. I will keep it as is for now.

PR #1907, opened 2024-12-18
Persisting literal values more conservativly

oxisto:
Fixes issues with python interface files


konradweiss (approved):
LGTM, If we see that there is a type we have not considered in this distinction that is not supposed to run into the default case, we can then add that distinction there.

PR #1862, opened 2024-11-21
Fixing Java field access

oxisto:
This PR consists of a major overhaul of Java reference and member expression parsing. It removes a LOT of old weird legacy code that was full of weird things.

It also introduces a new Java-specific pass, that takes care of deciding when a member access is actually static or not. This might be something that we could decide to adapt for other/all languages (maybe also tied to the discussion in #1863).

PR #1908, opened 2024-12-18
Providing an API to update the imported symbols for an `ImportDeclaration`

oxisto:
This introduces a new extension function `ImportDeclaration.updateImportedSymbols`. The idea behind that is that we might need to update the `importedSymbols` of an import declaration, if the symbols that "live" in the associated namespace change. This is mainly the case once we infer records out of known `Type` nodes. Therefore, the type resolver will update all the imported symbols before executing all the remaining passes. This way, the remainder of the passes can access imported symbols that were inferred.

Needed for #1862


konradweiss (approved):
LGTM as it only moves the function that updates the symbols and is called an additional time in the type resolver

PR #1909, opened 2024-12-19
Provide a function for languages to influence reference resolution out of candidates

oxisto:
This adds a new function `Language.bestViableReferenceCandidate` which takes the old implementation as a default behaviour.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
@@ -33,12 +33,15 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize
 import de.fraunhofer.aisec.cpg.*
 import de.fraunhofer.aisec.cpg.graph.Name
 import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.declarations.Declaration
 import de.fraunhofer.aisec.cpg.graph.declarations.FunctionDeclaration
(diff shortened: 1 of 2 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -361,6 +364,29 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
...
+     * certainty.
+     *
+     * If we encounter an ambiguous result, a warning is issued.
+     */
+    open fun bestViableReferenceCandidate(ref: Reference): Declaration? {
+        return if (ref.candidates.size > 1) {
+            Util.warnWithFileLocation(
+                ref,
+                log,
+                "Resolution of reference {} was ambiguous, cannot set refersTo correctly.",
```
konradweiss, line 382: I would mention what is used when this warning is printed. Here `null`.

konradweiss (approved):

konradweiss:
Now the default behavior changed from being `singleOrNull` in all cases to to one candidate if the list of candidates contains exactly one element and `null` otherwise. I hope this was the intention

oxisto:
> Now the default behavior changed from being `singleOrNull` in all cases to to one candidate if the list of candidates contains exactly one element and `null` otherwise. I hope this was the intention

The default behavior is unchanged, but a language can override it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
oxisto: What do you mean exactly?

PR #1918, opened 2025-01-04
Skip C++ neo4j tests if C++ frontend is not enabled

oxisto:
This PR skips the C++-based neo4j tests, if the C++ frontend is not enabled. We should also add more language-neutral tests to neo4j in the future.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ on:
 
 jobs:
   build:
-    runs-on: self-hosted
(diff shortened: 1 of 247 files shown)


konradweiss (approved):
LGTM

oxisto:
Thanks. I will keep it un-merged until we discuss it tomorrow.

oxisto:
We will move the C++ tests in neo4j to integration tests instead and exclude integration and performance test from `build`.

PR #1921, opened 2025-01-07
Fix/test eog for loop

konradweiss:
The missing context when creating nodes in the parameter list of the fluent dsl nodes caused them to be added at the wrong statement holders.


oxisto (approved):
I have no clue whether the adjustments to the EOG pass are correct, but since you are the EOG master I assume they are ;) Rest looks reasonable.

PR #1931, opened 2025-01-14
Update concepts

maximiliankaul:
This PR makes the `underlyingNode` `open` so that we can easily use it when implementing such nodes.

Additionally, the `Concept` and `Operation` nodes are removed from `cpg-core` as they will be part of a follow-up pr (https://github.com/Fraunhofer-AISEC/cpg/pull/1932) adding them to a `cpg-concepts` module.


maximiliankaul:
Closed in favor of https://github.com/Fraunhofer-AISEC/cpg/pull/1932

PR #1940, opened 2025-01-16
Basic implementation of `compliance scan` command

oxisto:
This is a basic implementation of the compliance scan command. It will output an example SARIF finding, but will do an actual CPG translation.

The syntax is the following: `codyze compliance --project-dir [project-directory] --sources [your-sources]`

This will do an actual CPG translation and print a "note" SARIF finding on the console


codyze-compliance/src/main/kotlin/de/fraunhofer/aisec/codyze/compliance/Command.kt
```diff
@@ -0,0 +1,112 @@
...
+ * A command that operates on a project. This class provides the common options and functions for
+ * all commands.
+ */
+abstract class ProjectCommand : CliktCommand() {
+    private val projectOptions by ProjectOptions()
+    private val translationOptions by TranslationOptions()
+
+    /** Loads the security goals from the project. */
+    fun loadSecurityGoals(): List<SecurityGoal> {
+        return loadSecurityGoals(projectOptions.directory.resolve("security-goals"))
```
KuechA, line 48: I'm not the biggest fan of these hardcoded paths and would prefer passing an argument in the CLI

codyze-compliance/src/main/kotlin/de/fraunhofer/aisec/codyze/compliance/Command.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * This method is called by the `run` method to perform the actual analysis. It is separated to
+     * allow for easier access from overriding applications.
+     */
+    protected fun analyze(): AnalysisResult {
+        // Load the security goals from the project
+        val goals = loadSecurityGoals(projectOptions.directory.resolve("security-goals"))
```
KuechA, line 57: I'm not the biggest fan of these hardcoded paths and would prefer passing an argument in the CLI

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Project.kt
```diff
@@ -0,0 +1,143 @@
...
+data class AnalysisResult(val translationResult: TranslationResult, val run: Run)
+
+/** Analyzes the project and returns the result. */
+fun analyze(config: TranslationConfiguration): AnalysisResult {
+    val run =
+        Run(
+            tool = Tool(driver = ToolComponent(name = "Codyze", version = "x.x.x")),
+            results =
+                listOf(
+                    Result(
```
KuechA, line 83: This seems a leftover or some sort of preparation for a TODO and should receive a comment

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Project.kt
```diff
@@ -0,0 +1,143 @@
...
+
+/** Builds a translation configuration from the given CLI options. */
+fun buildConfig(
+    projectOptions: ProjectOptions,
+    translationOptions: TranslationOptions,
+): TranslationConfiguration {
+    var builder =
+        TranslationConfiguration.builder()
+            .defaultPasses()
+            .optionalLanguage("de.fraunhofer.aisec.cpg.frontends.python.PythonLanguage")
```
KuechA, line 105: Can't we enable the other languages as well?

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Project.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    // We can either have a single source (using --sources) or multiple components (using
+    // --components)
+    translationOptions.sources?.let {
+        builder =
+            builder
+                .sourceLocations(translationOptions.sources!!.map { it.toFile() })
```
KuechA, line 112: We should avoid `!!`

codyze-core/src/main/resources/META-INF/kotlin/script/templates/de.fraunhofer.aisec.codyze.QueryScript.classname
KuechA, line 1: What is this file doing?

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/QueryHost.kt
```diff
@@ -0,0 +1,62 @@
...
+import de.fraunhofer.aisec.cpg.helpers.Benchmark
+import java.io.File
+import kotlin.reflect.full.functions
+import kotlin.script.experimental.api.*
+import kotlin.script.experimental.host.toScriptSource
+import kotlin.script.experimental.jvmhost.BasicJvmScriptingHost
+import kotlin.script.experimental.jvmhost.createJvmCompilationConfigurationFromTemplate
+import kotlin.script.experimental.jvmhost.createJvmEvaluationConfigurationFromTemplate
+
+fun TranslationResult.evalQuery(scriptFile: File, queryFunc: String): Any? {
```
KuechA, line 38: This method requires some documentation

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/QueryScriptDefinition.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlin.script.templates.ScriptTemplateDefinition
+
+@ScriptTemplateDefinition(scriptFilePattern = ".*\\.query\\.kts")
+@KotlinScript(
+    // File extension for the script type
+    fileExtension = "query.kts",
+    // Compilation configuration for the script type
+    compilationConfiguration = QueryScriptConfiguration::class,
+)
+abstract class QueryScript
```
KuechA, line 43: All classes and objects in this file need a documentation

codyze-core/src/main/resources/META-INF/kotlin/script/templates/de.fraunhofer.aisec.codyze.QueryScript.classname
oxisto: Magic. It is needed because of kotlin scripting foo

codyze-compliance/src/main/kotlin/de/fraunhofer/aisec/codyze/compliance/Command.kt
oxisto: My idea for the future would be to replace the security-goals.yaml with a project.yaml or something similar, which then has the "authority" to change these paths, I am not sure if they are really suitable for CLI args. My suggestion is to extract these hardcoded paths into a Project structure (which is already prepare) so we can make them dynamic in the future.

codyze-compliance/src/main/kotlin/de/fraunhofer/aisec/codyze/compliance/Command.kt
oxisto: see above

codyze-compliance/src/main/kotlin/de/fraunhofer/aisec/codyze/compliance/Command.kt
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * The `list-security-goals` command. This will list the names of all security goals in the
+ * specified project.
+ *
+ * This command assumes that the project contains a folder named `security-goals` that contains YAML
+ * files with the security goals.
+ */
+class ListSecurityGoals : ProjectCommand() {
```
fwendland, line 105: I wouldn't inherit from `ProjectCommand` because I currently see no reason to specify a project dir or CPG translation options. We're also not calling `analyze()`.

Should we use use a `CliktCommand` for now?

oxisto: We are calling `loadSecurityGoals` which loads the security goals in the project specified by the project options, so we are using the project dir.

fwendland: Missed that one 😔 

It might still be confusing to have the CPG Translation Options in the subcommand for `list-security-goals`.

oxisto: They don't hurt at the moment and I don't want to introduce too many different commands.

codyze-compliance/src/main/kotlin/de/fraunhofer/aisec/codyze/compliance/Command.kt
fwendland: I'm fine with leaving them hardcoded for now. We could add a reminder similar to [Project.kt#L101](https://github.com/Fraunhofer-AISEC/cpg/blob/25a7565d3085d5cc3e4d43b40d5a22b914f93409/codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Project.kt#L101).

If we decide to add a comment, we may also add it here: [Project.kt#L90-L94](https://github.com/Fraunhofer-AISEC/cpg/blob/25a7565d3085d5cc3e4d43b40d5a22b914f93409/codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Project.kt#L90-L94).

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/QueryScriptDefinition.kt
```diff
@@ -0,0 +1,79 @@
...
+        compilerOptions("-Xcontext-receivers", "-jvm-target=17")
+        defaultImports.append(
+            "de.fraunhofer.aisec.codyze.*",
+            "de.fraunhofer.aisec.cpg.*",
+            "de.fraunhofer.aisec.cpg.graph.*",
+            "de.fraunhofer.aisec.cpg.query.*",
+        )
+        ide { acceptedLocations(ScriptAcceptedLocation.Everywhere) }
+    }) {
+    private fun readResolve(): Any = QueryScriptConfiguration
```
fwendland, line 78: Is this another 'magic' element needed for Kotlin Scripting? Or can we remove it?

oxisto: It seems so, my IntelliJ complaint about some serialisation stuff

KuechA (approved):

fwendland (approved):

PR #1941, opened 2025-01-16
Handling `ImportDeclaration` in EOG

oxisto:
This silences a lot of warnings. We are using `handleDefault` instead, which adds it to the EOG, but does not add any children nodes.


KuechA (approved):
The change seems ok and might be required for some languages and I guess it doesn't harm connecting the `ImportDeclarations` with the EOG. However, I was wondering if the connection to the EOG should be configured by a language trait?

PR #1944, opened 2025-01-16
Pruning of if-branches in Python based on constant evaluation

oxisto:
Python is a dynamically interpreted language. This means that (function) declarations are declared dynamically when the code is actually executed rather than at compile time. That also means that it is possible to declare different versions of declarations in conditional branches, which makes analysis hard. This is frequently used in the type shed project to decide declarations based on the Python version and platform.

A typical example looks like this:

```Python
if sys.platform == "win32":
  def win32_func():
     pass

if sys.version_info < (3, 9):
  def old_python_func():
    pass
```

This PR does two things:
a) It replaces occurrences of `sys.platform` and `sys.version_info` with the actual literal value in an evaluation
b) It checks in if-statements, whether we can constantly evaluate the condition. Note that this is *very limited* to literals and operations on literals. We cannot decide this based on references, since variables are not resolved at this point.

### DIFF
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
@@ -36,6 +36,7 @@ import de.fraunhofer.aisec.cpg.graph.types.AutoType
 import de.fraunhofer.aisec.cpg.graph.types.Type
 import de.fraunhofer.aisec.cpg.helpers.CommentMatcher
 import de.fraunhofer.aisec.cpg.passes.PythonAddDeclarationsPass
+import de.fraunhofer.aisec.cpg.passes.UnreachableEOGPass
 import de.fraunhofer.aisec.cpg.passes.configuration.RegisterExtraPass
 import de.fraunhofer.aisec.cpg.sarif.PhysicalLocation
 import de.fraunhofer.aisec.cpg.sarif.Region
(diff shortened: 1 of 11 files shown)


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
```diff
@@ -76,9 +83,26 @@ class PythonLanguageFrontend(language: Language<PythonLanguageFrontend>, ctx: Tr
             it.set("content", fileContent)
             it.set("filename", file.absolutePath)
             it.exec("import ast")
+            it.exec("import sys")
             it.exec("parsed = ast.parse(content, filename=filename, type_comments=True)")
 
             val pyAST = it.getValue("parsed") as PyObject
+
+            // Populate system information
+            SysInfo.platform = it.getValue("sys.platform") as? String
```
KuechA, line 92: I'm not sure at all if this is really what we want. This checks the platform and python version which we use to analyze/parse the code which can be completely different from the environment which is used to execute the code. This is likely to introduce problems.

I'd prefer to pass such information to the CPG via arguments, configuration files or similar (as it's done below I guess?) which represent the "execution environment" of a certain program (if such information is available). IMHO it's better to have no information here instead of wrong assumptions.

Furthermore, I'd expect to find the SysInfo object somewhere in the translation result since these are some assumptions which may be useful for further analysis.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -728,18 +735,39 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+        // the programs to declare different symbols. This is frequently used in the Python typeshed
+        // to declare different declarations based on the python version and the platform in a way
+        // that C/C++ would use preprocessor macro. Therefore, we replace occurrences of
+        // sys.platform and sys.version_info by their actual values when we encounter the
+        // expression.
+        //
+        // It would be too complex (=impossible) for _all_ if statements to be evaluated and decided
+        // which branch would be executed at runtime. But we can have a simple heuristic that
+        // evaluates simple constant expressions.
+        val constCondition = ret.condition?.evaluate(constEvaluator) as? Boolean
```
KuechA, line 749: I really dislike that we move such logic into the frontend. In my understanding, the frontends were meant to be simple and have a "stupid" translation of AST nodes to the CPG's AST version. If we now decide to remove some code, perform some preliminary evaluations and so on, this contradicts my understanding of the goal. I also thought that the CPG would keep all available information in the code and not remove parts of the code (in particular not in the frontend and without any way to disable this behavior).

Why don't we use a pass to declare this branch as unreachable and, based on this, remove the affected declarations (or also flag the refersTo/usages/... edges as inappropriate)? This would be more clean in my opinion and make it easier to understand the final result.

KuechA (changes requested):

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
oxisto: Agree, fetching this from here was my initial approach and then I added the "environment" below, it probably makes sense to only have the environment symbols here.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
maximiliankaul: 100% agree. An optional pass executing early would be the preferred solution from my point of view, too. I do not like this in the frontend.

oxisto: I copy/pasted the old unreachable EOG pass for now, unfortunately this will not work on a large code base. Seems this is stuck until #1811

oxisto: In the meantime I will split some of the things in this PR into smaller ones.

oxisto:
Closed in favor of #1951

PR #1945, opened 2025-01-17
Add Concept and Operation for HTTP APIs and their calls

lshala:
This PR adds the Concepts and Operations for `HTTP` API and call nodes to the `cpg-concepts` module.


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/api/ApiBuilder.kt
```diff
@@ -0,0 +1,87 @@
...
+    httpMethod: String,
+    arguments: List<Node>,
+    concept: RestApiConcept,
+): RestApiOperation {
+    val apiOperation =
+        RestApiOperation(
+            underlyingNode = underlyingNode,
+            concept = concept,
+            httpMethod =
+                when (httpMethod) {
```
lshala, line 58: I'm not sure if this handling is too specific here and should rather be e.g. in the passes

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/api/RestApiConcept.kt
```diff
@@ -0,0 +1,44 @@
...
+import de.fraunhofer.aisec.cpg.graph.concepts.Concept
+
+/**
+ * Represents a REST API concept. This concept models an API endpoint with a specific role (e.g.,
+ * consumer or provider).
+ *
+ * @param apiUrl The URL of the API endpoint.
+ * @param role The role of the API either [ApiRole.CONSUMER] or [ApiRole.PROVIDER].
+ */
+class RestApiConcept(underlyingNode: Node, val apiUrl: String, val role: ApiRole) :
```
oxisto, line 38: Should we include the `Concept` suffix for these nodes or not? It is probably not needed.

For me it is not entirely clear "what" exactly this is. I guess on the server-side it represents an an API endpoint or an API handler? We could probably look at the ontology we have in the cloud property graph since we already modeled similar things there.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/api/RestApiOperation.kt
```diff
@@ -0,0 +1,55 @@
...
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.concepts.api
+
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.concepts.Operation
+
+/**
```
oxisto, line 31: Similar to the concept: Is this an API handling operation or an API call from the client side or both? Maybe we need to extract them into different concepts?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/api/RestApiOperation.kt
```diff
@@ -0,0 +1,55 @@
...
+ * @param concept The [RestApiConcept] this operation is associated with.
+ * @property invokes A list of other [RestApiOperation]s that this operation invokes.
+ */
+class RestApiOperation(
+    underlyingNode: Node,
+    val httpMethod: HttpMethod,
+    val arguments: List<Node>,
+    override val concept: RestApiConcept,
+) : Operation(underlyingNode = underlyingNode, concept = concept) {
+    val invokes: MutableList<RestApiOperation> = mutableListOf()
```
lshala, line 46: As a RestApiOperation is supposed to represent an API call, the question came up during the [PR](https://gitlab.cc-asp.fraunhofer.de/bsi-653/openstack-checker/-/merge_requests/3#note_2826702) whether the invokes property should be used to connect the corresponding calls on the concepts "layer" instead of the invokes edges of the underlying nodes. Any thoughts on this?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/api/RestApiConcept.kt
lshala: It is meant to represent an API endpoint either from the client-side (Consumer) or the server-side (Provider). The API call itself is represented by the Operation. Would client/server be better?

I would change it to RestApi and the Operation to ApiCall?

oxisto: > It is meant to represent an API endpoint either from the client-side (`Consumer`) or the server-side (`Provider`). The API call itself is represented by the Operation. Would `Client`/`Server` be better?

Yes

> 
> I would change it to `RestApi` and the Operation to `ApiCall`?

Yes, I am also thinking about changing this to `HttpCall` or `HttpRequest`. There are so many different API-styles and not all of them are necessarily 100 % "REST", but at least they all share a common http layer.

In the cloud property graph paper (https://ieeexplore.ieee.org/document/9582243) we had the following concepts / operations and maybe it would be good as an orientation:
- `HttpRequestHandler` is a graph node that represents a collection of different endpoints that are bundled together and serve as one "api" (e.g. `/v1/secrets`). It has a (prefix) path and a collection of `HttpEndpoint` nodes (see below) and it was usually connected to a handler structure such as a controller (class)
- `HttpEndpoint` is a graph node that represents one particular endpoint (e.g., PUT `/v1/secrets/{id}`). It has the HTTP operation (PUT, GET, etc.), a path, arguments and is connected to its handler function / method
- `HttpRequest` is a graph node that represents a (client) request operation to an `HttpEndpoint`. They are connected to a call expression and are also resolved and connected to their matching `HttpEndpoint`.

The challenge is that in this case we had 2 concepts (on the server side) and only 1 operation (on the client side), so this might be adopted, but still I think this would be a good starting point.

oxisto (approved):

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/http/HttpEndpoint.kt
```diff
@@ -0,0 +1,51 @@
...
+import de.fraunhofer.aisec.cpg.graph.concepts.Concept
+import de.fraunhofer.aisec.cpg.graph.concepts.Operation
+
+/** Represents a single [HttpEndpoint] on the server */
+class HttpEndpoint(
+    underlyingNode: Node,
+    val httpMethod: HttpMethod,
+    val path: String,
+    val arguments: List<Node>,
+) : Concept<HttpEndpointOperation>(underlyingNode = underlyingNode)
```
oxisto, line 38: We want to add a property "isTLS" and "supportedAuthentications"

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/http/HttpEndpoint.kt
```diff
@@ -0,0 +1,51 @@
...
+
+enum class HttpMethod {
+    GET,
+    POST,
+    PUT,
+    DELETE,
```
maximiliankaul, line 44:
```suggestion
    PATCH,
    OPTIONS,
    CONNECT,
    TRACE,
    DELETE,
```

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/http/HttpRequest.kt
```diff
@@ -0,0 +1,37 @@
...
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.concepts.http
+
+import de.fraunhofer.aisec.cpg.graph.Node
+
+/** Represents an [HttpRequest] from the client. */
```
maximiliankaul, line 30:
```suggestion
/** Represents an [HttpRequest] from the [HttpClient]. */
```

maximiliankaul (approved):
LGTM. Please add the new properties we just discussed.

PR #1946, opened 2025-01-17
Adding options to handle unreachable EOG branches in the `SymbolResolver`

oxisto:
This PR adds two configuration option to the `SymbolResolver`:
    - `skipUnreachableEOG` skips unreachable EOG paths when iterating the EOG. This means that symbols on unreachable EOG paths are not resolved. This can speed up resolution when large parts of a file are not reachable (e.g., because they are guarded by a platform switch)
    - `ignoreUnreachableDeclarations` will check, whether a declaration that is a candidate for a symbol is on an EOG path that is unreachable. This is useful in dynamically invoked languages, where symbols are declared differently on different paths.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/processing/strategy/Strategy.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/processing/strategy/Strategy.kt
@@ -51,7 +51,7 @@ object Strategy {
     }
 
     /** A strategy to traverse the EOG in forward direction, but only if the edge is reachable. */
-    fun nextReachableEOG(x: Node): Iterator<Node> {
+    fun REACHABLE_EOG_FORWARD(x: Node): Iterator<Node> {
         return x.nextEOGEdges.filter { !it.unreachable }.map { it.end }.iterator()
     }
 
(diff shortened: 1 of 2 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/processing/strategy/Strategy.kt
```diff
@@ -50,6 +50,11 @@ object Strategy {
         return x.nextEOG.iterator()
     }
 
+    /** A strategy to traverse the EOG in forward direction, but only if the edge is reachable. */
+    fun nextReachableEOG(x: Node): Iterator<Node> {
```
KuechA, line 54: should the name be harmonized with the rest? Like `REACHABLE_EOG_FORWARD`?

KuechA (approved):
The changes seem to be ok but we have to improve the unreachable eog pass to profit from this solution

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/processing/strategy/Strategy.kt
oxisto: In theory yes, although the uppercase naming is really bad. But we probably should rename all at the same time.

PR #1947, opened 2025-01-17
    Extending value evaluator in the python frontend

oxisto:
This PR extends the value evaluator with python specifics. This also makes the `cpg-analysis` mandatory for the python language frontend.

It also makes some necessary adjustments in the value evaluator itself, especially dealing with initializers.


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonValueEvaluator.kt
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+
+    override fun handleReference(expr: Reference, depth: Int): Any? {
+        // We need to handle sys.platform and sys.version_info specially, since it is often used in
+        // a pre-processor macro-style, and we want to replace this with the actual value (if we
+        // have it). This allows us to dynamically prune if-branches based on constant evaluation.
+        return when {
+            expr.reconstructedImportName.toString() == "sys.platform" ->
+                PythonLanguageFrontend.SysInfo.platform ?: super.handleReference(expr, depth)
```
KuechA, line 60: Could we place this information as a node (maybe via `Concept`?) somewhere in the `TranslationResult` or probably better in the `Component`? Even if we're in the python module, I somehow would prefer accessing this information through the graph.

This also affects the next lines of course.

oxisto: Hmm good point. I will think about a solution.

oxisto: I added the `SystemInformation` as an overlay node to the current TU(s)

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/SystemInformation.kt
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+
+/**
+ * Represents different system information that are used in the [PythonValueEvaluator] to evaluate
+ * expressions, such as `sys.platform` and `sys.version_info`.
+ *
+ * We model this as an overlay node so we can access this information later in the graph.
+ */
+data class SystemInformation(var versionInfo: VersionInfo? = null, var platform: String? = null) :
```
KuechA, line 57: This is not necessarily python specific but describes something like the runtime system of a program, so I'd say it can be moved to `cpg-concept`. Same for `VersionInfo` (maybe call it `RuntimeVersion` or so to not confuse it with the version of the program we analyze?)

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/SystemInformation.kt
```diff
@@ -0,0 +1,58 @@
...
+package de.fraunhofer.aisec.cpg.frontends.python
+
+import de.fraunhofer.aisec.cpg.graph.OverlayNode
+
+/** Represents the contents of `sys.version_info` which contains the Python version. */
+data class VersionInfo(var major: Long? = null, var minor: Long? = null, var micro: Long? = null) :
+    OverlayNode() {
+    /**
+     * Returns the version info as a tuple (major, minor, micro). The length of the tuple depends on
+     * the information set, e.g., if only major version is set, then the tuple is 1 element long.
```
KuechA, line 35:
```suggestion
     * the information set, e.g., if only major version is set, then the list is 1 element long.
```

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/SystemInformation.kt
KuechA: We may move these classes to `cpg-core` even if this also has drawbacks. But for now, I'd also be fine keeping them in python only

PR #1950, opened 2025-01-18
Faster type and symbol resolution

oxisto:
This PR speeds up type and symbol resolution across the whole code-base

* Optimized `extractScope` by using the `fqnScopeMap` instead of filtering ALL scopes.
* Removed unnecessary nested loop over all nodes/types again in `TypeManager`
* Faster type/scope merging in the `TranslationManager` the using parallel parsing (8ms instead of 5s)
* Do not try to resolve member expressions of an unknown base type. This will fail and will only result in a lot of spam and in the attempts to resolve the "UNKNOWN" scope (which does not exist)

This boosted my test case (parsing python type shed stdlib) from ~70s to ~~45s~~ 32s

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeResolver.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeResolver.kt
@@ -29,12 +29,12 @@ import de.fraunhofer.aisec.cpg.ScopeManager
 import de.fraunhofer.aisec.cpg.TranslationContext
 import de.fraunhofer.aisec.cpg.TypeManager
 import de.fraunhofer.aisec.cpg.graph.*
-import de.fraunhofer.aisec.cpg.graph.declarations.ImportDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.RecordDeclaration
 import de.fraunhofer.aisec.cpg.graph.types.DeclaresType
 import de.fraunhofer.aisec.cpg.graph.types.ObjectType
 import de.fraunhofer.aisec.cpg.graph.types.Type
 import de.fraunhofer.aisec.cpg.graph.types.recordDeclaration
+import de.fraunhofer.aisec.cpg.helpers.Benchmark
 import de.fraunhofer.aisec.cpg.helpers.SubgraphWalker
 import de.fraunhofer.aisec.cpg.passes.configuration.DependsOn
 import de.fraunhofer.aisec.cpg.passes.inference.tryRecordInference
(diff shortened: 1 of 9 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeResolver.kt
```diff
@@ -135,29 +137,17 @@ open class TypeResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
     override fun cleanup() {
         // Nothing to do
     }
 
     /** Resolves all types in [TypeManager.firstOrderTypes] using [resolveType]. */
     fun resolveFirstOrderTypes() {
         for (type in typeManager.firstOrderTypes.sortedBy { it.name }) {
-            if (type is ObjectType && type.typeOrigin == Type.Origin.UNRESOLVED) {
+            if (
+                type is ObjectType && type.typeOrigin == Type.Origin.UNRESOLVED ||
```
KuechA, line 148: Didn't you miss brackets here around the `||` expression?

KuechA: In line 94, there's a similar check. Should this change also apply there?

oxisto: The brackets should not be necessary from a logic point since it’s evaluated left to right of view but I can add them for clarity 

I can adjust the check in line 94 as well although in practice we only use „guessed“ in the old
Java fronted and that does not have typedefs 

we could problably just remove „guessed“ and make it unresolved as well

KuechA: Ok... I thought you want to have `type is ObjectType && (type.typeOrigin == Type.Origin.UNRESOLVED || type.typeOrigin == Type.Origin.GUESSED)`. Are there no GUESSED types which are not ObjectTypes?

I thought the origin GUESSED was introduced just recently mainly to clarify some things?

oxisto: > Ok... I thought you want to have `type is ObjectType && (type.typeOrigin == Type.Origin.UNRESOLVED || type.typeOrigin == Type.Origin.GUESSED)`. Are there no GUESSED types which are not ObjectTypes?
> 
> I thought the origin GUESSED was introduced just recently mainly to clarify some things?

 No, the GUESSED enum is a left over that is only used in the old Java frontend. And all of them are object types

KuechA: Ok. And in contrast we can have unresolved non-object types?

oxisto: No, I don’t think so

KuechA: Then, the check is unnecessary, isn't it?

KuechA (approved):
My comment is somewhat resolved/discussed. So, I won't block this PR.

PR #1951, opened 2025-01-19
Using `UnreachableEOGPass` in python

oxisto:
This PR adds a special version of the `UnreachableEOGPass` to the python frontend. It is executed before the symbol resolver and uses the python value evaluator.

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
@@ -29,7 +29,6 @@ import de.fraunhofer.aisec.cpg.TranslationContext
 import de.fraunhofer.aisec.cpg.analysis.ValueEvaluator
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.declarations.FunctionDeclaration
-import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
 import de.fraunhofer.aisec.cpg.graph.edges.flows.EvaluationOrder
 import de.fraunhofer.aisec.cpg.graph.statements.DoStatement
 import de.fraunhofer.aisec.cpg.graph.statements.ForStatement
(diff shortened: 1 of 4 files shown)


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
```diff
@@ -59,142 +62,143 @@ class UnreachableEOGPass(ctx: TranslationContext) : TranslationUnitPass(ctx) {
      * @param node every node in the TranslationResult
      */
     protected fun handle(node: Node, parent: Node?) {
-        if (node is FunctionDeclaration) {
-            val startState = UnreachabilityState()
-            for (firstEdge in node.nextEOGEdges) {
-                startState.push(firstEdge, ReachabilityLattice(Reachability.REACHABLE))
-            }
-            val finalState = iterateEOG(node.nextEOGEdges, startState, ::transfer) ?: return
+        val startState = UnreachabilityState()
```
KuechA, line 65: We should not run this for every node. The SubgraphWalker will iterate the whole graph and we should select some meaningful entry points for the analysis. So far, these were function declarations. It may make sense to use the first EOG starter node in the EOG instead but not all nodes in the graph

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
```diff
@@ -63,6 +64,7 @@ import kotlin.math.min
  * a value (on the first assignment).
  */
 @RegisterExtraPass(PythonAddDeclarationsPass::class)
+@RegisterExtraPass(PythonUnreachableEOGPass::class)
```
maximiliankaul, line 67: Do we want to have this pass on by default? Why isn't it part of `defaultPasses` (the non-Python version obviously) and then Python replaces it with this?
I'm just wondering if this is a future we only want to have active by default for Python?

maximiliankaul (commented):
The Python-part LGTM

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
oxisto: I changed the pass already to an eog starter pass so this is actually doing exactly that :)

KuechA: Right, my fault

KuechA (approved):

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
oxisto: Well since this pass is massively slow (at the moment), we probably do not want it active per-default. So we can either keep this PR open as blocked or I will remove the register extra pass and if someone wants it, they can activate it.

KuechA: We won't activate this pass by default.

PR #1956, opened 2025-01-21
Concept and Operation nodes for disk encryption

maximiliankaul:


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/diskEncryption/CipherOp.kt
```diff
@@ -0,0 +1,32 @@
...
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.concepts.diskEncryption
+
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.concepts.Operation
+
+abstract class CipherOp(underlyingNode: Node, override val concept: Cipher) :
```
oxisto, line 31: For consistency with other classes  I would prefer `Operation` instead of `Op` even though it is slightly longer

oxisto (approved):

Issue #1089, opened 2023-02-08
Support multiple return values


maximiliankaul:
The CPG should support multiple return values per statement. This is useful for language constructs like `a, b = (1, 2)` or `value, error := someFunction()`.
Note, that such constructs can also happen in unexpected places like loops `for a, b, c in ...` where we would have three loop variables in this example.

KuechA:
The expressions supporting multiple values will model the targets with lists.

Tasks:
* [ ] Replace the `=` Binary Operator with a dedicated `AssignmentExpression` class.
* [ ] Adapt the `variable` in the `ForEachStatement`.
* [ ] The support for multiple return types/values of functions has to be expanded (see TODOs in code)

Tricky problems:
* (1) `a, b = foo()`: The function call should be modeled only once and magically be "expanded"
* (2) `int i, a, b = 1;`: Not all declared variables are initialized
* (3) Various ways to initialize a variable in C++ should be represented in the same way. The different options should probably be kept in some properties.
* (4) Mixture of `DeclaredReferenceExpressions` and `VariableDeclarations` (e.g. `a` is an already defined variable and used in `a, b = 1,2`): This probably needs handling refs and decls at the same time
* (5) It should be easy to identify uninitialized variables (= write a query)

Discussed solutions:
1) Split multiple assignments into multiple nodes. Problems: 1
2) Assignment statement with lists for `lhs`, `rhs`, `declarations`. Problems: 1, 2 (keeping the lists synchronized). Needs to change the VariableDeclaration to get rid of the initializer

peckto:
If we treat multiple return values as a container, we could add an implicit "unpack" node in situations where the return values are unpacked to separate values.
The "unpack" functionality is probably language specific and could be implemented in separate pass.

oxisto:
@maximiliankaul Is this issue still relevant? I think we implemented this in python using our assign expression?

maximiliankaul:
> @maximiliankaul Is this issue still relevant? I think we implemented this in python using our assign expression? Or at least it should be done once #1729 is merged.

I think it's no longer relevant. I'll double check that we have tests for all the corner cases discussed above and then close this issue asap.

maximiliankaul:
There are still some bugs. See #1807 for fixes.

PR #1961, opened 2025-01-22
Support tuple deconstruction and follow value flows based on index

KuechA:
* [x] Add functionality to split up the DFG of a statement like `(a, b) = ...` where the edges are labeled with a new `index` granularity.
* [x] Add test
* [x] Try with and without brackets
* [x] Use this also in other scenarios (tuples in a return value)
* [x] Special follow DFG functions which have a stack of indices and only select matching tuple elements for unpacking

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
@@ -28,10 +28,12 @@ package de.fraunhofer.aisec.cpg.graph
 import de.fraunhofer.aisec.cpg.TranslationResult
 import de.fraunhofer.aisec.cpg.graph.declarations.*
 import de.fraunhofer.aisec.cpg.graph.edges.Edge
+import de.fraunhofer.aisec.cpg.graph.edges.flows.IndexedDataflowGranularity
 import de.fraunhofer.aisec.cpg.graph.statements.*
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.*
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.Block
 import de.fraunhofer.aisec.cpg.helpers.SubgraphWalker
+import java.util.ArrayDeque
 import kotlin.math.absoluteValue
 
 /**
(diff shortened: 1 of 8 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/Dataflow.kt
```diff
@@ -63,6 +63,15 @@ class PartialDataflowGranularity(
     val partialTarget: Declaration?
 ) : Granularity
 
+/**
+ * This dataflow granularity denotes that not the "whole" object is flowing from [Dataflow.start] to
+ * [Dataflow.end] but only parts of it. Common examples include tuples or array indices.
+ */
+class IndexedDataflowGranularity(
```
maximiliankaul, line 70: Will we then add a third class for partial struct access (e.g. field, offset)?

KuechA: The field-access is already modeled by the `PartialDataflowGranularity` which would receive the `FieldDeclaration` as an input. The offset could probably be used here too (putting the offset as an int as an input). For non-constant offsets, we have nothing matching yet but we can think about this in the future

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -122,18 +124,25 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : EOGStarterPass
                 }
             } else {
                 value.elements.forEach {
+                    var callingContext: CallingContext? = null
```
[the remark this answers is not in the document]
KuechA: I added a comment on this

KuechA:
We cover the basic cases of #1089 but miss some of the tests in  #1807, in particular, we do not split up strings into characters nor do we handle the case of the `*` unpacking operator which seems rather specific to python.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -228,6 +230,92 @@ fun Node.collectAllPrevFullDFGPaths(): List<List<Node>> {
...
+                currentNode.prevDFG
+            }
+        },
+        collectFailedPaths = collectFailedPaths,
+        findAllPossiblePaths = findAllPossiblePaths,
+        predicate = predicate,
+    )
+}
+
+class SimpleStack<T>() {
```
oxisto, line 297: Could be private. I am a little bit confused by this is necessary. doesn't ArrayDeque already implement Stack?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/Dataflow.kt
oxisto: They are probably all of the same type to some degree (tuple deconstruct, array index, struct index) so we should combine them at some point and also model the missing data flows. For example we don't model them for array subscription expressions at the moment.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
```diff
@@ -200,6 +201,31 @@ class PythonAddDeclarationsPass(ctx: TranslationContext) : ComponentPass(ctx), L
                     assignExpression.declarations += handled
                 }
             }
+            (target as? InitializerListExpression)?.let {
+                it.initializers.forEach { initializer ->
+                    (initializer as? Reference)?.let {
+                        it.access = AccessValues.WRITE
+                        val handled = handleWriteToReference(initializer)
+                        if (handled is Declaration) {
+                            // We cannot assign an initializer here because this will lead to
```
oxisto, line 210: Duplicate code from above, can you please refactor this into a separate method? also the line breaks in the comments are off.

oxisto (approved):
Looks good in general, minor details needed. We need to have a discussion about our different non-full dataflow types at some point but for now this is the best solution to our problem. Please make the PR title a little more meaningful ;)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: It does but I wanted to limit the functionality to the minimum. It also doesn't provide a single function for checking and removing the first value on the stack if I remember this right.

KuechA: I would not make the class private in case a user wants to implement an own follow Something function with a stack

maximiliankaul: I'd go with `ArrayDeque` and not implement stuff that's already provided by the language. You could have a `checkAndPop` extension to `ArrayDeque`. I think that's cleaner than re-implementing a stack. But I'm also ok if you want to keep this class.

PR #1964, opened 2025-01-23
Add On Import Translation of Python Files in Include Path

konradweiss:
The PR adds a preprocessing function to all languages that is supposed to return files located in the includePaths that are supposed to be analyzed under their own component. 

The current implementation for python identifies which file in the import paths should be used to fulfill which import statement and then recursively resolves additional imports in that file.

External folders can be added to `includePaths` and when turning it on with `loadIncludes` the folders are expanded into all files under those paths. The frontends then have access to all `externalSources` over `ctx.externalSources`. The frontend is supposed to then select which files should be added to `ctx.importedSources` and the translation manager then analyzes these as additional files.

In this PR I implemented this feature for the python language frontend to select the importedSources based on the import statements that would access their contained Records, fields and functions.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
```diff
@@ -140,15 +142,24 @@ private constructor(
         result: TranslationResult,
     ): Set<LanguageFrontend<*, *>> {
         val usedFrontends = mutableSetOf<LanguageFrontend<*, *>>()
+
+        val externalSources: MutableList<File> = mutableListOf()
```
maximiliankaul, line 146: Can we please have some doc here? What are these variables used for?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
```diff
@@ -252,9 +287,40 @@ private constructor(
...
+                )
+            }
+        }
+
         return usedFrontends
     }
 
+    private fun extractConfiguredSources(path: Path): MutableList<File> {
+        val rootFile = path.toFile()
+        return if (rootFile.exists())
```
maximiliankaul, line 318: Please refactor this code. Add some `{` as a start.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
```diff
@@ -326,6 +324,112 @@ class PythonLanguageFrontend(language: Language<PythonLanguageFrontend>, ctx: Tr
...
+    ): List<File> {
+
+        val pythonExternalSources =
+            externalSources
+                .filter { externalFile ->
+                    language.fileExtensions.any { externalFile.path.endsWith(it) }
+                }
+                .toMutableList()
+
+        val importRe = "(?m)^(?:from[ ]+(\\S+)[ ]+)?import[ ]+(\\S+)[ ]*\$".toRegex()
```
maximiliankaul, line 348: I object having this regex here.
- It's not documented.
- I don't like parsing the sources manually with a regex -> can't we use the `ast.parse` result for this?
- Can we please name groups, if we have to use a regex? This would make it much easier to understand what's going on.
- What about `import foo` without the `from` part?
- What about `as` in the import?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
```diff
@@ -326,6 +324,112 @@ class PythonLanguageFrontend(language: Language<PythonLanguageFrontend>, ctx: Tr
...
+        val allImportedSources: MutableList<File> = mutableListOf()
+        importRe.findAll(source.readText()).forEach { imp ->
+
+            // Only try to find file containing imports if we did not process this import so far
+            if (!processedImports.contains(imp.groupValues.get(0))) {
```
maximiliankaul, line 353: I don't think this approach works. An `from foo ...` import can refer to different imports, depending on the location of the file you're parsing here.

maximiliankaul (changes requested):
I think there is some discussion needed before merging this into `main`. See internal chat + comments here.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
konradweiss: I removed the regex approach.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
konradweiss: Imports are now analyzed based on the import handling in the `StatementHandler`. So from import is now also handled in the same way as we do for the generation of the import declaration.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
konradweiss: Added doc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
konradweiss: Did so too

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
```diff
@@ -470,6 +472,23 @@ object Util {
         conditionNodes.forEach { n.prevDFGEdges += it }
     }
 
+    /**
+     * Returns the RootPath that belongs to the absolute file path in `file` amongst the
```
oxisto, line 476: Please  use  the correct dokka syntax, like `[file]`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
```diff
@@ -470,6 +472,23 @@ object Util {
...
+     * `rootPaths`.
+     *
+     * @param file absolute file to find the rootPath for
+     * @param rootPaths candidates of root paths for the file
+     * @return the first identified root path candidate
+     */
+    fun getRootPath(file: File, rootPaths: List<Path>): Path {
```
oxisto, line 483: `fun File.rootPath(rootPaths: List<Path>)` would be more kotlin-like. Also there is no real need to put it into the `Util` object, we can just have it as a top-level func.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
oxisto: And the documentation is not accurate, since it will return the parent directory of the file if it does not "belong" to the root path. I would also suggest to use the wording "is relative to" instead of belonging.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
```diff
@@ -470,6 +472,23 @@ object Util {
...
+        rootPaths.forEach {
+            if (file.path.removePrefix(it.toString()) != file.path.toString()) {
```
oxisto, line 485: You can use `relativeToOrNull`  like this
```kotlin
return rootPaths.firstOrNull { file.relativeToOrNull(it.toFile()) != null }
            ?: file.toPath().parent
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -397,6 +397,8 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
         }
     }
 
+    abstract fun isBuiltinsFile(file: File): Boolean
```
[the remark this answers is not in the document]
oxisto: Can't we just have this as `return false` as default?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
```diff
@@ -228,6 +229,12 @@ class PythonLanguage :
         return super.tryCast(type, targetType, hint, targetHint)
     }
 
+    override fun isBuiltinsFile(file: File): Boolean {
+        return fileExtensions.any {
+            file.path == IDENTIFIER_INIT + language.namespaceDelimiter + it
```
oxisto, line 234: This confuses me. Why is `__init__.py` a built-in ? Also we should not use the namespace delimiter as the file extension "dot"

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -615,12 +620,40 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
         return declStmt
     }
 
+    private fun addExternallyImportedToAnalysis(importName: Name) {
```
oxisto, line 623: Missing documentation. What is "externally" imported?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -615,12 +620,40 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+        ctx?.let { ctx ->
```
oxisto, line 624: I would suggest checking ctx for null and then throw a translation exception instead of having this nested.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -615,12 +620,40 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+            var currentName = importName
+            while (currentName.isNotEmpty()) {
+                var importPath = currentName.toString().replace(language.namespaceDelimiter, "/")
+
+                language.fileExtensions.forEach { fileExtension ->
+
+                    // Includes a file in the analysis, if it has the root path and
+                    ctx.externalSources
+                        .firstOrNull {
```
oxisto, line 633: why only the first? And why are we not looking at the include paths? The external sources set confuses me a bit.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -615,12 +620,40 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+                            val relPath =
+                                it.relativeTo(
```
oxisto, line 635: you are basically relativizing the path twice, once inside getRootPath to get the root path and then again on the root path. You can just to the relativeOrNull check here and directly use the result if its non-null

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -615,12 +620,40 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+                                        Util.getRootPath(it, ctx.config.includePaths).toFile()
+                                    )
+                                    .path
+                            val ending = "." + fileExtension
+                            relPath == importPath + ending ||
+                                relPath ==
+                                    importPath + "/" + PythonLanguage.IDENTIFIER_INIT + ending
```
oxisto, line 642: should use File.separator

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -615,12 +620,40 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+                        }
+                        ?.let { ctx.importedSources += it }
+                }
+                currentName = currentName.parent ?: Name("")
```
oxisto, line 646: creating a new empty name is uncecessary, you can just use null instead and do a nil check in the while loop.

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/LoadIncludesTest.kt
```diff
@@ -0,0 +1,50 @@
...
+import de.fraunhofer.aisec.cpg.test.analyze
+import java.nio.file.Path
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class LoadIncludesTest {
+
+    @Test
+    fun testLoadIncludes() {
+        val topLevel = Path.of("src", "test", "resources", "python", "loadIncludes", "example")
```
oxisto, line 38: `load_includes` instead of `loadIncludes` would be more python-esque

cpg-language-python/src/test/resources/python/loadIncludes/stdlib/__init__.pyi
```diff
@@ -0,0 +1,2 @@
+class str(Sequence[str]):
```
oxisto, line 1: The builtins for python are in the `builtins` package. https://docs.python.org/3/library/builtins.html

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationContext.kt
```diff
@@ -56,4 +57,17 @@ class TranslationContext(
...
+     * of analysis but are for example the expanded folder structure of
+     * [TranslationConfiguration.includePaths].
+     */
+    var externalSources: MutableSet<File> = mutableSetOf(),
+    /**
+     * The additional sources from the [externalSources] chosen to be analyzed along with the code
+     * under analysis. The language frontends are supposed to fill this list, e.g. by analyzing
+     * their import statements.
+     */
+    var importedSources: MutableSet<File> = mutableSetOf(),
```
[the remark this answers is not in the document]
konradweiss: I heavily dislike the idea to parse all external sources if we can make it dependent on the once that are actually used. If we ever analyze a program where we include several external sources this could lead to a bloat in files to be analyzed. A frontend still CAN import all sources if it wants to but I would not forego the functionality to import only what is needed.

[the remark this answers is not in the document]
oxisto: I just don't understand why we need two sets or what the difference between the two is.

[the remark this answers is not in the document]
konradweiss: ah. So the user provides includePaths. All files contained in those paths are stored in `externalSources`. Then, the frontend looks at the import and the `externalSources` and decides whether or not one of the files should be analyzed. If it should, it is added to importedSources, which here acts as a worklist. The expansion from `includePaths` to `externalSources` happens such that we don't have to do many `file.exists()` operations.

What we could also do here is to reduce the time the external sources have to be checked, e.g. :
```
ctx.externalSources -= it // e.g. add this line
ctx.importedSources += it

```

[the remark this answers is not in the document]
oxisto: Hm maybe I get a better feeling for this once I implement this in Go, but at least the description you just gave here makes it clearer and should be inserted into either the description of externalSources or importedSources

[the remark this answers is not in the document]
konradweiss: Was this before or after I added the comments two days ago? Because it contains this description. I now am confused if you reviewed after or before I added that documentation. Maybe the comments I already provided seem only explanatory to me :/

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
```diff
@@ -140,16 +143,26 @@ private constructor(
         result: TranslationResult,
     ): Set<LanguageFrontend<*, *>> {
         val usedFrontends = mutableSetOf<LanguageFrontend<*, *>>()
+
+        // Without extracting the sources from the include paths to the external sources the feature
```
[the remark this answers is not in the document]
konradweiss: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
```diff
@@ -140,16 +143,26 @@ private constructor(
...
+
+        var useParallelFrontends = ctx.config.useParallelFrontends
+
         for (sc in ctx.config.softwareComponents.keys) {
             val component = Component()
             component.ctx = ctx
             component.name = Name(sc)
             result.addComponent(component)
 
+            // Every Component needs to reprocess the sources
```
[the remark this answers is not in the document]
konradweiss: removed

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
```diff
@@ -251,10 +265,72 @@ private constructor(
...
+        // A set of processed files from external sources that is used as negative to the worklist
+        // in ctx.importedSources
+        // it is used to filter out files that were already processed and to detect if new files
+        // were analyzed.
+        val processedExternalSources: MutableList<File> = mutableListOf()
+
+        do {
+            val oldProcessedSize = processedExternalSources.size
+            // Distribute all files by there root path prefix, parse them in individual component
+            // named
```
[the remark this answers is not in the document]
konradweiss: Done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
konradweiss: thanks, I used that instead

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
konradweiss: rewrote this using in two places to remove double raltivizing path and removed the now unnecessary utility function.

cpg-language-python/src/test/resources/python/loadIncludes/stdlib/__init__.pyi
konradweiss: I now moved the builtin file selection to a trait, and based it on all analyzed files. The resolution of the functions in the builtins does not work anymore as we decided to do that mapping in the SymbolSolver in another PR

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/LoadIncludesTest.kt
konradweiss: done

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
konradweiss: Removed that Name creation and changed the loop condition.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationContext.kt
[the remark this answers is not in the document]
oxisto: maybe the part about “for example from the include paths” was confusing me. We should clearly state here which function of which class is responsible of filling in the sources.

cpg-language-python/src/test/resources/python/loadIncludes/stdlib/__init__.pyi
oxisto: I unresolved this because I fear we need another solution / method of the interface. For the symbol resolver we need to know which namespace contains the built-in. Can we model it this way that the function returns a Name/String and we then resolve the bultins.py based on the name “bultins”.

konradweiss: Yes, this should be doable, one issue that we still have is that the extension may be ".py" or ".pyi" or even something else in the future if we build our own library representing file format. I will add the file extensions in the translation manager.

oxisto: > Yes, this should be doable, one issue that we still have is that the extension may be ".py" or ".pyi" or even something else in the future if we build our own library representing file format. I will add the file extensions in the translation manager.

But don't we already have the logic to translate a "package name" into a "filename" (in the python frontend), can't we just invoke that for the built-in name?

konradweiss: I could incorporate that function if you could point me to it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationContext.kt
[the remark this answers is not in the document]
konradweiss: I hope I now clarified my comments more explizitly

cpg-language-python/src/test/resources/python/loadIncludes/stdlib/__init__.pyi
konradweiss: I moved the trait to provide the builtin namespace instead of a boolean function that decides if a file is a builtin function. I also extracted the logic to decide if a file is a fitting candidate for a namespace to another function and used that to identify the builtins file and files, relevant when analyzing an import.

oxisto: Sounds good, only small comment about function vs. variable.

oxisto (approved):
LGTM except one last comment and it needs a rebase/merge because of a conflict in the python frontend. 

Any feedback from @maximiliankaul on this?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -425,6 +425,19 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
 
         return tu
     }
+
+    /**
+     * Returns the files that can represent the given name considering all viable file extensions.
+     */
+    open fun nameToLanguageFiles(name: Name): Set<File> {
```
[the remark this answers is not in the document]
konradweiss: This is not supposed to work without any change for every language. the purpose is to have a general convention on how files are named that have to be identified by the name of a namespace. If we ever build a set of library summaries or typing files we may have to reconsider this. I could add this reliance on a naming heuristic for files to this comment?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
```diff
@@ -228,6 +231,21 @@ class PythonLanguage :
...
+    override fun builtinsNamespace(): Name {
+        return Name("builtins")
+    }
+
+    /**
+     * Returns the files that can represent the given name. This includes all possible file
+     * extensions and the name plus the `__init__` identifier, as this is the name for declaration
+     * files if the namespace has sub-namespaces.
+     */
+    override fun nameToLanguageFiles(name: Name): Set<File> {
```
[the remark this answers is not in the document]
oxisto: This is something that we generally do not support (monkey patching), so I think the PR within the current limitations but does not expand them.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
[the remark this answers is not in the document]
oxisto: > This is not supposed to work without any change for every language. the purpose is to have a general convention on how files are named that have to be identified by the name of a namespace. If we ever build a set of library summaries or typing files we may have to reconsider this. I could add this reliance on a naming heuristic for files to this comment?

The issue is that basically python is the only language where the assumption you present in this function holds. So we should probably at least make the default implementation of this "null" and then then python language can override this.

What I don't understand is why we need this function exposed in the language in the first place. Looking at the occurrences where this is used it seems that we have code that "transfers" stuff from `externalSources` to `importedSources` in two places:
- In python's addExternallyImportedToAnalysis
- In the translation managers `runFrontends`

Since the description says the frontend is responsible for moving things from external to important, why are we doing it in the translation manager in the first place? We can probably remove the second occurrence and have this function internally to python.

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/LoadIncludesTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        val result =
+            analyze(listOf(topLevel.resolve("program.py").toFile()), topLevel, true) {
+                it.registerLanguage<PythonLanguage>()
+                it.loadIncludes(true)
+                it.includePath(
+                    Path.of("src", "test", "resources", "python", "load_includes", "stdlib")
+                )
+            }
+
+        assertEquals(result.ctx?.importedSources?.size, 3)
```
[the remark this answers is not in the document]
oxisto: Fun fact, if the code from the translation manager is removed (see discussion above), the value here is actually 2.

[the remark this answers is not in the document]
oxisto: > Fun fact, if the code from the translation manager is removed (see discussion above), the value here is actually 2.

The third is the builtin.pyi. After discussion with @konradweiss we will do that in a separate PR.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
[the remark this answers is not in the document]
oxisto: After discussion with @konradweiss the intuition of the first one was to parse builtins, but we will do that in a separate PR since the current solution has some issues.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
```diff
@@ -344,15 +342,15 @@ class PythonLanguageFrontend(language: Language<PythonLanguageFrontend>, ctx: Tr
                 }
             }
 
-        if (lastNamespace != null) {
+        (lastNamespace ?: tud).let {
```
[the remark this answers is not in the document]
konradweiss: Not required anymore, I used this for the `__init__.py` at rootlevel, but I removed this in the stdlib typings. So I will revert this back

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationManager.kt
```diff
@@ -140,6 +139,19 @@ private constructor(
         result: TranslationResult,
     ): Set<LanguageFrontend<*, *>> {
         val usedFrontends = mutableSetOf<LanguageFrontend<*, *>>()
+        val usedLanguages = mutableSetOf<Language<*>>()
+
+        // If loadIncludes is active, the files stored in the include paths are made available for
+        // conditional
```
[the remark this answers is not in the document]
konradweiss: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationContext.kt
```diff
@@ -56,4 +57,21 @@ class TranslationContext(
...
+
+    /**
+     * Set of files, that are available for additional analysis. They are not the primary subjects
+     * of analysis but are available to the language frontend. The files are obtained by expanding
+     * the paths in [TranslationConfiguration.includePaths].
+     *
+     * The frontend can decide to add some of the contained files to [importedSources] which will
+     * get them translated into the final graph by the [TranslationManager].
+     */
+    var externalSources: MutableSet<File> = mutableSetOf(),
```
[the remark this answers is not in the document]
konradweiss: renamed

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/LoadIncludesTest.kt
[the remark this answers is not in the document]
konradweiss: removed the builtins include per default

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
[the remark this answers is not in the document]
konradweiss: Moved to python language exclusively

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/StatementHandler.kt
```diff
@@ -641,12 +647,42 @@ class StatementHandler(frontend: PythonLanguageFrontend) :
...
+        val ctx = ctx
+        if (ctx == null) {
+            throw TranslationException(
+                "A translation context is needed for the import dependent addition of external sources."
+            )
+        }
+        var currentName: Name? = importName
+        while (!currentName.isNullOrEmpty()) {
+
+            // Includes a file in the analysis, if it has the root path and
```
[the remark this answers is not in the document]
konradweiss: completed the comment

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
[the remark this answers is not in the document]
konradweiss: @maximiliankaul I now provided a comment and did not remove this functionality as it should should still be a good change.

PR #1965, opened 2025-01-23
Keep argument name information when converting member call expressions to call expressions

oxisto:
Previously, this information was lost because we were only copying/moving the nodes and not the edges

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/Extensions.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/Extensions.kt
@@ -80,3 +80,10 @@ fun <
     val edge = edgeProperty.call(this)
     return edge.delegate()
 }
+
+/** Returns the first edge with the given [name] or `null` if no such edge exists. */
+operator fun <NodeType : Node, EdgeType : Edge<NodeType>> EdgeList<NodeType, EdgeType>.get(
+    name: String
+): EdgeType? {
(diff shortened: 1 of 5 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/Extensions.kt
```diff
@@ -80,3 +80,10 @@ fun <
     val edge = edgeProperty.call(this)
     return edge.delegate()
 }
+
+/** Returns the first edge with the given [name] or `null` if no such edge exists. */
+operator fun <NodeType : Node, EdgeType : Edge<NodeType>> EdgeList<NodeType, EdgeType>.get(
```
maximiliankaul, line 85: Is this new function used in this PR at all? :confused:

oxisto: Yes it it. https://github.com/Fraunhofer-AISEC/cpg/pull/1965/files#diff-71de8469911ec3c4f5162303939f8730ee3e8513676d0b125218e58beb113b16R1724

maximiliankaul: Oops. Ok.

maximiliankaul (approved):
I do not like the use of `.first`. I think we should always consider all matches and print a warning if there is more than one thing and our code only considers the first one.

Otherwise I'm fine with the pr.

PR #1969, opened 2025-01-24
Extend concepts

KuechA:
It may be useful to extend the concepts as follows:

* Add property in HttpRequest to link it to an HttpEndpoint


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/http/HttpRequest.kt
```diff
@@ -34,4 +34,6 @@ class HttpRequest(
     val arguments: List<Node>,
     val httpMethod: HttpMethod,
     override val concept: HttpClient,
-) : HttpClientOperation(underlyingNode = underlyingNode, concept = concept)
+) : HttpClientOperation(underlyingNode = underlyingNode, concept = concept) {
+    val calls = mutableListOf<HttpEndpoint>()
```
oxisto, line 38: In the ontology we called this `to`. Maybe better suitable?

KuechA: renamed

oxisto (approved):

PR #1973, opened 2025-01-24
Do not infer field declarations in static methods

KuechA:
The `PythonAddDeclarationsPass` infers a `FieldDeclaration` even if we're in a static function. This is not really correct. We now infer a `VariableDeclaration` instead.


oxisto (approved):
Can we have a short test, if this works correctly?

PR #1979, opened 2025-01-24
Lookup for matching `ParameterDeclaration`s before generating new variables

### LINKED ISSUE #1978: `PythonAddDeclarationsPass` ignores `ParameterDeclarations` and does not resolve symbols to them
In the following example
```py
def foo(a):
    a = bar(a)
    return a
```
all three `Reference`s in the function body have the `refersTo` set to a `VariableDeclaration` but they ignore that the symbol is already defined by the `ParameterDeclaration`.

In contrast., the `Reference`s refer to the `ParameterDeclaration` in the case of
```py
def foo(a):
    b = bar(a)
    return a
```

KuechA:
Fixes #1978 

We could also split the big when-statement up into multiple parts.

### DIFF
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
@@ -85,10 +85,10 @@ class PythonAddDeclarationsPass(ctx: TranslationContext) : ComponentPass(ctx), L
     }
 
     /**
-     * This function will create a new dynamic [VariableDeclaration] if there is a write access to
-     * the [ref].
+     * This function will create a new dynamic [Declaration] if there is a write access to the
+     * [ref].
      */
-    private fun handleWriteToReference(ref: Reference): VariableDeclaration? {
+    private fun handleWriteToReference(ref: Reference): Declaration? {
         if (ref.access != AccessValues.WRITE) {
             return null
         }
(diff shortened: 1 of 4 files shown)


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt
```diff
@@ -155,12 +155,28 @@ class PythonAddDeclarationsPass(ctx: TranslationContext) : ComponentPass(ctx), L
         // If we didn't create any field up to this point and if we are still have not returned, we
         // can create a normal variable. We need to take scope modifications into account.
         var decl =
-            if (field == null && targetScope != null) {
-                scopeManager.withScope(targetScope) { newVariableDeclaration(ref.name) }
-            } else if (field == null) {
-                newVariableDeclaration(ref.name)
-            } else {
-                field
+            when {
```
oxisto, line 158: This is not a good solution to the problem. Actually the problem lies elsewhere. The code in the original method is correct, but the scoping is "wrong".

`refA` in your example case lives in a "Function Scope", where as the current scope at the time of the resolution is a "Local Scope". Therefore, the lookup in line 115 is not successful. Actually both should be in the same scope since Python does not have any local scopes, only function scopes. So we could either try to set the correct scope (which is probably not that easy), or we need to adapt the scope filter in line 115. But I am strongly against this solution here that introduces two additional lookups that are unnecessary.

oxisto (changes requested):

PR #1984, opened 2025-01-25
Do not resolve unqualified symbols in current record without `HasImplicitReceiver`

### LINKED ISSUE #1983: Python incorrectly resolves fields in unqualified lookup
Currently our python frontend (and probably others?) resolve fields when doing an unqualified lookup, i.e., just looking for `symbol` (but not `self.symbol`). But that is not allowed in Python (as well as in Go actually) and this leads to horrendous wrong results.

Consider the following snippet:

```Python
a = "Hello"

class MyClass:
    def foo(self):
        self.a = 1
        print(a)
```

The `a` in the print statement should resolve to the global `a`. Instead it currently resolves to the field declaration of `a`, which is only accessible through `self.a`.

This also applies to imports as well 
```Python
import os

class MyClass:
    def bar(self):
        self.os = 1
        print(os.name)
```

Here, `os` is also wrongfully resolved to `MyClass.os`.

This might be a bigger issue and effecting all language frontends that require some sort of "qualified" lookup.

oxisto:
We need to consider the trait `HasImplicitReceiver` (or the lack thereof) when resolving symbols in an unqualified lookup to not accidentally resolve fields/methods of a class without an receiver.

Fixes #1983

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
@@ -176,9 +176,7 @@ sealed class Scope(
                 } else {
                     // If our language needs explicit lookup for fields (and other class members),
                     // we need to skip record scopes unless we are in a qualified lookup
-                    if (
-                        languageOnly is HasExplicitReceiverOnly && scope.parent is RecordScope
-                    ) {
+                    if (languageOnly is HasExplicitReceiverOnly && scope.parent is RecordScope) {
                         scope.firstParentOrNull { it !is RecordScope }
                     } else {
                         // Otherwise, we can just go to the next parent
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
@@ -26,12 +26,11 @@
 package de.fraunhofer.aisec.cpg.graph.scopes
 
 import com.fasterxml.jackson.annotation.JsonBackReference
-import de.fraunhofer.aisec.cpg.frontends.HasExplicitMemberAccess
+import de.fraunhofer.aisec.cpg.frontends.HasExplicitReceiverOnly
 import de.fraunhofer.aisec.cpg.frontends.Language
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.declarations.Declaration
 import de.fraunhofer.aisec.cpg.graph.declarations.ImportDeclaration
-import de.fraunhofer.aisec.cpg.graph.firstParentOrNull
 import de.fraunhofer.aisec.cpg.graph.statements.LabelStatement
 import de.fraunhofer.aisec.cpg.graph.statements.LookupScopeStatement
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.Reference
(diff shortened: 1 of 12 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
```diff
@@ -227,6 +239,20 @@ sealed class Scope(
             list += value
         }
     }
+
+    fun firstParentOrNull(predicate: (Scope) -> Boolean): Scope? {
```
KuechA, line 243: `firstParentScopeOrNull` to not confuse it with the `astParent`?

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/SymbolResolverTest.kt
```diff
@@ -0,0 +1,65 @@
...
+        assertNotNull(a)
+
+        val fieldA = result.records["MyClass"]?.fields["a"]
+        assertNotNull(fieldA)
+
+        val aRefs = result.refs("a")
+        aRefs.filterIsInstance<MemberExpression>().forEach { assertRefersTo(it, fieldA) }
+        aRefs.filter { it !is MemberExpression }.forEach { assertRefersTo(it, a) }
+
+        val osRefs = result.refs("os")
```
KuechA, line 59: A short comment would be helpful here to understand that `os.name` does not have a reference any more because this used to be different/wrong until recently

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
oxisto: Was my original idea as well. I can change it.

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/SymbolResolverTest.kt
oxisto: Well it was always wrong, but yes it makes sense to comment the unit test.

oxisto:
After some digging, I actually discovered that we already have the opposite trait `HasImplicitReceiver`, so we should probably use the existing one instead.

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/SymbolResolverTest.kt
```diff
@@ -0,0 +1,72 @@
...
+                it.registerLanguage<PythonLanguage>()
+            }
+
+        val a =
+            result.namespaces["fields"]
+                .variables[{ it.name.localName == "a" && it !is FieldDeclaration }]
+        assertNotNull(a)
+
+        val fieldA = result.records["MyClass"]?.fields["a"]
+        assertNotNull(fieldA)
```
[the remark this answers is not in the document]
oxisto: Unncessary in this case, because we already know that fieldA is a field declaration (because `fields` returns a Field Declaration), `assertIs` is only necessary if the variable `fieldA` would be a "Node" or a "Declaration" and we want to assert its a field.

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/SymbolResolverTest.kt
```diff
@@ -0,0 +1,90 @@
...
+
+        val aRefs = result.refs("a")
+        aRefs.filterIsInstance<MemberExpression>().forEach { assertRefersTo(it, fieldA) }
```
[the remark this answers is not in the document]
oxisto: As discussed offline. We cannot fix the `copyA = a` problem, so I adjusted this line for now and created issue https://github.com/Fraunhofer-AISEC/cpg/issues to track it.

The rest works after adjustment in the python add declaration pass, which was missing an FQN.

PR #1986, opened 2025-01-25
Fix unity build

peckto:
#1960 introduced a bug when using the unity build mode without software components.
I'm not sure why the `run`  is need for the else case of the `let`, but it fixes the problem.
I also added a test-case for the unity build mode (with and without software components).


oxisto (approved):
Interesting, no idea why the previous one didn't work. 🤷

PR #1988, opened 2025-01-26
Basic implementation of following context-sensitive calls in `followNextDFGUntilHit`

oxisto:
I am not sure if this really solves *all* the problems, but at least this tries to use the information that we have in the calling context in/out so that we are properly following dataflows into functions and out of functions using the correct call (stack).

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
@@ -28,6 +28,9 @@ package de.fraunhofer.aisec.cpg.graph
 import de.fraunhofer.aisec.cpg.TranslationResult
 import de.fraunhofer.aisec.cpg.graph.declarations.*
 import de.fraunhofer.aisec.cpg.graph.edges.Edge
+import de.fraunhofer.aisec.cpg.graph.edges.flows.CallingContextIn
+import de.fraunhofer.aisec.cpg.graph.edges.flows.CallingContextOut
+import de.fraunhofer.aisec.cpg.graph.edges.flows.ContextSensitiveDataflow
 import de.fraunhofer.aisec.cpg.graph.edges.flows.IndexedDataflowGranularity
 import de.fraunhofer.aisec.cpg.graph.statements.*
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.*
@@ -579,7 +582,7 @@ fun Node.followPrevCDGUntilHit(
  * not mandatory**. If the list "failed" is empty, the path is mandatory.
  */
 inline fun <reified T> Node.followXUntilHit(
-    x: (Node, SimpleStack<T>?) -> Collection<Node>,
+    noinline x: (Node, SimpleStack<T>?) -> Collection<Node>,
     collectFailedPaths: Boolean = true,
     findAllPossiblePaths: Boolean = true,
     stack: SimpleStack<T>? = null,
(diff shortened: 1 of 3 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -1231,3 +1291,12 @@ val Expression.isImported: Boolean
     get() {
         return this.importedFrom.isNotEmpty()
     }
+
+private fun <T> ArrayDeque<T>.removeIfFirst(element: T): Boolean {
```
KuechA, line 1295: Isn't this feature already provided by `SimpleStack`? Is there a reason not to use that class?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -691,27 +718,60 @@ fun Node.followNextDFGEdgesUntilHit(
                             // from).
                             // We try to pop from the stack and only select the elements with the
                             // matching index.
-                            indexStack?.checkAndPop(it.granularity as IndexedDataflowGranularity) ==
-                                true
+                            ctx.indexStack.checkAndPop(
+                                it.granularity as IndexedDataflowGranularity
+                            ) == true
```
KuechA, line 723: Is the comparison with `true` still needed?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -691,27 +718,60 @@ fun Node.followNextDFGEdgesUntilHit(
...
                         // CurrentNode is the child and nextDFG goes to ILE => currentNode's written
                         // to. Push to stack
-                        indexStack?.push(it.granularity as IndexedDataflowGranularity)
+                        ctx.indexStack.push(it.granularity as IndexedDataflowGranularity)
                     }
                 }
-                currentNode.nextDFG
+
+                // We need to filter out the edges which based on the stack
+                val selected =
```
KuechA, line 748: We should probably make a single run over the edges based on the `ctx` and its stacks and make this context configurable and include the push/pop logic. Otherwise, we'll have many parts of code which share very similar functionality and will run into issues with the maintainability. Might be a goal for the next few weeks

KuechA (approved):
I like that we will provide a context sensitive analysis. However, we should take care of few aspects:

1. maintainability of the selection of the "next step" for `followXUntilHit` is getting increasingly complex with a number of similar code fragments (push on stack, pop from stack, select subset of edges, ...)
2. `prevDFG` does not have this feature
3. The computational complexity will increase if we cannot disable certain stacks in the long term. We should keep this in mind when running it on larger code bases and allow configuring the precision.
4. Some functions already provide an interprocedural analysis and probably miss it (e.g. following the PDG, CDG), the EOG does not feature this at all.

I'm fine with the changes but we should discuss how to design a cleaner interface here in the next weeks.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
oxisto: Well the functionality is actually already implemented by the deque so why use the simple stack?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
oxisto: Probably not

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
oxisto: Agreed, should be part of https://github.com/Fraunhofer-AISEC/cpg/issues/1991

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
oxisto: I switched to the simple stack

PR #1999, opened 2025-01-28
Following EOG edges interprocedurally with stack

### LINKED ISSUE #1997: Follow EOG interprocedurally in query api
followNext/PrevEOGUntilHit works only intraprocedurally. This limits what we can identify and we should probably also provide an interprocedural traversal

KuechA:
Fixes #1997 

We use the `SimpleStack` to identify possible paths and follow the invokes/calledBy edges to extend the intraprocedural EOG to an interprocedural version thereof. A unification/better interface of the context and `SimpleStack` or an improved version of the Query API will follow in separate PRs.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -1067,13 +1143,15 @@ val Node?.assigns: List<AssignExpression>
...
  * @param predicate the search predicate
  */
-fun Node.firstParentOrNull(predicate: (Node) -> Boolean): Node? {
+inline fun <reified T : Node> Node.firstParentOrNull(
+    noinline predicate: ((T) -> Boolean)? = null
+): T? {
 
     // start at searchNodes parent
-    var node: Node? = this.astParent
+    var node = this.astParent
```
KuechA, line 1151: If `this` fulfills the predicate and type criterion, we won't return it. Fine for me but we have few lines where we would like to have the first matching node of this and the ast parents. Should we add a function for this too?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -1067,13 +1143,15 @@ val Node?.assigns: List<AssignExpression>
  *
  * @param predicate the search predicate
  */
-fun Node.firstParentOrNull(predicate: (Node) -> Boolean): Node? {
+inline fun <reified T : Node> Node.firstParentOrNull(
```
oxisto, line 1146: This change is so awesome! Can we extract this please into a separate PR because I think we are using this pattern in many more places

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -799,18 +805,60 @@ fun Node.followNextDFGEdgesUntilHit(
...
+        x = { currentNode, ctx, _ ->
+            if (
+                interproceduralAnalysis &&
+                    currentNode is CallExpression &&
+                    currentNode.invokes.isNotEmpty()
+            ) {
+                // We follow the invokes edges and push the call expression on the call stack, so we
+                // can jump back here after processing the function.
+                ctx.callStack.push(currentNode)
+                currentNode.invokes.flatMap { it.eogStarters }
```
oxisto, line 821: Why are we returning the EOG starters here?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -799,18 +805,60 @@ fun Node.followNextDFGEdgesUntilHit(
...
+            }
         },
         collectFailedPaths = collectFailedPaths,
         findAllPossiblePaths = findAllPossiblePaths,
         predicate = predicate,
     )
 }
 
+/**
+ * Returns the a [Collection] of last nodes in the EOG of this [FunctionDeclaration]. If there's no
```
oxisto, line 845: typo

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -799,18 +805,60 @@ fun Node.followNextDFGEdgesUntilHit(
...
+ * body, it will return a list of this function declaration.
+ */
+val FunctionDeclaration.lastEOGNode: Collection<Node>
```
oxisto, line 848: Shouldn't this be `lastEOGNodes` then?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: Of course. When I do so, I'd also address my comment above on whether we should immediately start in `node.astParent` or if we should return `node` immediately if it fulfills the `predicate` and is of `T`

KuechA: or should I introduce 2 functions?

KuechA: See #2018

oxisto (approved):

PR #2006, opened 2025-01-29
Merging `PythonAddDeclarationPass` into `SymbolResolver` using `HasDynamicDeclarations` trait

oxisto:
Work in progress. Also tries to merge the python declaration pass into the symbol resolver.

Certain PRs are needed before we can merge this:
- #2020
- #2010
- #2037

PR #2010, opened 2025-01-30
Execute `ComponentPass` passes according to the dependencies between components

oxisto:
This PR makes the same import-dependencies object that is currently available on a `Component` (between TUs) available on the `TranslationResult` (between components). I decided to generalize the `ImportDependencies` class for this. This might sounds overkill but my feeling is that this might pay off at some point.

To access that information, I added two new "strategies" (`TRANSLATION_UNITS_LEAST_IMPORTS` and `COMPONENTS_LEAST_IMPORTS`). I do not really like the UPPERCASE naming on these strategies, but we probably want to change that for all of them at some point. So they are named like this for consistency.

Furthermore, `ComponentPass` passes are now executed according to the `COMPONENTS_LEAST_IMPORTS` strategy and a `TranslationUnitPass` is executed according to first the `COMPONENTS_LEAST_IMPORTS` and then the `TRANSLATION_UNITS_LEAST_IMPORTS` inside the component. This should make it harmonised for all passes. 

The most important pass to use that is the `SymbolResolver`, which now resolves symbols of components in the desired order as preparation needed for #2006.

Because the `ImportResolver` populates this map, it cannot use this strategy and I had to promote this to a `TranslationResultPass`, which I think is ok. It still gathers the information on a component level, just the entry point is the translation result.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationResult.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationResult.kt
@@ -34,11 +34,14 @@ import de.fraunhofer.aisec.cpg.graph.edges.ast.astEdgesOf
 import de.fraunhofer.aisec.cpg.graph.edges.unwrapping
 import de.fraunhofer.aisec.cpg.helpers.MeasurementHolder
 import de.fraunhofer.aisec.cpg.helpers.StatisticsHolder
+import de.fraunhofer.aisec.cpg.passes.ImportDependencies
+import de.fraunhofer.aisec.cpg.passes.ImportResolver
 import de.fraunhofer.aisec.cpg.passes.Pass
 import de.fraunhofer.aisec.cpg.persistence.DoNotPersist
 import java.util.*
 import java.util.concurrent.ConcurrentHashMap
 import org.neo4j.ogm.annotation.Relationship
+import org.neo4j.ogm.annotation.Transient
 
 /**
  * The global (intermediate) result of the translation. A [LanguageFrontend] will initially populate
@@ -61,6 +64,11 @@ class TranslationResult(
      * of software.
      */
     val components by unwrapping(TranslationResult::componentEdges)
+
+    /** The import dependencies of [Component] nodes of this translation result. */
+    @Transient
+    @PopulatedByPass(ImportResolver::class)
+    var componentDependencies: ImportDependencies<Component>? = null
 
     /**
      * Scratch storage that can be used by passes to store additional information in this result.
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationResult.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationResult.kt
@@ -38,6 +38,7 @@ import de.fraunhofer.aisec.cpg.passes.ImportDependencies
 import de.fraunhofer.aisec.cpg.passes.ImportResolver
 import de.fraunhofer.aisec.cpg.passes.Pass
 import de.fraunhofer.aisec.cpg.persistence.DoNotPersist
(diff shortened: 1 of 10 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ImportResolver.kt
```diff
@@ -36,82 +39,85 @@ import de.fraunhofer.aisec.cpg.graph.scopes.NameScope
...
+        return added
     }
 
     /**
      * A work-list, which contains a local copy of our dependency map, so that we can remove items
      * from it while determining the order.
      */
-    class WorkList(start: ImportDependencies) :
-        HashMap<TranslationUnitDeclaration, MutableSet<TranslationUnitDeclaration>>() {
+    class WorkList<T : Node>(start: ImportDependencies<T>) : HashMap<T, MutableSet<T>>() {
```
KuechA, line 83: Would it be beneficial for the performance if we would use `IdentitySet` and `IdentityHashMap`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Component.kt
```diff
@@ -52,9 +53,13 @@ open class Component : Node() {
...
     val translationUnits by unwrapping(Component::translationUnitEdges)
 
+    /**
+     * The import dependencies of [TranslationUnitDeclaration] nodes of this component. The
+     * preferred way to access this is via [Strategy.TRANSLATION_UNITS_LEAST_IMPORTS].
+     */
     @Transient
     @PopulatedByPass(ImportResolver::class)
-    var importDependencies = ImportDependencies(translationUnits)
+    var translationUnitDependencies: ImportDependencies<TranslationUnitDeclaration>? = null
```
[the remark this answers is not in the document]
oxisto: Before, this was populated by an empty list of translation units in the constructor of the component and then we later replaced it with a complete new instance in the import resolver. This made it impossible to detect, whether we already resolved the imports or not. Now its clear that if this is "null", this information is not (yet) available and we can act accordingly (as the strategy does).

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationResult.kt
```diff
@@ -62,6 +66,14 @@ class TranslationResult(
      */
     val components by unwrapping(TranslationResult::componentEdges)
 
+    /**
+     * The import dependencies of [Component] nodes of this translation result. The preferred way to
+     * access this is via [Strategy.COMPONENTS_LEAST_IMPORTS].
+     */
+    @Transient
+    @PopulatedByPass(ImportResolver::class)
+    var componentDependencies: ImportDependencies<Component>? = null
```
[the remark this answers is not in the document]
oxisto: See above

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ImportResolver.kt
oxisto: That could work, I will have a look

oxisto: Seems that performance gain is negligible, 9.3s instead of 9.8 when parsing the stdlib typings. It still makes sense to be consistent to the other passes, but I am still not sure where the big performance loss comes from.

oxisto: Ok it came from the usage of the `namespaces` extension, which we used to collect sub-namespaces. This collected all children nodes in a a namespace and then filtered them. Instead we just looked at child declarations now. 1s instead of 10s in our stdlib typings test case now 🚀

KuechA (approved):

PR #2016, opened 2025-01-31
Handle `InitializerListExpression`s as variables of `ComprehensionExpression`s in the DFG

### LINKED ISSUE #1957: `InitializerListExpression` is not yet supported in the `ComprehensionExpression` in the CF-DFG pass
When analysing real-world python code, such as OpenStack, I get A LOT of these exceptions.

```
21:21:59,335 ERROR ControlFlowSensitiveDFGPass The variable of type class de.fraunhofer.aisec.cpg.graph.statements.expressions.InitializerListExpression is not yet supported in the ComprehensionExpression
```

KuechA:
Adapts the `ControlFlowSensitiveDFGPass` to handle multiple declarations and references in the variable of `ComprehensionExpression`s. To enhance the readability, the logic is moved to an own function.

Fixes #1957, #2017 

It further introduces a new interface `HasAccess` for Nodes which can be read from or written to. This allows to set the attribute `access: AccessValues` for more nodes than `Reference` and it also allows to implement a propagation logic per node class. Currently, the classes `Reference`, `InitializerListExpression` and `SubscriptExpression` implement this interface. It is mainly used to determine the direction of DFG edges.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Interfaces.kt
```diff
@@ -149,3 +149,11 @@ interface HasOverloadedOperation : HasOperatorCode {
      */
     val operatorBase: Expression
 }
+
+interface HasAccess {
```
oxisto, line 153: add a small doc string for the interface itself please.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
```diff
@@ -47,7 +46,7 @@ class ForEachStatement : LoopStatement(), BranchingNode, StatementHolder {
         astOptionalEdgeOf<Statement>(
             onChanged = { _, new ->
                 val end = new?.end
-                if (end is Reference) {
+                if (end is HasAccess) {
```
oxisto, line 49: Could be written as `(end as? HasAccess)?.access = ....` to be consistent with the other assignments.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/AssignExpression.kt
```diff
@@ -72,9 +72,9 @@ class AssignExpression :
                 }
 
                 if (isSimpleAssignment) {
```
oxisto, line 74: Could be written as 
```kotlin
(end as? HasAccess)?.access = if(isSimpleAssignment) {
  AccessValues.WRITE
} else {
  AccessValues.READWRITE
}
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/ComprehensionExpression.kt
```diff
@@ -43,7 +44,7 @@ class ComprehensionExpression : Expression(), ArgumentHolder {
             of = ProblemExpression("Missing variableEdge in ${this::class}"),
             onChanged = { _, new ->
                 val end = new?.end
-                if (end is Reference) {
+                if (end is HasAccess) {
```
oxisto, line 47: (it.end as? HasAccess)?.access to be consistent

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Interfaces.kt
```diff
@@ -149,3 +149,11 @@ interface HasOverloadedOperation : HasOperatorCode {
...
+    /**
+     * Is this reference used for writing data instead of just reading it? Determines dataflow
+     * direction
+     */
+    var access: AccessValues
```
oxisto, line 158: On second thought. Wouldn't it make sense to just implements this on all `Expression` nodes? The memory overhead is probably negligible and we would remove a lot of ugly `as? HasAccess`  casts. Basically this should apply to all expressions.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -538,6 +505,53 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : EOGStarterPass
...
+                    is SubscriptExpression -> (writtenToIt.arrayExpression as? Reference)?.refersTo
+                    else -> {
+                        log.error(
+                            "The variable of type ${writtenToIt.javaClass} is not yet supported in the ComprehensionExpression"
+                        )
+                        null
+                    }
+                }
+
+            // state.push(writtenToIt, PowersetLattice(identitySetOf(currentNode.iterable)))
```
oxisto, line 546: leftover?

oxisto (commented):
Some smaller remarks. In general this looks good, but I wonder whether we should move `access` to `Expression`.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Interfaces.kt
KuechA: I decided against this because many `Expression`s cannot receive any data (i.e., `AccessValues.WRITE` never makes sense for them and for some, also `READ` might be confusing), so the access value is not useful there. Are the casts are a major concern? Any opinions on this @konradweiss ?

oxisto: But all of them (most?) are part of the DFG and it is interesting to have a harmonized way to access this information even for a lot of them it will statically stay with READ.

oxisto: Probably altough a little bit awkward :(

KuechA: Well, I kept it with Expressions for now and removed the interface.

oxisto: Sounds like a reasonable compromise

oxisto (approved):
A little but "controversial" to fix the access here in this PR but its ok in my opinion. Let's move forward with this.

PR #2018, opened 2025-01-31
Refactor firstParentOrNull to accept type parameter and optional predicate

KuechA:
As discussed here: https://github.com/Fraunhofer-AISEC/cpg/pull/1999#discussion_r1936926974

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
@@ -1067,13 +1067,15 @@ val Node?.assigns: List<AssignExpression>
  *
  * @param predicate the search predicate
  */
-fun Node.firstParentOrNull(predicate: (Node) -> Boolean): Node? {
+inline fun <reified T : Node> Node.firstParentOrNull(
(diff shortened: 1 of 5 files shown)


KuechA:
Open discussion: Should we return `this` if it matches the predicate?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -1067,13 +1067,15 @@ val Node?.assigns: List<AssignExpression>
  *
  * @param predicate the search predicate
  */
-fun Node.firstParentOrNull(predicate: (Node) -> Boolean): Node? {
+inline fun <reified T : Node> Node.firstParentOrNull(
```
oxisto, line 1070: Two things
1) The description should be adjusted to reflect that this selects the first parent of type [T] that matches the (optional) predicate
2) It would be nice to also have this for the firstParentScope function.

oxisto: LGTM, minor documentation issue and possibly a small enhancement

oxisto (approved):

oxisto (approved):

PR #2019, opened 2025-01-31
Create a `LocalScope` for `CollectionComprehensions` and generate a `Declaration`

### LINKED ISSUE #2005: Python comprehensions should open a local scope and set/infer `VariableDeclaration`s for the variables
Original post by @maximiliankaul : https://github.com/Fraunhofer-AISEC/cpg/pull/1982#discussion_r1933827675

The generators, comprehensions and so on do not open a scope yet.

KuechA:
Fixes #2005 by

* Entering a `LocalScope` for list, set, dict comprehensions and generators. This somewhat simulates [PEP 709](https://peps.python.org/pep-0709/) where the loop variable is available only inside the loop.
* Generating a `VariableDeclaration` inside this scope in the `PythonAddDeclarationsPass` for the variables.
* Adds test that the local variables are not accessible outside the comprehension and that the `refersTo` edges are set
* For `AssignExpressions` with the `:=` operator inside a list/set/dict comprehension, the variable is looked up or created outside of the comprehension's `LocalScope` to address [PEP 572](https://peps.python.org/pep-0572/)


cpg-language-python/src/test/resources/python/comprehension.py
```diff
@@ -21,4 +22,10 @@ def dictComp(x, y):
...
 def generator(x, y):
     a = (i**2 for i in range(10) if i == 10)
-    b = (i**2 for i in range(10))
\ No newline at end of file
+    b = (i**2 for i in range(10))
+
+def bar(k, v):
+    return k+v
+
+def tupleComp(x):
```
oxisto, line 30: python code style -> `tuple_comp`

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
```diff
@@ -99,13 +99,25 @@ class ExpressionHandler(frontend: PythonLanguageFrontend) :
         }
     }
 
+    /**
+     * Works similar to [apply] but before executing [block], it enters the scope for this object
+     * and afterward leaves the scope again.
+     */
+    private inline fun <T : Node> T.applyWithScope(block: T.() -> Unit): T {
```
oxisto, line 106: This could be a helpful bit, also for the future node builder API, should we move this to the scope manager? Or keep it intentionally private for now?

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
```diff
@@ -129,10 +141,10 @@ class ExpressionHandler(frontend: PythonLanguageFrontend) :
      * a [CollectionComprehension].
      */
     private fun handleSetComprehension(node: Python.AST.SetComp): CollectionComprehension {
-        return newCollectionComprehension(rawNode = node).apply {
+        return newCollectionComprehension(rawNode = node).applyWithScope {
```
oxisto, line 144: I am not sure, if this is really relevant in our case, but it seems that the exact definition is that only the control variable is not leaked into the outer scope, but I guess the only sane way to model this is exactly like you did by opening up a local scope for the comprehension.

oxisto: Interesting, from https://peps.python.org/pep-0709/ -> "Comprehensions are currently compiled as nested functions, which provides isolation of the comprehension’s iteration variable, but is inefficient at runtime."

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
KuechA: I thought the same and wasn't sure where would be a good place. Maybe `Node` directly?

oxisto: When in doubt place it in `Extensions.kt`:D

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
oxisto: > Hm, this sounds as if the current behavior would be appropriately modeled with the local scope but the intended/future behavior might be different.

Yes, although I am not sure if this "new" bevahiour will leak the control variable or not. Maybe our PEP master @maximiliankaul understands this better than us.

> 
> Actually, we're only generating a `VariableDeclaration` inside this local scope for the variable of the comprehension expression and this is done inside the `PythonAddDeclarationsPass`. Leaving the scope earlier (i.e., right after generating the `ComprehensionExpression`) here won't impact on resolving variables, right?

I would assume so

KuechA: From my understanding, the variable should be available only in the `LocalScope` tied to the `ComprehensionExpression`. Obviously, we can always enter this scope to incorrectly resolve a variable but this shouldn't be the case. I also added a test which simulates the test case he provided in the original comment.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
```diff
@@ -117,10 +117,10 @@ class ExpressionHandler(frontend: PythonLanguageFrontend) :
      * into a [CollectionComprehension].
      */
     private fun handleListComprehension(node: Python.AST.ListComp): CollectionComprehension {
-        return newCollectionComprehension(rawNode = node).apply {
+        return newCollectionComprehension(rawNode = node).applyWithScope {
             statement = handle(node.elt)
             comprehensionExpressions += node.generators.map { handleComprehension(it, node) }
-            type = objectType("list") // TODO: Replace this once we have dedicated types
+            type = objectType("list")
```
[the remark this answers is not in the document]
KuechA: Well, the TODO as it is written here does not make sense because we do have a dedicated type which is looked up in the `builtInTypes` map by `objectType`. That's why it is removed. I think that the original code was written when this was not been the case (yet).

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
KuechA: I added a switch for different versions. For all versions > python 3.0.0, we create the `LocalScope`. If nothing is specified or for smaller versions, we do not create the `LocalScope`. I also added a test for it

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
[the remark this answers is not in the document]
KuechA: Changed to `primitiveType` which throws an exception if the type is not part of the `builtInTypes`. This will help us identify changes with unwanted effects

oxisto (approved):

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguageFrontend.kt
```diff
@@ -356,13 +356,17 @@ class PythonLanguageFrontend(language: Language<PythonLanguageFrontend>, ctx: Tr
  * Returns the version info from the [TranslationConfiguration] as [VersionInfo] or `null` if it was
  * not specified.
  */
-fun getVersionInfo(config: TranslationConfiguration): VersionInfo? {
-    return config.symbols["PYTHON_VERSION_MAJOR"]?.toLong()?.let { major ->
-        val minor = config.symbols["PYTHON_VERSION_MINOR"]?.toLong()
-        val micro = if (minor != null) config.symbols["PYTHON_VERSION_MICRO"]?.toLong() else null
-        VersionInfo(major, minor, micro)
+val TranslationConfiguration.versionInfo: VersionInfo?
```
KuechA, line 359: I'll this and the other changes related to `VersionInfo` to a separate PR since the change is no longer needed for the functionality of this PR and thus introduces unnecessary changes

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/CollectionComprehensionTest.kt
```diff
@@ -0,0 +1,1790 @@
...
+            listComprehensionWithTupleAndAssignmentToListElement,
+            innerVariableDeclarationA.scope?.astNode,
+            "We expect that the reference \"a\" refers to a VariableDeclaration with localName \"a\" which is not null and whose scope is the LocalScope of the list comprehension.",
+        )
+        assertRefersTo(
+            refB,
+            bDeclaration,
+            "We expect that the reference \"b\" in the tuple refers to the VariableDeclaration of \"b\" which is added outside the list comprehension (in statement 0).",
+        )
+        assertRefersTo(
```
KuechA, line 1533: We should discuss what we expect in this case. My version of Python3.13 crashes with the error "UnboundLocalError: cannot access local variable 'a' where it is not associated with a value".

We may interpret this by setting the `refersTo` to same as the reference `a` in the second item and later, an analysis must figure out that this variable is not initialized yet (we should check the EOG for this too).

We may also interpret this that `refersTo` should be `null`

oxisto: I would say that `refersTo` should be null. I think we did that in other cases where we have an unbound error as well

KuechA: Ok, since the `SymbolResolver` is in charge of this, I added issue #2035 which blocks this PR

oxisto: I would propose to remove the assert there, unblock and merge this PR, otherwise we will get stuck on many follow-ups, this is only a minor detail and actually has nothing to do with the PR itself.

Issue #1995, opened 2025-01-28
Provide a better version of the `dataflow` query extension


oxisto:
There exist some problems (@KuechA will fill these)
- Distinguish between "all" paths, any paths, etc.
- It's hard to differentiate between good and bad cases in the query tree

Issue #1991, opened 2025-01-27
Discuss interface providing configurable selection of sensitivity when following certain subgraphs


KuechA:
> [...][W]e should take care of few aspects:
> 
> 1. maintainability of the selection of the "next step" for `followXUntilHit` is getting increasingly complex with a number of similar code fragments (push on stack, pop from stack, select subset of edges, ...)
> 2. `prevDFG` does not have this feature
> 3. The computational complexity will increase if we cannot disable certain stacks in the long term. We should keep this in mind when running it on larger code bases and allow configuring the precision.
> 4. Some functions already provide an interprocedural analysis and probably miss it (e.g. following the PDG, CDG), the EOG does not feature this at all.
>
> I'm fine with the changes but we should discuss how to design a cleaner interface here in the next weeks.

_Originally posted by @KuechA in https://github.com/Fraunhofer-AISEC/cpg/pull/1988#pullrequestreview-2574667440_

PR #2025, opened 2025-02-04
New sketch of API for dataflow queries

### LINKED ISSUE #2024: Distinguish between "all" paths, any paths, etc. of dataflow queries

KuechA:
This PR aims to enable a more precise configuration of the queries following dataflows and others.

Fixes #2024 #2042 #2045 #1991 #1995


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -0,0 +1,528 @@
...
+ * Determines how far we want to follow edges within the analysis. [maxSteps] defines the total
+ * number of steps we will follow in the graph (unlimited depth if `null`).
+ */
+sealed class AnalysisScope(val maxSteps: Int? = null)
+
+/**
+ * Only intraprocedural analysis. [maxSteps] defines the total number of steps we will follow in the
+ * graph (unlimited depth if `null`).
+ */
+class INTRAPROCEDURAL(maxSteps: Int? = null) : AnalysisScope(maxSteps)
```
oxisto, line 44: Any particular reason for UPPERCASING a class name?

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -0,0 +1,528 @@
...
+/**
+ * Enable interprocedural analysis. [maxCallDepth] defines how many function calls we follow at most
+ * (unlimited depth if `null`). [maxSteps] defines the total number of steps we will follow in the
+ * graph (unlimited depth if `null`).
+ */
+class INTERPROCEDURAL(val maxCallDepth: Int? = null, maxSteps: Int? = null) :
+    AnalysisScope(maxSteps)
+
+/** Determines in which direction we follow the edges. */
+enum class AnalysisDirection {
```
oxisto, line 55: I wonder whether it would make sense to make this a sealed class as well and then make the enum entries objects instead. And then maybe (if that makes sense) we could move some of the analysis logic actually into these Classes instead of having a lot of "if this direction, then" logic in the followX functions.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -0,0 +1,528 @@
...
+    ): Array<AnalysisSensitivity> {
+        return arrayOf(*this.toTypedArray(), other)
+    }
+
+    operator fun plus(other: Array<AnalysisSensitivity>): Array<AnalysisSensitivity> {
+        return arrayOf(*other, this)
+    }
+}
+
+fun dataFlowBase(
```
oxisto, line 109: The naming "base" is misleading here. Isn't this the "extended" version of the function and the other one is the simpler one. Although I don't know if that is really necessary when we can just have default parameters 🤷

oxisto (commented):
Interesting approach. I would propose the usage of sealed objects in order to move some of the logic from the follow-functions to the analysis parameters

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
KuechA: No, it's not necessary and the old functions are now deleted.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
KuechA: I think we cannot easily remove code from the followX functions (yet) because each subgraph works quite different. But I think that at least some steps can be moved to these classes. E.g. the max depth checks should work already. For others, I'm less confident that it would work right now mainly because the DFG and EOG work quite differently.

I made this also a sealed class now to prepare such a change in this interface.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -0,0 +1,654 @@
...
+    earlyTermination: ((Node) -> Boolean)? = null,
+    predicate: (Node) -> Boolean,
+): QueryTree<Boolean> {
+    val collectFailedPaths = type == AnalysisType.MUST || verbose
+    val findAllPossiblePaths = type == AnalysisType.MUST || verbose
+    val earlyTermination = { n: Node, ctx: Context -> earlyTermination?.let { it(n) } == true }
+
+    val evalRes =
+        if (direction is Bidirectional) {
+                arrayOf(Forward(), Backward())
```
KuechA, line 342: There must be a better way to achieve this (and ideally get rid of the `fold`). Any smart ideas?

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -0,0 +1,654 @@
...
+
+/** Consider the calling context when following paths (e.g. based on a call stack). */
+class ContextSensitive() : AnalysisSensitivity() {
+    override fun followEdge(
+        currentNode: Node,
+        edge: Edge<Node>,
+        ctx: Context,
+        analysisDirection: AnalysisDirection?,
+    ): Boolean {
+        return if (edge is ContextSensitiveDataflow && analysisDirection is Forward) {
```
KuechA, line 180: Currently, this only works for `Dataflow` edges but we also need something for the `EvaluationOrder`. However, it will probably be somewhat different...

I also dislike that the code for `Forward` and `Backward` looks similar with the only difference being when to use `CallingContextIn` and `CallingContextOut`. Good ideas on how to improve this are welcome

oxisto (approved):
There is some potential to merge things with `Strategy` (or the other way around). But we will do that in a future PR.

PR #2034, opened 2025-02-07
Resolve import types for superclasses

lshala:
This PR adds the handling for nested `AST.Attribute`s, ensuring that superclass imports are correctly resolved


lshala:
Unfortunately, this change is causing errors in the `SymbolResolver`. I tried debugging and noticed that the `getInvocationCandidatesFromParents` method is running in an endless loop. Need your help @oxisto

oxisto:
> Unfortunately, this change is causing errors in the `SymbolResolver`. I tried debugging and noticed that the `getInvocationCandidatesFromParents` method is running in an endless loop. Need your help @oxisto

Where exactly? Do you have a test file?

lshala:
> > Unfortunately, this change is causing errors in the `SymbolResolver`. I tried debugging and noticed that the `getInvocationCandidatesFromParents` method is running in an endless loop. Need your help @oxisto
> 
> Where exactly? Do you have a test file?

I tried to parse the `cinder/api` folder and the error occurs here: [urlmap.py](https://github.com/openstack/cinder/blob/a1fb7da213b72040b8c3fa2b2708f23c21da20b2/cinder/api/urlmap.py#L124)

oxisto:
> > > Unfortunately, this change is causing errors in the `SymbolResolver`. I tried debugging and noticed that the `getInvocationCandidatesFromParents` method is running in an endless loop. Need your help @oxisto
> > 
> > 
> > Where exactly? Do you have a test file?
> 
> I tried to parse the `cinder/api` folder and the error occurs here: [urlmap.py](https://github.com/openstack/cinder/blob/a1fb7da213b72040b8c3fa2b2708f23c21da20b2/cinder/api/urlmap.py#L124)

There seems to be a confusion in https://github.com/openstack/cinder/blob/a1fb7da213b72040b8c3fa2b2708f23c21da20b2/cinder/api/urlmap.py#L168 the super declaration is the same as the declaration after resolving. Which is weird because the record is `cinder.api.urlmap.URLMap` and the super type is supposed to be `paste.urlmap.URLMap`, so either the parsed name is wrong, or somehow it gets resolved incorrectly.

PR #2046, opened 2025-02-14
Change kwargs to isVariadic and add HasDefaultArguments to PythonLanguage

lshala:
- For Methods with default parameters or `**kwargs` , the ScopeManager failed to match the signatures. This leads to missing or incorrectly set `invokes` edges. This PR fixes the issue by setting `isVariadic = true` for `**kwargs` and adding `HasDefaultMembers` for `PythonLanguage` to properly handle default parameters.
@maximiliankaul Thx for your support.

- The `attachCallParameters` function, responsible for setting DFG edges between functions and their corresponding calls, is being refactored to extend its handling beyond positional arguments. It now also supports named arguments, variadic arguments (both `*args` and `**kwargs`) and default arguments.


cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/ArgumentsHandlerTest.kt
```diff
@@ -195,4 +195,10 @@ class ArgumentsHandlerTest {
         assertNotNull(parameterC?.default, "Expected the parameter `c` to have a default value.")
         assertEquals(2.toLong(), parameterC.default?.evaluate())
     }
+
+    @Test
+    fun testSignatureMatch() {
+        val func = result.functions["call"]
+        assertNotNull(func)
```
KuechA, line 202: This tests nothing

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
```diff
@@ -45,9 +45,11 @@ class PythonLanguage :
     HasShortCircuitOperators,
     HasOperatorOverloading,
     HasFunctionStyleConstruction,
-    HasMemberExpressionAmbiguity {
+    HasMemberExpressionAmbiguity,
+    HasDefaultArguments {
```
KuechA, line 49:
```suggestion
    HasDefaultArguments,
{
```

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/scopes/ScopeManagerTest.kt
```diff
@@ -137,4 +137,60 @@ internal class ScopeManagerTest : BaseTest() {
             assertNotNull(scope)
         }
     }
+
+    @Test
```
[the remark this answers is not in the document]
KuechA: The test is ok here

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
```diff
@@ -383,33 +385,47 @@ object Util {
...
-                        j++
+        var parameterIndex = 0
+
+        for (param in functionParameters) {
+            // Handle default parameters
+            if (param.default != null) {
+                val argumentEdge = call.argumentEdges.getOrNull(parameterIndex)
+                if (argumentEdge != null) {
+                    val isNamedArgument = argumentEdge.name != null
+                    if (isNamedArgument) {
```
KuechA, line 399: Aren't there also named arguments which are not variadic? When will those be handled?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
```diff
@@ -383,33 +385,47 @@ object Util {
...
-                    while (j < arguments.size) {
-                        // map all the following arguments to this variadic param
-                        param.prevDFGEdges.addContextSensitive(
-                            arguments[j],
-                            callingContext = CallingContextIn(call),
-                        )
-                        j++
+        var parameterIndex = 0
+
+        for (param in functionParameters) {
```
KuechA, line 393: I'm wondering which information are set in the argument `Edge`s on which we can rely here. The names and the index for sure but is there more?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
lshala: Done

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
KuechA: I think it's ok like this. Maybe some of the names and comments (e.g. kwargs, args) are somewhat specific to python and some parts could benefit from an additional comment but the functionality is fine.

PR #2053, opened 2025-02-18
First Proposal for marking EOG and DFG as overlaying nodes, a computed property is used to retrieve that information.

konradweiss:


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/Edge.kt
```diff
@@ -73,6 +74,13 @@ abstract class Edge<NodeType : Node> : Persistable, Cloneable {
 
     abstract var labels: Set<String>
 
+    /**
+     * Whether the Edge is pointing to or from the overlay network is defined by its connected
```
KuechA, line 78: I'd change this description to something like "`true` if one of the connected nodes is an overlay node".

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/Edge.kt
oxisto: also `Edge` should be `[Edge]`

konradweiss: Done

PR #2057, opened 2025-02-19
Providing a `Language::translationUnitForInference` function that is used for global inference

oxisto:
Currently, we use the global scope's `astNode` as a start node for inference of "global" things. However, this is very buggy, since the `astNode` of the global scope is assigned to the last TU we parse. If this TU was parsed in another language, we can run into situations where we infer things in the wrong language. 

This PR fixes that by introducing a new `Language::translationUnitForInference` function that defaults to the first TU in a component and can be used by a language to provide a specific TU for specific declarations.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
@@ -868,6 +868,27 @@ class ScopeManager : ScopeProvider {
 
         return symbols.singleOrNull()
     }
+
+    /**
+     * Returns the [TranslationUnitDeclaration] that should be used for inference, especially for
+     * global declarations.
+     *
+     * @param TypeToInfer the type of the node that should be inferred
+     * @param source the source that was responsible for the inference
+     */
+    fun <TypeToInfer : Node> translationUnitForInference(
+        source: Node
+    ): TranslationUnitDeclaration? {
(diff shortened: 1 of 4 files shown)


KuechA (commented):
From your description of this PR, I understand that the main issue is the `language` used for the inference but not the astNode itself. Wouldn't it be more appropriate to pass the language of the `source.translationUnit` to the inference system instead of changing the "base node"? Does the "base node" where the inference starts somehow affect the result and we would now infer a node multiple times for each TU instead of once globally?

oxisto:
> From your description of this PR, I understand that the main issue is the `language` used for the inference but not the astNode itself. Wouldn't it be more appropriate to pass the language of the `source.translationUnit` to the inference system instead of changing the "base node"? Does the "base node" where the inference starts somehow affect the result and we would now infer a node multiple times for each TU instead of once globally?

Good question. I am unsure myself why so many tests are failing now

oxisto:
> From your description of this PR, I understand that the main issue is the `language` used for the inference but not the astNode itself. Wouldn't it be more appropriate to pass the language of the `source.translationUnit` to the inference system instead of changing the "base node"? Does the "base node" where the inference starts somehow affect the result and we would now infer a node multiple times for each TU instead of once globally?

I fixed the tests by adding the existing old ast node as a backup in cases where we do not have access to the TU, especially if the "start" is a type node. This is not really a good solution but the alternative would be to rework the inference system alltogether.

KuechA (approved):

PR #2061, opened 2025-02-20
Modelling program entrypoints as concepts

oxisto:
This includes main and library entry points and also makes an `HttpEndpoint` a remote entry point. Also adds support for an operating system concept and linking library entry points with dynamic loading operations.

### DIFF
--- a/cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/arch/OperatingSystem.kt
+++ b/cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/arch/OperatingSystem.kt
@@ -23,34 +23,24 @@
  *                    \______/ \__|       \______/
  *
  */
-package de.fraunhofer.aisec.cpg.graph.concepts.http
+package de.fraunhofer.aisec.cpg.graph.concepts.arch
 
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.concepts.Concept
-import de.fraunhofer.aisec.cpg.graph.concepts.Operation
-import de.fraunhofer.aisec.cpg.graph.concepts.flows.RemoteEntryPoint
 
-/** Represents a single [HttpEndpoint] on the server */
-class HttpEndpoint(
-    underlyingNode: Node,
-    val httpMethod: HttpMethod,
-    val path: String,
-    val arguments: List<Node>,
-    val supportedAuthentications: MutableList<String>,
-) : RemoteEntryPoint(underlyingNode = underlyingNode)
+/** Represents an architecture of an operating system. */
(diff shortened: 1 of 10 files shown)


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/flows/EntryPoint.kt
```diff
@@ -0,0 +1,58 @@
...
+    underlyingNode: Node,
+    /**
+     * If this entry point is specifically designed to be invoked on a certain
+     * [OperatingSystemArchitecture], it can be specified here.
+     */
+    var os: OperatingSystemArchitecture? = null,
+) : EntryPoint(underlyingNode = underlyingNode)
+
+/** The main function of a program. */
+class Main(underlyingNode: Node, os: OperatingSystemArchitecture? = null) :
```
fwendland, line 50: We may need some logic later to enforce that `Main` is a `main` function, i.e., the underlying node points to a `main` function declaration.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/arch/OperatingSystem.kt
```diff
@@ -0,0 +1,55 @@
...
+ *
+ */
+package de.fraunhofer.aisec.cpg.graph.concepts.arch
+
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.concepts.Concept
+import kotlin.reflect.full.isSubclassOf
+
+/** Represents an architecture of an operating system. */
+abstract class OperatingSystemArchitecture(underlyingNode: Node) :
```
fwendland, line 33: We could add an `Any`/`Agnostic`/`Independent` OS type for languages like Java, (some) Python, etc. to not have to do the `null` check everywhere.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/ConceptPass.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * This function is called for each node in the graph. It needs to be overridden by subclasses
+     * to handle the specific node.
+     */
+    abstract fun handleNode(node: Node?, tu: TranslationUnitDeclaration)
+
+    /**
+     * Gets the [DynamicLoading] concept for this [TranslationUnitDeclaration] or creates a new one
```
fwendland, line 62: Update comment. We're now getting or creating a new concept of the requested type instead of a `DynamicLoading` concept.

fwendland (approved):

PR #2064, opened 2025-02-21
Drawing additional DFG edges

oxisto:


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/config/Configuration.kt
```diff
@@ -0,0 +1,110 @@
...
+    init {
+        conf.groups.add(this)
+    }
+}
+
+/**
+ * Represents a configuration option within one [group]. Usually there is one option for each entry
+ * in a configuration file.
+ */
+class ConfigurationOption(underlyingNode: Node, var group: ConfigurationGroup) :
```
KuechA, line 64: So, we do not explicitly define that there's a key of the option and a value?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/ConceptPass.kt
```diff
@@ -46,9 +48,16 @@ abstract class ConceptPass(ctx: TranslationContext) : TranslationUnitPass(ctx) {
     override fun accept(tu: TranslationUnitDeclaration) {
         ctx.currentComponent = tu.component
         walker = SubgraphWalker.ScopedWalker(ctx.scopeManager)
+        walker.strategy = Strategy::EOG_FORWARD
         walker.registerHandler { node -> handleNode(node, tu) }
 
-        walker.iterate(tu)
+        // Gather all resolution EOG starters; and make sure they really do not have a
+        // predecessor, otherwise we might analyze a node multiple times
+        val nodes = tu.allEOGStarters.filter { it.prevEOGEdges.isEmpty() }
```
KuechA, line 56: Don't we have similar code already somewhere?

oxisto: In theory there is the EOGStarterPass, but if I change it to one, I somehow get incorrect results when retrieving  the current TU :(

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/config/Configuration.kt
oxisto: hm I can add that. currently the underlying Node is the "key" -> in case of the ini frontends its a field declaration and the value would be its initialiser.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/ConceptPass.kt
```diff
@@ -46,9 +48,16 @@ abstract class ConceptPass(ctx: TranslationContext) : TranslationUnitPass(ctx) {
     override fun accept(tu: TranslationUnitDeclaration) {
         ctx.currentComponent = tu.component
         walker = SubgraphWalker.ScopedWalker(ctx.scopeManager)
+        walker.strategy = Strategy::EOG_FORWARD
```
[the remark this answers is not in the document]
oxisto: There is no harm in walking the EOG and in some cases it’s needed.

PR #2066, opened 2025-02-24
Update ProgramDependenceGraph (PDG) and enable tracking implicit dataflows

KuechA:
This PR aims to achieve the following things:

* [x] Introduce `ProgramDependence` edges as a common super class for the CDG and DFG edges.
* [x] Get rid of cloning edges to add them to the PDG now that OGM is gone.
* [x] Implement the new dataflow-following functionality `Implicit` which follows the PDG instead of the DFG.
* [ ] Improve the documentation of the effects of the `Sensitivities`.
* [ ] Document the difference between implicit and explicit dataflows with few examples.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/ProgramDependence.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/ProgramDependence.kt
@@ -31,6 +31,7 @@ import de.fraunhofer.aisec.cpg.graph.edges.collections.EdgeSet
 import de.fraunhofer.aisec.cpg.graph.edges.collections.MirroredEdgeCollection
 import de.fraunhofer.aisec.cpg.passes.ProgramDependenceGraphPass
 import kotlin.reflect.KProperty
+import org.neo4j.ogm.annotation.RelationshipEntity
 
 /** The types of dependences that might be represented in the CPG */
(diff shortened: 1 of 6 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/ProgramDependence.kt
```diff
@@ -70,8 +69,32 @@ class ProgramDependences<NodeType : Node> :
...
+
+/**
+ * This edge class defines that there's some kind of dependency between [start] and [end]. The
+ * nature of this dependency is defined by [dependence].
+ */
+@RelationshipEntity
+open class ProgramDependence(start: Node, end: Node, dependence: DependenceType) :
+    Edge<Node>(start, end) {
+    init {
+        this.dependence = dependence
```
oxisto, line 84: would it make sense to move the dependence field to this class instead of having it in every edge?

oxisto (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/ProgramDependence.kt
KuechA: I guess so because I don't know what should be the use-case of having it in the `Edge` class in general. I'll update this.

PR #2069, opened 2025-02-24
Merge DFG granularities

### LINKED ISSUE #1970: Discuss and align different partial DFG edges
There are various ways how only a part of an objects data flow into another object. Currently, we support edge labels based on indices (only implemented for tuples) and labels based on (Field)Declarations. We should collect a more complete list of requirements. E.g. array indices or other offsets.

Original post: https://github.com/Fraunhofer-AISEC/cpg/pull/1961#discussion_r1926551254

KuechA:
There are different DFG-granularities which did not share anything yet. Now there's a common base-class which we can use. This PR also merges special handling for `TupleDeclaration`s to be the same as tuples with references (crazy!) and all other partial-DFG handling.

Most changes come from re-naming the function `partial` to `field` and the old `PartialDataflowGranularity` to `FieldDataflowGranularity`. Names can be discussed, similar if this approach is what was expected.

Fixes #1970


oxisto:
> > how would be model an access to an array by a (constant) expression, something like `a["b"]`?
> 
> I think we can simply use `PartialDataflowGranularity<String>` to model this. A bit more tricky is something like `a[b]`, in particular when `b` is not constant. For this case, we may use the Reference or we may need something smarter but I think this wouldn't generalizable that easily. I made the `PartialDataflowGranularity` as open class instead of sealed class to allow this.

Hmm for me this would be closer to the `IndexedDataflowGranularity` and would basically be something like `ArrayItemDataflowGranularity`  (weird naming) and then have an "evaluatedIndex" which can be Any? Also not really pretty :(

oxisto (approved):

PR #2073, opened 2025-02-25
improve assertion

maximiliankaul:
This pr makes `assertLiteralValue` guarantee that the checked `expr` is not null if the assertion succeeds. This enables us to not have a null check after `assertLiteralValue`.


cpg-core/src/testFixtures/kotlin/de/fraunhofer/aisec/cpg/test/TestUtils.kt
```diff
@@ -314,10 +316,16 @@ fun assertLocalName(localName: String, node: Node?, message: String? = null) {
 }
 
 /**
- * Asserts that a) the expression in [expr] is a [Literal] and b) that it's value is equal to
- * [expected].
+ * Asserts that
+ * - the expression in [expr] is a [Literal] and
+ * - that it's value is equal to [expected].
+ *
+ * Guarantees that [expr] is not null if the assertion on the value succeeds. Note: this is an ugly
```
oxisto, line 323: Why is that a hack? That is just he way `assertIs` works?

oxisto (approved):

PR #2081, opened 2025-02-26
Fix: importedFrom now checks the import instead of name

maximiliankaul:
All references are normalized against alias imports. Thus, we have to check against the imported symbol and not the imported name.

TODO:
-[ ] Fix Python test (issue with normalizing an aliased call)


oxisto:
Closed in favour of #2085

oxisto (approved):

PR #2082, opened 2025-02-26
Python logging concept

maximiliankaul:
Initial draft for Pythons `import logging`

### DIFF
--- a/cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/logging/LoggingConceptTest.kt
+++ b/cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/logging/LoggingConceptTest.kt
@@ -165,10 +165,7 @@ class LoggingConceptTest : BaseTest() {
 
         assertTrue(
             dataFlow(startNode = literalERROR) {
-                    it is Log &&
-                        it.name.contains(
-                            "__name__"
-                        ) // TODO: checking for the correct logger like this is not pretty
+                    it is Log && it.underlyingNode?.code == "log.getLogger(__name__)"
                 }
                 .value,
             "Expected to find a dataflow from the literal \"ERROR\" to the logging node based on the `getLogger(__name__)` call.",
@@ -200,16 +197,15 @@ class LoggingConceptTest : BaseTest() {
             "Expected to find 3 logging nodes. One from the `import logging as log` declaration and one `foo` and one `bar` logger from the `log.getLogger()` calls. The other `getLogger()` calls are duplicates and must not create new loggers.",
         )
 
-        val defaultLogger =
-            allLoggers.singleOrNull { it.name.toString() == "Log[]" } // TODO: nicer logger check
+        val defaultLogger = allLoggers.singleOrNull { it.underlyingNode is ImportDeclaration }
         assertNotNull(defaultLogger)
 
         val fooLogger =
-            allLoggers.singleOrNull { it.name.toString() == "Log[foo]" } // TODO: nicer logger check
+            allLoggers.singleOrNull { it.underlyingNode?.code == "logging.getLogger('foo')" }
         assertNotNull(fooLogger)
 
         val barLogger =
-            allLoggers.singleOrNull { it.name.toString() == "Log[bar]" } // TODO: nicer logger check
+            allLoggers.singleOrNull { it.underlyingNode?.code == "logging.getLogger('bar')" }
         assertNotNull(barLogger)
 
         // Testing setup. A map of a logger name (and thus the literal written to the logger) and
--- a/cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/logging/Log.kt
+++ b/cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/logging/Log.kt
@@ -33,6 +33,4 @@ import de.fraunhofer.aisec.cpg.graph.concepts.Concept
  *
  * @param underlyingNode The underlying CPG node.
  */
-class Log(underlyingNode: Node) : Concept(underlyingNode = underlyingNode), IsLogging {
-    // TODO: target (stdout / err / file / ...), logger level, ...
-}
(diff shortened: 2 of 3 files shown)


cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/logging/LoggingConceptTest.kt
```diff
@@ -0,0 +1,175 @@
...
+import de.fraunhofer.aisec.cpg.test.assertLiteralValue
+import java.nio.file.Path
+import kotlin.test.*
+import org.junit.jupiter.api.Tag
+
+/**
+ * A class for integration tests. They depend on the Python frontend, so we classify them as an
+ * integration test. This might be replaced with a language-neutral test at some point.
+ */
+@Tag("integration")
```
oxisto, line 48: Not needed anymore

cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/logging/LoggingConceptTest.kt
```diff
@@ -0,0 +1,175 @@
...
+class LoggingConceptTest : BaseTest() {
+    @Test
+    fun test01() {
```
oxisto, line 51: pls use a proper function name

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/logging/LogWriteOperation.kt
```diff
@@ -0,0 +1,57 @@
...
+    INFO,
+    DEBUG,
+    TRACE,
+    UNKNOWN,
+}
+
+/**
+ * A log write operation e.g. `log.warn("...")`.
+ *
+ * @param underlyingNode The underlying CPG node.
```
oxisto, line 45: Is it possible to add "@param" here at the class level? Are you sure about this?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/PythonLoggingConceptPass.kt
```diff
@@ -0,0 +1,212 @@
...
+ *                   $$  __$$\ $$  __$$\ $$  __$$\
+ *                   $$ /  \__|$$ |  $$ |$$ /  \__|
+ *                   $$ |      $$$$$$$  |$$ |$$$$\
+ *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.passes.concepts
```
oxisto, line 26: Should be in directory logging.python

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/PythonLoggingConceptPass.kt
```diff
@@ -0,0 +1,212 @@
...
+import de.fraunhofer.aisec.cpg.passes.ComponentPass
+import de.fraunhofer.aisec.cpg.passes.configuration.ExecuteLate
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+/**
+ * This pass collects Python's `import logging` logs and maps them to the corresponding CPG
+ * [Concept] nodes.
+ */
+@ExecuteLate
```
oxisto, line 46: Rather than having execute later, wouldn't it be better to explicitly state the dependencies it has, e.g. -> SymbolResolver/EOG?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/logging/LoggingNode.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.concepts.Concept
+
+/**
+ * A logging node, i.e. a node corresponding to a log where [LogWriteOperation] can write to.
+ *
+ * @param underlyingNode The underlying CPG node.
+ */
+class LoggingNode(underlyingNode: Node) : Concept(underlyingNode = underlyingNode), IsLogging {
```
oxisto, line 36: Wouldn't this just be a "Log"?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/PythonLoggingConceptPass.kt
oxisto: > I don't think so. I believe the concept passes build upon "the cpg", thus `ExecuteLate` is exactly correct. We can add more dependencies like "depends on other concept pass" or depends on "symbolresolver/eog" to make this more explicit, but I'd like to keep `ExecuteLate`

This is not necessarily true. We do have some concept passes that are executed in-between "normal" CPG passes, for example concepts about dynamic loading. So not every concept pass should have execute late.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -492,6 +492,15 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
 
         // We also set the callee's refersTo
         callee.refersTo = call.invokes.firstOrNull()
+        val refersToName = callee.refersTo?.name
```
oxisto, line 495: You need to revert the symbol resolver back to main. It looks like you merged in an old version of the #2085 branch

PR #2083, opened 2025-02-26
Fixed bug in `translationUnitForInference`

oxisto:
There was a bug in `translationUnitForInference` where we returned an incorrect translation result.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
@@ -37,6 +37,7 @@ import de.fraunhofer.aisec.cpg.ancestors
 import de.fraunhofer.aisec.cpg.graph.Name
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.OverlayNode
+import de.fraunhofer.aisec.cpg.graph.component
 import de.fraunhofer.aisec.cpg.graph.declarations.Declaration
 import de.fraunhofer.aisec.cpg.graph.declarations.FunctionDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.NamespaceDeclaration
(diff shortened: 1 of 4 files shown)


cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt
```diff
@@ -1923,4 +1923,22 @@ class PythonFrontendTest : BaseTest() {
         assertNotNull(myClass)
         assertNotNull(myClass.ancestors)
     }
+
+    @Test
+    fun testNestedReplace() {
```
KuechA, line 1928: I have no idea what is the purpose of this test. We expect to infer functions in the translation unit where it is missing?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -416,10 +417,17 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
      * @param source the source that was responsible for the inference
      */
     fun <TypeToInfer : Node> translationUnitForInference(source: Node): TranslationUnitDeclaration {
-        val tu = source.ctx?.currentComponent?.translationUnits?.firstOrNull()
+        // The easiest way for the current component would be the traversing the AST, but that does
```
KuechA, line 420:
```suggestion
        // The easiest way to identify the current component would be traversing the AST, but that does
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -405,8 +406,8 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
      *
      * Therefore, we give the language a chance to return a [TranslationUnitDeclaration] where the
      * declaration should be placed. If the language does not override this function, the default
-     * implementation will return the first [TranslationUnitDeclaration] in
-     * [TranslationContext.currentComponent].
+     * implementation will return the first [TranslationUnitDeclaration] in the component of
```
KuechA, line 409:
```suggestion
     * implementation will return the first [TranslationUnitDeclaration] in the [Component] of
```

KuechA (approved):

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt
oxisto: This was needed because of an intricate bug that occured when we need to infer something based on a node that is an argument of a call expression that was replaced :D

PR #2084, opened 2025-02-26
Indexed (and inverted) subscript expression in dfg

KuechA:
Aims to improve DFG edges for subscript expressions


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/HasType.kt
```diff
@@ -134,7 +134,7 @@ interface HasType : ContextProvider, LanguageProvider {
          * A helper function that can be used for [EdgeSingletonList.onChanged]. It unregisters this
          * [TypeObserver] with the [old] node and registers it with the [new] one.
          */
-        fun <NodeType : Node> exchangeTypeObserver(
+        fun <NodeType : Node> exchangeTypeObserverWithAccessPropagation(
```
oxisto, line 137: Can't we just add another mandatory (!) parameter `propagateAccess`. Seems error-prone to update two different functions

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
```diff
@@ -466,7 +468,21 @@ class DFGPass(ctx: TranslationContext) : ComponentPass(ctx) {
...
     protected fun handleSubscriptExpression(node: SubscriptExpression) {
-        node.prevDFGEdges += node.arrayExpression
+        if (node.access == AccessValues.WRITE) {
+                node.nextDFGEdges
+            } else {
+                node.prevDFGEdges
+            }
+            .add(node.arrayExpression) {
+                granularity =
+                    if ((node.subscriptExpression as? Literal<*>)?.value is Number) {
```
oxisto, line 478: You should pay 5 € into our deposit for bad code for this!

Please use something like
```kotlin
                val literalValue = (node.subscriptExpression as? Literal<*>)?.value
                granularity =
                    when (literalValue) {
                        is Number -> indexed(literalValue)
                        is String -> indexed(literalValue)
                        else -> partial(node.subscriptExpression)
                    }
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -498,6 +439,101 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : EOGStarterPass
...
+                        PowersetLattice(identitySetOf(currentNode))
+                }
+            }
+
+            AccessValues.READ -> {
+                val writtenDeclaration = (base as? Reference)?.refersTo
+
+                if (writtenDeclaration != null && subElement != null) {
+                    // We do an ugly hack here: We store a (unique) hash out of field declaration
+                    // and
```
oxisto, line 472: The comment line breaks are off, probably due to moving it. please double check.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
```diff
@@ -498,6 +439,101 @@ open class ControlFlowSensitiveDFGPass(ctx: TranslationContext) : EOGStarterPass
         return state
     }
 
+    protected fun handlePartialAccessExpression(
```
oxisto, line 442: Maybe just a quick one-line comment here, with possibly a reference to the partial granularity thingie.

oxisto (approved):
Awesome! Just some minor clean-up that would make the code a bit more manageable.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlFlowSensitiveDFGPass.kt
KuechA: We do not actually set the partial granularities here

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/types/HasType.kt
KuechA: We now have one implementation which we have to maintain but two (external) functions which set the flag. This keeps the code in the nodes a bit smaller.

PR #2089, opened 2025-02-28
Added fallback for `translationUnitForInference`

oxisto:
It seems that sometimes `source.scope?.astNode?.component` is null and I added a fallback to definitly get a component.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
@@ -420,8 +420,10 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
     fun <TypeToInfer : Node> translationUnitForInference(source: Node): TranslationUnitDeclaration {
         // The easiest way to identify the current component would be traversing the AST, but that
         // does not work for types. But types have a scope and the scope (should) have the
-        // connection to the AST
-        val component = source.scope?.astNode?.component
+        // connection to the AST. We add several fallbacks here to make sure that we have a
+        // component.
+        val component =
+            source.scope?.astNode?.component ?: source.component ?: ctx?.currentComponent
 
         // We should also make sure that the language matches
         val tu = component?.translationUnits?.firstOrNull { it.language == this }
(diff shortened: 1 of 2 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -420,8 +420,10 @@ abstract class Language<T : LanguageFrontend<*, *>> : Node() {
     fun <TypeToInfer : Node> translationUnitForInference(source: Node): TranslationUnitDeclaration {
         // The easiest way to identify the current component would be traversing the AST, but that
         // does not work for types. But types have a scope and the scope (should) have the
-        // connection to the AST
-        val component = source.scope?.astNode?.component
+        // connection to the AST. We add several fallbacks here to make sure that we have a
+        // component.
+        val component =
+            source.scope?.astNode?.component ?: source.component ?: ctx?.currentComponent
```
KuechA, line 426: Doesn't this introduce the issue that we may use a component with a wrong language and so on?

oxisto: No, because we have a check below that checks if the language matches, so in case we pick the "wrong" component, we still fail at the end; but in some cases this helps us recover. We still need to check at some point the astParents of all nodes because there still seem to be issues

KuechA: Hm, that's not really convincing. Wouldn't it make more sense to try to find a suitable component then before crashing? I.e., if there are multiple components in the result, couldn't we look for one which has a translation unit with a matching language. If nothing exists, we will still fail of course but then at least we tried everything.

Doesn't throwing the exception somehow fail to meet our design goals: https://fraunhofer-aisec.github.io/cpg/CPG/impl/design_principles/#the-cpg-transformation-should-be ?

oxisto: It still makes sense to try the current component, because we first want to do everything to keep the analysis in the current component and actually the problem that we fail here is with OUR crappy code and not the analysed code.

oxisto: I extended the error "handling" and we now fail early if we cannot find the component which is the bigger problem.

PR #2091, opened 2025-02-28
Improvements to the configuration

oxisto:
1) Attaching configuration operation to "sub"-configuration item instead of the whole config: This makes the API a little bit cleaner since we do not need to pass the conf object, which is already known and it attaches the operations to their respective counterparts, e.g. reading an option is an operation on an option now and not of the overall configuraton. It also introduces `allOps` as a way to get all the operations of the whole config again.

2) Added `ConfigurationSource` (and group, option...). This decouples parsing the configuration and source and we introduce a final pass that combines them both. This mirrors what we do for HTTP operations. The advantage of that is that we can still parse configuration files independently from the source and even check if they are used or not (e.g., whether they have any "provide" operation).


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
```diff
@@ -203,7 +203,7 @@ object Util {
         vararg arguments: Any?,
     ) {
         log.warn(
-            String.format("%s: %s", PhysicalLocation.locationLink(node.location), format),
+            String.format("%s: %s", PhysicalLocation.locationLink(node?.location), format),
```
oxisto, line 206: Note to self: this is not needed anymore.

cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/ConfigurationPassTest.kt
```diff
@@ -76,6 +79,21 @@ class ConfigurationPassTest {
 
         val conf = result.conceptNodes.filterIsInstance<Configuration>().singleOrNull()
         assertNotNull(conf, "There should be a single configuration node")
+        assertEquals(11, conf.allOps.size, "There should be 21 overall ops in the configuration")
+
+        val confSources = result.conceptNodes.filterIsInstance<ConfigurationSource>()
+        assertNotNull(confSources, "There should be three configuration source nodes")
+        assertEquals(3, confSources.size, "There should be 2 configuration source nodes")
```
fwendland, line 86:
```suggestion
        assertEquals(3, confSources.size, "There should be 3 configuration source nodes")
```

cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/ConfigurationPassTest.kt
```diff
@@ -76,6 +79,21 @@ class ConfigurationPassTest {
 
         val conf = result.conceptNodes.filterIsInstance<Configuration>().singleOrNull()
         assertNotNull(conf, "There should be a single configuration node")
+        assertEquals(11, conf.allOps.size, "There should be 21 overall ops in the configuration")
```
fwendland, line 82:
```suggestion
        assertEquals(11, conf.allOps.size, "There should be 11 overall ops in the configuration")
```

PR #2096, opened 2025-03-02
New `ConceptPass` based on tasks

oxisto:
This PR introduces a `Task`, which is a granular sub-functionality within a certain `Pass`. The idea is that big passes, such as the one dealing with concepts can divide their work into small sub-items, while still only iterating through the graph once.

In case of the `ConceptPass` this allows us to traverse the whole graph via EOG and perform different concept-related things "in-place" where they should happen. This now should make it possible to do things like, read a config, provide dynamic loading based on the config and then read a config again in a later EOG path. This was not possible with the previous approach where load config and dynamic loading where distinct passes that could only be performed once.

This PR might be too big to merge-in at once, so this could possibly be split up in multiple pieces:
- It already incorporates  #2091, so this should be merged first
- Introducing the `Task` and adjusting the ordering helper could be split off


oxisto:
Reverting this back to draft since this needs more work and possibly the split between a concept and operation pass. I will prepare a separate PR for the "Task" stuff.

PR #2097, opened 2025-03-02
Add `additionalSources` Option and contained folders as includePaths, and therefore as folders

konradweiss:
Allows to specify an `additionalSources` folder, current default is `libraries`. All folders inside of this folder are added to the include pahts and therefore constitute a component in the analysis. That folder then contains the namespace. If a components should contain the library `somelib` and the namespace should start with the libraryname `somelib`, then the folder structure currently would be `libraries/somelib/somelib/...`. 

He advantage is that the user can just use the default folders provided or specify one folder with external sources. However, the folder with additional sources has to adhere to this folder structure.

Having, a folder named `libraries` in the project directory will automatically turn on the on import parsing of dependencies by setting `loadIncludes` to true.


oxisto (approved):

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Project.kt
```diff
@@ -96,6 +97,16 @@ class AnalysisProject(
     var securityGoalsFolder: Path? = projectDir?.resolve("security-goals"),
     /** The folder where the components are located. */
     var componentsFolder: Path? = projectDir?.resolve("components"),
+    /**
+     * The folder where the additional Sources are located. The subfolders are used to create
```
oxisto, line 101: typo: Sources -> sources
should we talk about "libraries" instead of additionalSources in the "format"?
maybe just also name the variable "librariesPath"?

konradweiss: Jepp we can do so

PR #2098, opened 2025-03-02
Removing `verbose` of `dataFlow` and `executionPath`

oxisto:
It seems that the verbose flag in the DFG and EOG analysis functions was true per default so the analysis type (MAY or MUST) had no effect. This cluttered the SARIF output.

### DIFF
--- a/codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Sarif.kt
+++ b/codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Sarif.kt
@@ -43,50 +43,54 @@ import io.github.detekt.sarif4k.*
  * [Boolean] and that the [QueryTree.children] represent the individual findings.
  */
 fun QueryTree<Boolean>.toSarif(ruleID: String): List<Result> {
-    return this.children.map {
-        Result(
-            ruleID = ruleID,
-            message =
-                if (this.value) {
-                    Message(text = "Query was successful")
-                } else {
-                    Message(text = "Query failed")
-                },
-            level =
-                if (this.value) {
-                    Level.None
-                } else {
-                    Level.Error
-                },
-            kind =
-                if (this.value) {
-                    ResultKind.Pass
-                } else {
-                    ResultKind.Fail
-                },
-            locations = listOfNotNull(it.node?.toSarifLocation()),
-            stacks = it.node?.toSarifCallStack(),
-            codeFlows =
-                it.children.mapNotNull { child ->
-                    if (child.value is List<*>) {
-                        CodeFlow(
-                            threadFlows =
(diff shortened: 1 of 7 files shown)


codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Sarif.kt
```diff
@@ -43,50 +43,54 @@ import io.github.detekt.sarif4k.*
...
-                                        message = Message(text = "Thread flow"),
-                                        locations =
-                                            (child.value as List<*>)
-                                                .filterIsInstance<Node>()
-                                                .toSarifThreadFlowLocation(),
-                                    )
-                                )
-                        )
+    // Our children contain all the results (good and bad), but we only are interested in the one
+    // matching our overall result
```
KuechA, line 47: I'm not sure at all if this statement is true. If I have some locations which are ok and one which violates a rule, why would I never be interested in the good result?

oxisto: Maybe this is misleading. If we have an overall result of "false", we are only interested in the violated flows, not the "good" ones. The main reason behind this is that it otherwise clusters the SARIF file and we have currently now way to differentiate between a good and a bad flow. Therefore the flow must macht the overall kind/pass situation.

oxisto: Hm maybe this also depends on must or may?
- If the analysis is MAY and I find paths where it the assumption holds and 10 branches where it does not -> I only want the good branch
- If the analysis is MAY and I find paths where the assumption does not hold -> I probably only want the bad branch
- If the analysis is MUST then I am probably interested in all result

Shouldn't this actually be done in the "dataflow" function? I am a bit confused that these results are here in the first place.

KuechA: I'm wondering if we actually want/need this flag. It doesn't harm I guess but I also assume that nobody will use it

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Sarif.kt
```diff
@@ -43,50 +43,54 @@ import io.github.detekt.sarif4k.*
...
+    return this.children
+        .filter { it.value == true }
```
KuechA, line 49: Shouldn't this be `it.value == this.value` based on the comment? Is it still required after updating the default value for `verbose`? And I agree, the expected output probably depends on the analysis type (must/may) most of the time.

KuechA (approved):

PR #2105, opened 2025-03-06
implementing a helper function to work with dataFlow results

maximiliankaul:
@KuechA can you support here? The internal variables need some re-naming (what's going on here)? And all the casting and `map` doesn't feel right... But it seems to work...


KuechA:
Thinking about it, maybe we can re-use some stuff which we have for the sarif-printing.

PR #2110, opened 2025-03-07
Initial work on moving creation of language to translation manager

oxisto:


cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/testcases/FlowQueriesTest.kt
```diff
@@ -61,7 +62,7 @@ class FlowQueriesTest {
                     .defaultPasses()
                     .registerPass<ControlDependenceGraphPass>()
                     .registerPass<ProgramDependenceGraphPass>()
-                    .registerLanguage(TestLanguage("."))
```
KuechA, line 64: Can we please avoid such changes within a PR having some actual changes wrt. the functionality in the future? It makes it much more time-consuming to review the PR and changes.

oxisto: As discussed, this is unfortunately necessary because we changed the creation point from the config to the manager.

konradweiss (approved):
LGTM, I can imagine though that we have to be careful with the EmptyTranslationContext, and for the Language holding it not being used.

oxisto:
> LGTM, I can imagine though that we have to be careful with the EmptyTranslationContext, and for the Language holding it not being used.

Yes the `EmptyTranslationContext` is a ugly workaround that is hopefully gone once the ctx stuff is fully implemented.

PR #2113, opened 2025-03-08
Work in progress: Splitting up scope manager into smaller API pieces

oxisto:


oxisto:
Too many interfaces, Ill try a different approach.

PR #2114, opened 2025-03-08
Revamp of node system and injecting `ctx` everywhere

oxisto:
This is a HUGE PR and I might need to split this into smaller pieces, but I am not sure if this is possible. Several things are happening here

## Introduction of several intermediate `Node` sub-classes:

Not all our nodes need complete access to all graph edges (such as the EOG, etc.). For example, currently it is possible to set EOG edges on types, scope and language nodes, which does not make sense at all. There, this PR introduces three new node classes: `EvaluatedNode` which is the basis for the EOG, `DataflowNode`, which extends the EOG node and adds dataflow-related edges and finally `AstNode` which is the basis of all nodes that belong to the AST. (Currently, types are still AST nodes, but I want to entangle this as well).


oxisto:
Closed in favour of #2126

PR #2118, opened 2025-03-10
Moving `ValueEvaluator` to `cpg-core`

oxisto:
This PR moves `ValueEvaluator` to `cpg-core` and makes it possible for a language to override the default.


oxisto:
Meh, this move now makes a simple test that prints AST nodes flakey... :(

oxisto:
> Meh, this move now makes a simple test that prints AST nodes flakey... :(

Ok the test was crap... It should have never passed. We were just lucky with the hash code so far.

PR #2135, opened 2025-03-13
Do not take inferred namespaces into account when building dependencies

oxisto:


KuechA (approved):
I really like this hack :)

PR #2149, opened 2025-03-17
Remove `cpg-console`

oxisto:
Unfortunarly, ki-shell is unmaintained and we do not have the possiblity to offer a suitable CLI shell at the moment. We will explore a future web-based version in #2139

This will finally unblock #1867 and #1872


KuechA (approved):
We should also update the figure which we have in the docs but we may just replace it with the web-frontend. Apart from that, it looks fine.

PR #2151, opened 2025-03-17
Concepts/Operations remove not needed duplicate check

maximiliankaul:
This PR partially reverts a previous change that added a check to see if a `WriteFile` node already exists before creating a new one. Such a check is unnecessary unless the `ConceptPass` mistakenly processes the same `CallExpression` multiple times, which would indicate another bug. If this check were required, it should be applied generally to all Operation nodes, not just `WriteFile`, and handled in a more systematic way.


KuechA:
Have you systematically tested that the issue of duplicated concept/operation nodes never occurs?

KuechA (approved):

PR #2152, opened 2025-03-17
Fix EOG starters for Namespace declarations

### LINKED ISSUE #2121: EOG after with does not continue
In Python, we model a `with` statement as a `TryStatemtent`. There seems to be an error connecting the `else` block of the `TryStatment` with nodes after the `with` statement.

```python
with open('foo.txt') as file:
    file.read()

print("Done")
```
There should be an EOG path from `file.read()` to `print("Done")`.

See #2120 for a failing test.

konradweiss:
This PR replicates some of the changes attempted in #2122 and correcting an issue that appeared then.

The issue was that we split the EOG creation in the not static statements and the static statements. But we also made them default static, which was not correct in this case as we assumed them to be static if they were no block with a isStatic field.
Closes #2121 

In this PR make the namespace declaration the only eog starter inside of the namespace declaration. All statements and declarations contained that are part of the statement holders statements are added to the same EOG path, the contained functions are the start of their own EOG path.


oxisto:
Should be obsolete by #2196

PR #2180, opened 2025-03-24
Remove redundant EOGStarters when executing an `EOGStarterPass`

### LINKED ISSUE #2176: Fix `EOGStarterPass`
Currently, the `EOGStarterPass` will consider all EOG starters, whether the are part of a "bigger" EOG starter chain or not. Instead we need to check if the EOG starter has a prev EOG and ignore those. Similar code already exists in the `SymbolResolver` and we need to transfer it to the `EOGStarterPass`.

KuechA:
Fixes #2176 by moving the logic filtering EOG starters without incoming EOG edges to `executePass`.

Suggested PR chain: This PR, then #2188 , then #2187 .


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/OrderingHelper.kt
```diff
@@ -0,0 +1,165 @@
...
+            dependencies.remove(nextFunction)
+            // It's no longer an unsatisfied dependency.
+            dependencies.forEach { (_, v) -> v.remove(nextFunction) }
+        }
+        orderedList.addAll(nextFunctions.sortedBy { it.name })
+    }
+
+    // We add all things which are not a function declaration to the end because they won't be
+    // called at a specific point in time (we hope)
+    orderedList.addAll(noFunction)
```
KuechA, line 163: This is rather tricky. Could be at the beginning or here... Not sure what's best

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -228,7 +228,7 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
             }
 
         // Find a list of candidate symbols. In most cases, we can just perform a lookup by name
-        // which either performs an unqualified lookup beginning from the current scope "up-wards",
+        // which either performs an unqualified lookup beginning from the current scope "upwards",
```
oxisto, line 231: Can we make the SymbolResolver an EOG starter pass then? We should probably also move the TU/Import ordering logic to the `Pass` class then.

KuechA: I'm trying to implement this in #2188 and #2187 but there's an issue in the go integration tests which I don't understand yet.

KuechA: The issue with the SymbolResolver is fixed now

oxisto (approved):

PR #2183, opened 2025-03-25
Webconsole: manually add concepts

maximiliankaul:
This adds a new feature to the webconsole: at AST-overlay view one can now click on a node and manually add a `Concept` node.
This feature will be (more) useful once we have a "query" functionality in the webconsole.

- [x] Svelte / npm errors/warnings need to be resolved @oxisto 
- [x] Don't reload page on "concept added"
- [x] Resolve "TODOs"
- [x] Download YAML as file


KuechA (approved):
I guess the changes are ok even if I'm not familiar with the UI magic

PR #2187, opened 2025-03-27
Make the `SymbolResolver` an `EOGStarterPass`

KuechA:


KuechA:
> There seems to be an issue with the integration tests for go which somehow do not end. This needs further investigation.

The issue is fixed. The problem is that we create a new instance of the `SymbolResolver` for each EOG Starter and with this in mind, the cached templates were lost. As they are now kept in the companion object (which survives the individual instances of the `SymbolResolver`), the bug is fixed. We should probably keep this in mind for other passes too: Caching information in the pass' object itself won't lead to the intended/expected result and not cache anything. This is particularly problematic for EOGStarterPasses.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -137,34 +137,28 @@ open class SymbolResolver(ctx: TranslationContext) : ComponentPass(ctx) {
...
+        walker.iterate(eogStarter)
     }
 
     override fun cleanup() {
         templateList.clear()
     }
 
     /** This function caches all [TemplateDeclaration]s into [templateList]. */
-    private fun cacheTemplates(component: Component) {
+    private fun cacheTemplates(component: Component?) {
```
KuechA, line 148: I'm not really sure if we still need to get the templates for the `Component` or if it would be sufficient to fetch/cache those of the current `EOGStarter`.

oxisto: I hve no idea and I dont want to touch the template code :D So if this makes the tests green -> go for it.

PR #2190, opened 2025-03-28
Stop iterating the EOG if no more node can fulfill the requirement

KuechA:


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -357,11 +358,22 @@ fun Node.alwaysFlowsTo(
                     nodeToTrack in n.prevDFG &&
                     (n as? Reference)?.access == AccessValues.WRITE)
         }
+        val eogScope =
+            if (scope is Interprocedural) {
+                InterproceduralWithDfgTermination(
```
konradweiss, line 363: Is it a good idea to swap out the strategy in any case the old `scope` is `Interprocedural`? Should it not be configurable or if the old `Interprocedural` scope should always use this DFG termination criterion, should we then replace `Interprocedural` with this?

konradweiss (approved):
Generally ok, as it does what it says, I am only questioning whether we want to use theDFG termination strategy for all `Interprocedural`s, and if so, can  we make it explicitly

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
KuechA: I'm also not completely happy with this. Initially, I thought about placing this in the `earlyTermination` but I don't have all data available there that I need to take this decision.

I can add another flag to the function to disable this behavior if that makes you happier. Maybe we could configure the function directly with this strategy as it should not have an impact on following the DFGs. I could test that if you want

konradweiss: Ok, the flag is default true, but at least one can disable it. I am content.

PR #2196, opened 2025-03-31
Fixes some missing EOG Edges and solves and issue where declarations were added twice to the AST  and enable CFSDFG also for non-function declarations

### LINKED ISSUE #2121: EOG after with does not continue

konradweiss:
This PR does two things:

- It properly considers EOGStarterHolders when building the EOG
- It adds newly build declarations in the PythonAddDeclarationPass to either the local AST(for `VariableDeclaration`s) or the AST of the scoped node(for `FieldDeclarations`)

Both changes are necessary together as the EOG changes lead to issues if the AST is not a proper tree.

Closes #2121

### DIFF
--- a/codyze-console/src/main/webapp/pnpm-lock.yaml
+++ b/codyze-console/src/main/webapp/pnpm-lock.yaml
@@ -46,7 +46,7 @@ importers:
         version: 10.1.1(eslint@9.23.0(jiti@2.4.2))
       eslint-plugin-svelte:
         specifier: ^3.0.0
-        version: 3.3.3(eslint@9.23.0(jiti@2.4.2))(svelte@5.25.2)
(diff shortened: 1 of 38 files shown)


konradweiss:
There is one last change I forgot to carry over from previous tests where namespaces are added twice to the AST. Once in `namespaces` and once in `declarations`

KuechA (approved):

PR #2200, opened 2025-04-01
Read Concepts from YAML

maximiliankaul:
- [x] ~Support `Operation` nodes, too~
- [x] Resolve TODOs
- [x] Match file: absolute / relative

Needs #2201


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/RegionUtils.kt
```diff
@@ -66,3 +70,22 @@ fun getCodeOfSubregion(
     end = min(end, code.length)
     return code.substring(start, end)
 }
+
+// TODO: not sure if this is the right place for this function
```
KuechA, line 74: Remove TODO. I think it's ok here.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+@ExecuteBefore(ConceptPass::class, softDependency = true)
+class LoadPersistedConcepts(ctx: TranslationContext) : TranslationResultPass(ctx) {
+
+    /**
+     * The [PassConfiguration] enabling the user to configure the persisted concepts to be loaded by
+     * setting [persistedConcepts].
+     *
+     * @param persistedConcepts A list of files containing the persisted concepts to be loaded.
+     */
+    class Configuration(var persistedConcepts: List<File> = listOf()) : PassConfiguration()
```
KuechA, line 68: could this be a `val`?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+                mapper.readValue<PersistedConceptsEntry>(file)
+            } catch (ex: Exception) {
+                logger.error(
+                    "Error reading persisted concepts from ${file.path}: ${ex.message}",
+                    ex,
+                )
+                return
+            }
+
+        entries.concepts?.let { concepts ->
```
KuechA, line 108: can be simplified to `entries.concepts?.forEach { concept ->`

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+
+    /**
+     * The [PassConfiguration] enabling the user to configure the persisted concepts to be loaded by
+     * setting [conceptSummaryFiles].
+     *
+     * @param conceptSummaryFiles A list of files containing the persisted concepts to be loaded.
+     */
+    class Configuration(var conceptSummaryFiles: List<File> = listOf()) : PassConfiguration()
+
+    val logger: Logger = LoggerFactory.getLogger(LoadPersistedConcepts::class.java)
```
oxisto, line 70: Shouldn't this already exist by the `Pass` class?  named "log"?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+@ExecuteBefore(ConceptPass::class, softDependency = true)
+class LoadPersistedConcepts(ctx: TranslationContext) : TranslationResultPass(ctx) {
+
+    /**
+     * The [PassConfiguration] enabling the user to configure the persisted concepts to be loaded by
+     * setting [conceptSummaryFiles].
+     *
+     * @param conceptSummaryFiles A list of files containing the persisted concepts to be loaded.
+     */
+    class Configuration(var conceptSummaryFiles: List<File> = listOf()) : PassConfiguration()
```
oxisto: just `conceptFiles`?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+    /**
+     * This function retrieves the nodes matching the provided [signature] from the
+     * [translationResult]. Currently, this function only matches on the node being a
+     * [CallExpression] and the [CallExpression.reconstructedImportName] matching the provided FQN.
+     *
+     * @param translationResult The [TranslationResult] to search for nodes.
+     * @param signature The [SignatureEntry] containing the signature to match.
+     * @return A list of nodes matching the provided [SignatureEntry].
+     */
+    private fun getNodesBySignature(
```
oxisto, line 143: Could be made an extension function of `TranslationResult`
```suggestion
    private fun TranslationResult.getNodesBySignature(
```

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+    /**
+     * This function retrieves the nodes matching the provided [location] from the
+     * [translationResult].
+     *
+     * @param translationResult The [TranslationResult] to search for nodes.
+     * @param location The [LocationEntry] containing the location to match.
+     * @return A list of nodes matching the provided [LocationEntry] or an empty list of the
+     *   [location] cannot be parsed.
+     */
+    private fun getNodesByLocation(
```
oxisto, line 159: same here

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+     * This class represents a single entry in the YAML/JSON file. It contains the concept and
+     * optionally a location and/or a signature entry. The pass itself enforces that exactly one of
+     * location or signature is set.
+     *
+     * @param concept The [ConceptEntry] containing the concept information.
+     * @param location The [LocationEntry] containing the [PhysicalLocation] information to match
+     *   against.
+     * @param signature The [SignatureEntry] containing the signature of nodes to match.
+     */
+    private data class ConceptRootEntry(
```
oxisto, line 244: Hmm needs a better name. Maybe `PersistedConceptEntry` for this and `PersistedConceptEntryList` or `PersistedConcepts` for the outer one? (in line 232)

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+                    ?: emptyMap(),
+            connectDFGUnderlyingNodeToConcept =
+                concept.dfg?.fromThisNodeToConcept ?: false, // TODO: this `?: false` is not nice
+            connectDFGConceptToUnderlyingNode =
+                concept.dfg?.fromConceptToThisNode ?: false, // TODO: this `?: false` is not nice
+        )
+    }
+
+    /** The root node of our YAML/JSON structure. It contains a list of [ConceptRootEntry]s. */
+    private data class PersistedConceptsEntry(val concepts: List<ConceptRootEntry>?)
```
oxisto, line 232: This is not really the entry but this is actually the "root", so the naming is confusing.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+    )
+
+    /**
+     * This function retrieves the [Node.calls] from the [TranslationResult] by their fully
+     * qualified name (FQN). The match is performed on the [CallExpression.reconstructedImportName].
+     *
+     * @param fqn The fully qualified name of the calls to retrieve.
+     * @return A list of [CallExpression] nodes matching the provided FQN.
+     */
+    private fun TranslationResult.getCallsByFQN(fqn: String): List<Node> {
```
KuechA, line 315: This should be part of the `Extensions` as it looks like a useful functionality in many cases. I would have assumed that this should even be exposed as a shortcut as part of the `calls(...)` and `calls[...]` behavior but probably it got lost with recent changes of how the name is created.

oxisto (approved):
Some minor things, but otherwise looks good!

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+
+    /**
+     * This class represents a single signature entry in the YAML/JSON file.
+     *
+     * @param fqn The fully qualified name of the [CallExpression] to match against. E.g.
+     *   `foo.bar.baz`.
+     */
+    private data class SignatureEntry(
+        val fqn: String
+        // Extend this class with arguments / types as needed in the future
```
KuechA, line 305: This comment is rather useless

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+        val fromConceptToThisNode: Boolean?,
+    )
+
+    /**
+     * This class represents a single constructor argument entry in the YAML/JSON file.
+     *
+     * @param name The name of the constructor argument.
+     * @param value The value of the constructor argument.
+     */
+    private data class ConstructorArgumentEntry(val name: String, val value: String)
```
KuechA, line 285: Can this be aligned with the code used in #2183? It looks like at least the ConstructorArguments are somehow redundant.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+    private fun addConcept(underlyingNode: Node, concept: ConceptEntry) {
+        logger.debug("Adding concept {} to node {}.", concept, underlyingNode)
+        underlyingNode.conceptBuildHelper(
+            name = concept.name,
+            underlyingNode = underlyingNode,
+            constructorArguments =
+                concept.constructorArguments?.associate { arg -> arg.name to arg.value }
+                    ?: emptyMap(),
+            connectDFGUnderlyingNodeToConcept =
+                concept.dfg?.fromThisNodeToConcept ?: false, // TODO: this `?: false` is not nice
```
KuechA, line 225: Remove the TODO. Can't we use a default value for the dfg entries if nothing is provided by the user?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+     * [translationResult]. Currently, this function only matches on the node being a
+     * [CallExpression] and the [CallExpression.reconstructedImportName] matching the provided FQN.
+     *
+     * @param translationResult The [TranslationResult] to search for nodes.
+     * @param signature The [SignatureEntry] containing the signature to match.
+     * @return A list of nodes matching the provided [SignatureEntry].
+     */
+    private fun getNodesBySignature(
+        translationResult: TranslationResult,
+        signature: SignatureEntry,
```
KuechA, line 145: Rather than having the `getNodesBySignature` and `getNodesByLocation` as member of the pass, couldn't we let both classes inherit from a common class `NodeSelector` (name TBD) which has a function `getNodes(translationResult: TranslationResult): List<Node>`? This would shift the logic of selecting the nodes from the pass to these classes and maybe allow a somewhat cleaner interface. As we do not need any information from the `Pass` object, this should be easy to implement. It may even allow us to re-use the selection logic in other parts of the code later-on (e.g. to select targets of DFG edges or similar in a more generic way)

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+     */
+    private fun getNodesByLocation(
+        translationResult: TranslationResult,
+        location: LocationEntry,
+    ): List<Node> {
+        val loc =
+            try {
+                val regex =
+                    Regex(
+                        "(?<startLine>\\d+):(?<startColumn>\\d+)-(?<endLine>\\d+):(?<endColumn>\\d+)"
```
KuechA, line 167: Is there an existing parser? Otherwise, we could consider specifying the attributes `startLine`, `startColumn`, `endLine`, and `endColumn` as part of the yml. This would remove parsing the string here and it would also help clarify the format for the user (the numbers and the cryptic format are probably more error-prone)

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
```diff
@@ -0,0 +1,318 @@
...
+     * This class represents the DFG connections to be created between the concept and the
+     * underlying node.
+     *
+     * @param fromThisNodeToConcept Whether to add a DFG connection from the underlying node to the
+     *   concept node.
+     * @param fromConceptToThisNode Whether to add a DFG connection from the concept node to the
+     *   underlying node.
+     */
+    private data class DFGEntry(
+        val fromThisNodeToConcept: Boolean?,
```
KuechA, line 275: can we make these non-null and specify a default value?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
oxisto: Yes!

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
oxisto: I like the condensed format

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
KuechA: Well, in both cases we store similar/the same(?) information, so I thought we could maybe merge the two classes. Also, when we generate a yaml file later, this may be useful

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
KuechA: I can live with either of them

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/LoadPersistedConcepts.kt
oxisto: Slightly different syntax and we need to store more information in the other one such as the type, etc. Let's keep them separate and maybe harmonize later.

PR #2204, opened 2025-04-02
Set AST parent to new node when replacing a node

KuechA:
Somehow, replacing a node (here: `CallExpression`s) results in the ast children still pointing to the old parent even if they are now part of the new node. This PR sets this relationship manually. However, in the long-term, I'd encourage a different solution.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/SubgraphWalker.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/SubgraphWalker.kt
@@ -455,6 +455,7 @@ private fun CallExpression.duplicateTo(call: CallExpression, callee: Reference) 
 fun MemberCallExpression.toCallExpression(callee: Reference): CallExpression {
     val call = CallExpression()
     duplicateTo(call, callee)
+    call.arguments.forEach { it.astParent = call }
 
     return call
 }
@@ -462,6 +463,7 @@ fun MemberCallExpression.toCallExpression(callee: Reference): CallExpression {
 fun CallExpression.toMemberCallExpression(callee: MemberExpression): MemberCallExpression {
     val call = MemberCallExpression()
     duplicateTo(call, callee)
+    call.arguments.forEach { it.astParent = call }
 
     return call
 }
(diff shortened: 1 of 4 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/SubgraphWalker.kt
```diff
@@ -455,20 +455,23 @@ private fun CallExpression.duplicateTo(call: CallExpression, callee: Reference)
...
     duplicateTo(call, callee)
+    call.arguments.forEach { it.astParent = call }
 
     return call
 }
 
 fun CallExpression.toMemberCallExpression(callee: MemberExpression): MemberCallExpression {
     val call = MemberCallExpression()
     duplicateTo(call, callee)
+    call.arguments.forEach { it.astParent = call }
```
konradweiss, line 466: Isnt't this the same thing we are doing at the callee of this function? If we look at the internals of dublicateTo, they appear to copy the argumentEdges. Would it be possible to set the ast Parent in there?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/SubgraphWalker.kt
```diff
@@ -455,20 +455,23 @@ private fun CallExpression.duplicateTo(call: CallExpression, callee: Reference)
...
 
     return call
 }
 
 fun CallExpression.toConstructExpression(callee: Reference): ConstructExpression {
     val construct = ConstructExpression()
     duplicateTo(construct, callee)
+    construct.arguments.forEach { it.astParent = construct }
```
konradweiss, line 474: if we look at the callsite of `toConstructExpression` , however, it appears to be missing

konradweiss (approved):
I would rather these inside of the `dublicateTo` implementation

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/SubgraphWalker.kt
oxisto: it was also my impression that we (should) already do this inside the duplicate function. Let me have a Quick Look before we merge this please.

KuechA: It sets the ast parent only in some cases

PR #2211, opened 2025-04-04
Trying to implement EOG-based `ConceptsPass`es

KuechA:
This PR provides a common interface for `ConceptPasses` which rely on iterating the EOG to identify (all possible) concepts and operations which have to be added to a `Node`.

Problem: The EOG is likely to visit the same node multiple times. This can result in creating equal overlay nodes multiple times for the same underlying node. Currently, there is no good way to remove these again as they may be connected to various other nodes.

### DIFF
--- a/cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/file/FileConceptTest.kt
+++ b/cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/file/FileConceptTest.kt
@@ -28,7 +28,7 @@ package de.fraunhofer.aisec.cpg.concepts.file
 import de.fraunhofer.aisec.cpg.frontends.python.PythonLanguage
 import de.fraunhofer.aisec.cpg.graph.*
 import de.fraunhofer.aisec.cpg.graph.concepts.file.*
-import de.fraunhofer.aisec.cpg.passes.concepts.file.python.PythonFileConceptPass
+import de.fraunhofer.aisec.cpg.passes.concepts.file.python.PythonFileEOGConceptPass
 import de.fraunhofer.aisec.cpg.query.Must
 import de.fraunhofer.aisec.cpg.query.dataFlow
 import de.fraunhofer.aisec.cpg.query.executionPath
@@ -53,7 +53,7 @@ class FileConceptTest : BaseTest() {
                 usePasses = true,
             ) {
                 it.registerLanguage<PythonLanguage>()
-                it.registerPass<PythonFileConceptPass>()
+                it.registerPass<PythonFileEOGConceptPass>()
                 it.symbols(mapOf("PYTHON_PLATFORM" to "linux"))
             }
         assertNotNull(result)
@@ -130,7 +130,7 @@ class FileConceptTest : BaseTest() {
                 usePasses = true,
             ) {
                 it.registerLanguage<PythonLanguage>()
-                it.registerPass<PythonFileConceptPass>()
+                it.registerPass<PythonFileEOGConceptPass>()
                 it.symbols(mapOf("PYTHON_PLATFORM" to "linux"))
             }
         assertNotNull(result)
@@ -183,7 +183,7 @@ class FileConceptTest : BaseTest() {
                 usePasses = true,
             ) {
                 it.registerLanguage<PythonLanguage>()
-                it.registerPass<PythonFileConceptPass>()
+                it.registerPass<PythonFileEOGConceptPass>()
                 it.symbols(mapOf("PYTHON_PLATFORM" to "linux"))
             }
         assertNotNull(result)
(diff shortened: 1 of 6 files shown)


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/Operation.kt
```diff
@@ -49,5 +49,15 @@ abstract class Operation(
         return other is Operation && super.equals(other) && other.concept == this.concept
     }
 
+    override fun equalWithoutUnderlying(other: OverlayNode): Boolean {
```
oxisto, line 52: I think we can exclude the underlying node in the default equals instead of having this very very weird piece of code.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/GenericConceptBuilder.kt
```diff
@@ -43,6 +43,21 @@ inline fun <reified ConceptClass : Concept> MetadataProvider.newConcept(
         NodeBuilder.log(this)
     }
 
+/**
+ * This function creates a new [Concept] node based on [ConceptClass]. It is neither connected by
+ * the EOG nor the DFG.
+ */
+inline fun <reified ConceptClass : Concept> MetadataProvider.newConceptNoConnect(
```
oxisto, line 50: I really dont like this, we should instead have a parameter to `newConcept` like `doNotConnectEOG` or something instead.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/Pass.kt
```diff
@@ -140,6 +141,25 @@ object EOGStarterLeastTUImportSorter : Sorter<Node>() {
             .toList()
 }
 
+/**
+ * First, sorts the [TranslationUnitDeclaration]s with the [LeastImportTranslationUnitSorter] and
+ * then gathers all resolution EOG starters; and make sure they really do not have a predecessor,
+ * otherwise we might analyze a node multiple times. Note that the [EOGStarterHolder]s are not
+ * sorted.
+ */
+object EOGStarterLeastTUImportSorterWithCatchAfterTry : Sorter<Node>() {
```
oxisto, line 150: What? :D

cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/file/FileConceptTest.kt
```diff
@@ -62,7 +62,10 @@ class FileConceptTest : BaseTest() {
         assertTrue(fileNodes.isNotEmpty())
```
oxisto, line 62: Did you try to re-enable the ignored test?

oxisto (approved):

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/EOGConceptPass.kt
```diff
@@ -0,0 +1,256 @@
...
+import de.fraunhofer.aisec.cpg.passes.EvaluationOrderGraphPass
+import de.fraunhofer.aisec.cpg.passes.configuration.DependsOn
+import kotlin.collections.set
+
+typealias NodeToOverlayStateElement = MapLattice.Element<Node, PowersetLattice.Element<OverlayNode>>
+
+typealias NodeToOverlayState = MapLattice<Node, PowersetLattice.Element<OverlayNode>>
+
+/**
+ * An abstract pass that is used to identify and create [Concept] and [Operation] nodes in the
```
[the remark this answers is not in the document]
KuechA: I don't think there's a limitation when it should not be used. whenever the mapping of a node `a` to a concept or operation depends on another node`b`which occurs in the EOG before `a`, it should be used.

cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/file/FileConceptTest.kt
KuechA: yes and it seems to work

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/Operation.kt
KuechA: Not sure if that's a good idea because if there is an `underlyingNode` and it's different, the overlay node should also be different

KuechA: What we could/should do instead is to implement this weird logic and just call it in `equals` together with the check for the `underlyingNode`

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/GenericConceptBuilder.kt
KuechA: Added a flag which is mandatory for all builders to prevent that somebody accidentally forgets setting it

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/Pass.kt
KuechA: Updated the documentation and the name

PR #2218, opened 2025-04-08
Experiment about a function-based `ConceptPass`

oxisto:


oxisto:
We will revisit this concept at a later stage.

PR #2219, opened 2025-04-09
Remove namespaces and includes as a persisting field. 

### LINKED ISSUE #2197: Namespaces are added twice to the AST of a TU
Translations units contain this addDeclarations function:
```
    override fun addDeclaration(declaration: Declaration) {
        if (declaration is IncludeDeclaration) {
            addIfNotContains(includeEdges, declaration)
        } else if (declaration is NamespaceDeclaration) {
            addIfNotContains(namespaceEdges, declaration)
        }
        addIfNotContains(declarationEdges, declaration)
    }
```
so namespaces and includes are added to two different lists that are saves as AST edges. Therefore, we get this type of double edges:

Image

in the future we should put the last addIfNotContains in an else clause, adapt handleTranslationUnit in the EOG Pass and fix the tests that expects to find all declarations in the declarations list. Or we should remove saving namespaces and includes in separate declarations lists.

Code example:
`cpg/cpg-language-python/src/test/resources/python/vars.py`

konradweiss:
Remove `namespaces` and `includes` as a persisting field. Includes and namespaces are still in declarations and can be filtered for. Previously, namespaces and includes were included twice in fields that were persisted as AST nodes. Leading to these nodes to be reachable by two distinct AST edges, which contradicts the tree property.

Closes #2197

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/TranslationUnitDeclaration.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/TranslationUnitDeclaration.kt
@@ -44,14 +44,12 @@ class TranslationUnitDeclaration :
     override val declarations by unwrapping(TranslationUnitDeclaration::declarationEdges)
 
     /** A list of includes within this unit. */
-    @Relationship(value = "INCLUDES", direction = Relationship.Direction.OUTGOING)
-    val includeEdges = astEdgesOf<IncludeDeclaration>()
-    val includes by unwrapping(TranslationUnitDeclaration::includeEdges)
+    val includes
+        get() = declarations.filterIsInstance<IncludeDeclaration>()
 
     /** A list of namespaces within this unit. */
-    @Relationship(value = "NAMESPACES", direction = Relationship.Direction.OUTGOING)
-    val namespaceEdges = astEdgesOf<NamespaceDeclaration>()
-    val namespaces by unwrapping(TranslationUnitDeclaration::namespaceEdges)
+    val namespaces
+        get() = declarations.filterIsInstance<NamespaceDeclaration>()
 
     /** The list of statements. */
     @Relationship(value = "STATEMENTS", direction = Relationship.Direction.OUTGOING)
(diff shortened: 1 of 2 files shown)


KuechA:
Is this really the most appropriate way to fix the issue? Couldn't we also just not add the includes and namespaces to the declarations field? Won't we miss the labels (INCLUDE and NAMESPACE) on the edges in the persisted graph (this encodes potentially interesting information I guess)?

oxisto:
> Is this really the most appropriate way to fix the issue? Couldn't we also just not add the includes and namespaces to the declarations field? Won't we miss the labels (INCLUDE and NAMESPACE) on the edges in the persisted graph (this encodes potentially interesting information I guess)?

Well... what makes includes and namespaces special so that they are not just a "declaration"? For example we could argue we would then need the same for functions or imports or variables?

oxisto:
We should however (and this is probably a bigger chance at some point) revisit the semantics of our extensions. For example we have a `namespaces` extension that gathers *all* namespaces in all sub-nodes and we now have still have a `namespaces` variable that only returns the direct namespace declarations. We could potentially have this as an extension function as well and have a semantic like:
- `tu.namespaces`: Returns the direct `NamespaceDeclaration` children
- `tu.allNamespaces`: Returns all `NamespaceDeclaration` children

but this is a potential BIG change since we have other extensions such as `functions` which are widely used and return *all* functions. We could of course also have something like:

- `tu.functions`: Returns *all* functions
- `tu.directFunctions` (needs a better name): Returns only the direct children

oxisto (approved):
I would say yes lets do it this way which is backwards compatible, but then we need to address the extensions problem at some point!

konradweiss:
> We should however (and this is probably a bigger chance at some point) revisit the semantics of our extensions. For example we have a `namespaces` extension that gathers _all_ namespaces in all sub-nodes and we now have still have a `namespaces` variable that only returns the direct namespace declarations. We could potentially have this as an extension function as well and have a semantic like:
> 
> * `tu.namespaces`: Returns the direct `NamespaceDeclaration` children
> * `tu.allNamespaces`: Returns all `NamespaceDeclaration` children
> 
> but this is a potential BIG change since we have other extensions such as `functions` which are widely used and return _all_ functions. We could of course also have something like:
> 
> * `tu.functions`: Returns _all_ functions
> * `tu.directFunctions` (needs a better name): Returns only the direct children

Yes, we should definitely keep that in mind

PR #2222, opened 2025-04-09
Resolved bug in member expression ambiguity

### LINKED ISSUE #2220: Incorrect inferred FunctionDeclarations
This [method](https://github.com/openstack/cinder/blob/ab30d6e58ad054f6aca1138b28ac216ef5449879/cinder/context.py#L208) is called in many places (typically like [here](https://github.com/openstack/cinder/blob/ab30d6e58ad054f6aca1138b28ac216ef5449879/cinder/api/v3/types.py#L44)), but they aren´t resolved propertly because of the missing type. Instead, two inferred `FunctionDeclaration` s are created (which have the `calledby` edges) but with `location = null` and different `code` each time its parsed.

oxisto:
There was a bug in resolving member expression ambiguities when the name of a member's base was identical to an import. This PR should resolve that issue.

Fixes #2220


lshala (approved):
Not much familiar with this code, but tested it and it works. LGTM, appreciate the fast fix! :rocket:

PR #2229, opened 2025-04-13
Replacing nodejs with deno in cpg-language-typescript

oxisto:
Deno is a new Rust-based JS/TS runtime engine and it seems to be quite good and fast. See https://deno.com

This PR exchanges the nodejs-backed runtime in the `cpg-language-typescript` frontend to Deno. This should make the build faster (< 3m now!) and it does not require nodejs during runtime, since we can bundle our parser into a binary (which contains a light version of the runtime)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.12.1-bin.zip
```
KuechA, line 3: why do we downgrade gradle here?

cpg-language-typescript/build.gradle.kts
```diff
@@ -37,23 +38,74 @@ mavenPublishing {
...
+            "src/main/typescript/src/parser.ts",
+        )
+        outputs.dir("build/resources/main/typescript")
         outputs.cacheIf { true }
+    }
 
-        workingDir.set(file("src/main/nodejs"))
-        pnpmCommand.set(listOf("run", "bundle"))
-        dependsOn(tasks.getByName("pnpmInstall"))
+val compileLinuxArm64 =
```
KuechA, line 92: So, there is no windows support?

cpg-language-typescript/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/typescript/TypeScriptLanguageFrontend.kt
```diff
@@ -71,20 +71,30 @@ class TypeScriptLanguageFrontend(
     private val mapper = jacksonObjectMapper()
 
     companion object {
-        private val parserFile: File = createTempFile("parser", ".js")
+        private val parserFile: File = createTempFile("parser", "")
 
         init {
-            val link = this::class.java.getResourceAsStream("/nodejs/parser.js")
+            val arch =
+                System.getProperty("os.arch").replace("aarch64", "arm64").replace("x86_64", "amd64")
```
KuechA, line 78: Why do we make these replacements instead of simply naming the binaries correctly in the `build.gradle.kts`?

gradle/wrapper/gradle-wrapper.properties
oxisto: Because of https://github.com/MaSch0212/deno-gradle-plugin/issues/4 :(

cpg-language-typescript/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/typescript/TypeScriptLanguageFrontend.kt
oxisto: Good point, I can rename the binary

cpg-language-typescript/build.gradle.kts
oxisto: Ah there is windows support (see https://docs.deno.com/runtime/reference/cli/compile/#supported-targets), so I can add that

KuechA (approved):

PR #2233, opened 2025-04-14
Move logic to list concepts and operations to `cpg-concepts`.

KuechA:


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extension.kt
```diff
@@ -47,3 +49,11 @@ val Node.conceptNodes: Set<Concept>
  */
 val Node.operationNodes: Set<Operation>
     get() = this.nodes.flatMapTo(mutableSetOf()) { it.overlays.filterIsInstance<Operation>() }
+
+/** Returns a [Set] of all subclasses of [Concept]. */
+fun listConceptClasses(): Set<Class<out Concept>> =
+    Reflections("de.fraunhofer.aisec", SubTypes).getSubTypesOf(Concept::class.java)
+
+/** Returns a [Set] of all subclasses of [Concept]. */
+fun listOperationClasses(): Set<Class<out Operation>> =
```
oxisto, line 58: Wouldnt' it make more sense to turn this into a refined generic function, so I can potentially call this on all (sub)-classes of operations/concepts,overlays,etc.

oxisto (approved):
LGTM, minor improvement suggestion

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extension.kt
KuechA: I'll change this.

I'm also not sure how we can improve this filter for the package. Get rid of it completely or make it configurable by an argument?

oxisto: Maybe as parameter with the current value as default. Also listOverlays is misleading. OverlayClasses?

PR #2237, opened 2025-04-15
Removing `equalWithoutUnderlying`

oxisto:
Instead `equals` now excludes the `underylingNode` and the EOGConceptPass collects intermediate states which are then merged into a final state and then uses `finalCleanup` to set the underlying node based on the state.


konradweiss:
@maximiliankaul Should test if this works.

PR #2239, opened 2025-04-15
Fixed problem with `finalState` in `iterateEOG`

oxisto:
If the EOG ended in a loop, we previously "forgot" to add to the final state.

### DIFF
--- a/cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/file/FileConceptTest.kt
+++ b/cpg-concepts/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/concepts/file/FileConceptTest.kt
@@ -414,15 +414,15 @@ class FileConceptTest : BaseTest() {
 
         val files = conceptNodes.filterIsInstance<File>()
         assertEquals(
-            listOf("a", "b"),
-            files.map { it.fileName },
+            setOf("a", "b"),
+            files.map { it.fileName }.toSet(),
             "Expected to find two `File` nodes (\"a\" and \"b\").",
         )
 
         val writes = conceptNodes.filterIsInstance<WriteFile>()
         assertEquals(
-            listOf("a", "b"),
-            writes.map { it.file.fileName },
+            setOf("a", "b"),
+            writes.map { it.file.fileName }.toSet(),
             "Expected to find two `WriteFile` nodes (to \"a\" and \"b\").",
         )
     }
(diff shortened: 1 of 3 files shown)


KuechA:
Could we add a test-case to check if this also works in the presence of loops? I.e., if we have code such as:
```python
file = open("a")
for bla in x:
    file.write(bla)
    file = open("b")
```

Obviously, this is sort of weird but should work to illustrate a loop and a change in the concepts/operations

KuechA:
> > > Could we add a test-case to check if this also works in the presence of loops? I.e., if we have code such as:
> > > ```python
> > > file = open("a")
> > > for bla in x:
> > >     file.write(bla)
> > >     file = open("b")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Obviously, this is sort of weird but should work to illustrate a loop and a change in the concepts/operations
> > 
> > 
> > I can add a test like that... What exactly would be the expected result here?
> 
> I am confused, this code does not produce ANY file-related concepts AT ALL. Neither with our without the changes in this PR :(

Hm, maybe the code is not exactly what the current pass expects? 

at least for the following python3 code, the operations for `write` and `close` are executed on two different file objects which means, there should probably be two write and two close operations with different `File`objects as `concept`.

```python
file = open("a", "a")
>>> for bla in ["123", "456"]:
...     file.write(bla)
...     file.close()
...     file = open("b", "a")
... file.close()
```

The code runs on my machine and afterwards, I have the contents "123" in file "a" and "456" in file "b".

oxisto:
> > > > Could we add a test-case to check if this also works in the presence of loops? I.e., if we have code such as:
> > > > ```python
> > > > file = open("a")
> > > > for bla in x:
> > > >     file.write(bla)
> > > >     file = open("b")
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Obviously, this is sort of weird but should work to illustrate a loop and a change in the concepts/operations
> > > 
> > > 
> > > I can add a test like that... What exactly would be the expected result here?
> > 
> > 
> > I am confused, this code does not produce ANY file-related concepts AT ALL. Neither with our without the changes in this PR :(
> 
> Hm, maybe the code is not exactly what the current pass expects?
> 
> at least for the following python3 code, the operations for `write` and `close` are executed on two different file objects which means, there should probably be two write and two close operations with different `File`objects as `concept`.
> 
> ```python
> file = open("a", "a")
> >>> for bla in ["123", "456"]:
> ...     file.write(bla)
> ...     file.close()
> ...     file = open("b", "a")
> ... file.close()
> ```
> 
> The code runs on my machine and afterwards, I have the contents "123" in file "a" and "456" in file "b".

The original code did not have the "close", but apart from that I have two file nodes and two write nodes, so it should work then?

KuechA (approved):

PR #2242, opened 2025-04-15
Adds Assumptions as Overlay Node

konradweiss:
This PR adds Assumption nodes as Overlay nodes to the CPG. Assumptions can be attached to a node or edge, as well as, created with an identifier when not connected to a specific node.

Further documentation of this PR will follow.

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
@@ -25,6 +25,9 @@
  */
 package de.fraunhofer.aisec.cpg.query
 
+import de.fraunhofer.aisec.cpg.assumptions.Assumption
+import de.fraunhofer.aisec.cpg.assumptions.AssumptionType
+import de.fraunhofer.aisec.cpg.assumptions.getCurrentFileAndLine
 import de.fraunhofer.aisec.cpg.evaluation.compareTo
 import de.fraunhofer.aisec.cpg.graph.Node
 
@@ -67,6 +70,12 @@ open class QueryTree<T>(
      * in [stringRepresentation].
      */
     open var node: Node? = null,
+
+    /**
+     * Assumptions can be created in the QueryTree object with the [assume] function ore by adding
+     * an assumption manually.
+     */
+    open var assumptions: MutableList<Assumption> = mutableListOf(),
 ) : Comparable<QueryTree<T>> {
     fun printNicely(depth: Int = 0): String {
         var res =
@@ -334,6 +343,34 @@ fun QueryTree<*>.successfulLastNodes(): List<Node> {
     val innerPath = successfulPaths.flatMap { it.children }
     val finallyTheEntirePaths = innerPath.map { it.value }
     return finallyTheEntirePaths.mapNotNull { (it as? List<*>)?.last() }.filterIsInstance<Node>()
+}
+
+/**
+ * @param assumptionType The type of assumption used to differentiate between assumptions and group
+ *   similar assumptions.
+ * @param aID an ID chosen by the caller to identify the assumption across translation, e.g.
+ *   filename or classname, with a function name and a potential counter.
+ * @param scope The scope that the assumption has validity for, here the scope is a node, because
+ *   the assumption is valid for every node in its ast subtree.
+ * @param message The message describing the assumption that was taken.
+ */
+fun <T> QueryTree<T>.assume(
+    assumptionType: AssumptionType,
+    message: String,
+    aID: String?,
+    scope: Node? = null,
+): QueryTree<T> {
+    // This connects the assumption as an overlay node to the code graph
+    this.assumptions.add(
+        Assumption(
+            assumptionType,
+            message,
+            getCurrentFileAndLine(),
+            aID = aID,
+            assumptionScope = scope,
+        )
+    )
+    return this
 }
 
 sealed interface TerminationReason {
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
@@ -1,29 +1,75 @@
 package de.fraunhofer.aisec.cpg.assumptions
 
+import de.fraunhofer.aisec.cpg.graph.ContextProvider
+import de.fraunhofer.aisec.cpg.graph.Name
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.OverlayNode
 
 class Assumption(assumptionType: AssumptionType,
                  node: Node,
                  scope: Node = node,
-                 fileAndLine: String,
+                 assumptionLocation: String, // If the assumption was made the CPG sourcecode, the cpg sourcecode locations is used
                  message: String) : OverlayNode() {
 
     init {
         super.underlyingNode = scope
+        name = Name(assumptionType.name)
+        location = node.location
     }
 
+    companion object {
+        /**
+         *
+         * @param assumptionType The type of assumption used to differentiate between assumptions and group similar assumptions.
+         *
+         * @param node The node that cause the assumption to be necessary, even if the assumption has validity for the
+         * entire program, the node serves as location to be reported to the user.
+         *
+         * @param scope The scope that the assumption has validity for, here the scope is a node, because the assumption is
+         * valid for every node in its ast subtree.
+         *
+         * @param message The message describing the assumption that was taken.
+         */
+        public fun assume(
+            assumptionType: AssumptionType,
+            node: Node,
+            scope: Node = node,
+            message: () -> String,
+        ) {
+            // This connects the assumption as an overlay node to the code graph
+            Assumption(assumptionType, node, scope, getCurrentFileAndLine(), message.toString())
+        }
+
+
+        private fun getCurrentFileAndLine(): String {
+            val stackTrace = Thread.currentThread().stackTrace
+            return "File: ${stackTrace[3].fileName}, Line: ${stackTrace[3].lineNumber}"
+        }
+    }
 
 }
 
 enum class AssumptionType {
-    InferenceAssumption, ClosedMacroAssumption, UnsupportedLanguageProblem, MissingCodeProblem,
+    InferenceAssumption,
+    ClosedMacroAssumption,
+    UnsupportedLanguageProblem,
+    MissingCodeProblem,
     AmbiguityAssumption,
-    ConceptPlacementAssumption, ExhaustivEnumerationAssumption,
-    CompletenessAssumption, SoundnessAssumption,
-    CFIntegrityAssumption, NoExceptionsAssumption, CFAllOrNothingExecutesAssumption, TrustedConfigAssumption, ExternalDataAssumption,
-    NetworkAvailableAssumption, ResourceExistsAssumption, ServiceReachableAssumption,
+    ConceptPlacementAssumption,
+    ExhaustiveEnumerationAssumption,
+    CompletenessAssumption,
+    SoundnessAssumption,
+    CFIntegrityAssumption,
+    NoExceptionsAssumption,
+    CFAllOrNothingExecutesAssumption,
+    TrustedConfigAssumption,
+    ExternalDataAssumption,
+    NetworkAvailableAssumption,
+    ResourceExistsAssumption,
+    ServiceReachableAssumption,
     AtomicExecutionAssumption,
-    TrustBoundaryAssumption, DataRangeAssumption, TrustedInputAssumption,
+    TrustBoundaryAssumption,
+    DataRangeAssumption,
+    TrustedInputAssumption,
 }
 
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
@@ -1,0 +1,29 @@
+package de.fraunhofer.aisec.cpg.assumptions
+
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.OverlayNode
+
+class Assumption(assumptionType: AssumptionType,
+                 node: Node,
+                 scope: Node = node,
+                 fileAndLine: String,
+                 message: String) : OverlayNode() {
+
+    init {
+        super.underlyingNode = scope
+    }
+
+
+}
+
+enum class AssumptionType {
+    InferenceAssumption, ClosedMacroAssumption, UnsupportedLanguageProblem, MissingCodeProblem,
+    AmbiguityAssumption,
+    ConceptPlacementAssumption, ExhaustivEnumerationAssumption,
+    CompletenessAssumption, SoundnessAssumption,
+    CFIntegrityAssumption, NoExceptionsAssumption, CFAllOrNothingExecutesAssumption, TrustedConfigAssumption, ExternalDataAssumption,
+    NetworkAvailableAssumption, ResourceExistsAssumption, ServiceReachableAssumption,
+    AtomicExecutionAssumption,
+    TrustBoundaryAssumption, DataRangeAssumption, TrustedInputAssumption,
+}
+
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
@@ -1,52 +1,63 @@
+/*
+ * Copyright (c) 2025, Fraunhofer AISEC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *                    $$$$$$\  $$$$$$$\   $$$$$$\
+ *                   $$  __$$\ $$  __$$\ $$  __$$\
+ *                   $$ /  \__|$$ |  $$ |$$ /  \__|
+ *                   $$ |      $$$$$$$  |$$ |$$$$\
+ *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
 package de.fraunhofer.aisec.cpg.assumptions
 
-import de.fraunhofer.aisec.cpg.graph.ContextProvider
 import de.fraunhofer.aisec.cpg.graph.Name
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.OverlayNode
+import org.neo4j.ogm.annotation.Relationship
 
-class Assumption(assumptionType: AssumptionType,
-                 node: Node,
-                 scope: Node = node,
-                 assumptionLocation: String, // If the assumption was made the CPG sourcecode, the cpg sourcecode locations is used
-                 message: String) : OverlayNode() {
+class Assumption(
+    val assumptionType: AssumptionType,
+    node: Node?,
+    assumptionScope: Node? = node,
+    // If the assumption was made the CPG sourcecode, the cpg sourcecode locations is used
+    var assumptionLocation: String,
+    var message: String,
+    var status: AssumptionStatus = AssumptionStatus.Unconfirmed,
+) : OverlayNode() {
+
+    @Relationship(value = "ASSUMPTION_SCOPE", direction = Relationship.Direction.OUTGOING)
+    /**
+     * The Scope in which this assumption has validity. It is equal to [underlyingNode] or can be
+     * connected to its ancestor nodes in the AST if the scope is wider.
+     */
+    val assumptionScope = assumptionScope
 
     init {
-        super.underlyingNode = scope
+        super.underlyingNode = node
         name = Name(assumptionType.name)
-        location = node.location
+        location = node?.location
     }
+}
 
-    companion object {
-        /**
-         *
-         * @param assumptionType The type of assumption used to differentiate between assumptions and group similar assumptions.
-         *
-         * @param node The node that cause the assumption to be necessary, even if the assumption has validity for the
-         * entire program, the node serves as location to be reported to the user.
-         *
-         * @param scope The scope that the assumption has validity for, here the scope is a node, because the assumption is
-         * valid for every node in its ast subtree.
-         *
-         * @param message The message describing the assumption that was taken.
-         */
-        public fun assume(
-            assumptionType: AssumptionType,
-            node: Node,
-            scope: Node = node,
-            message: () -> String,
-        ) {
-            // This connects the assumption as an overlay node to the code graph
-            Assumption(assumptionType, node, scope, getCurrentFileAndLine(), message.toString())
-        }
-
-
-        private fun getCurrentFileAndLine(): String {
-            val stackTrace = Thread.currentThread().stackTrace
-            return "File: ${stackTrace[3].fileName}, Line: ${stackTrace[3].lineNumber}"
-        }
-    }
-
+enum class AssumptionStatus {
+    Rejected,
+    Confirmed,
+    Unconfirmed,
 }
 
 enum class AssumptionType {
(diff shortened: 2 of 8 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,144 @@
...
+import de.fraunhofer.aisec.cpg.graph.Name
+import de.fraunhofer.aisec.cpg.graph.Node
+import de.fraunhofer.aisec.cpg.graph.OverlayNode
+import de.fraunhofer.aisec.cpg.graph.edges.Edge
+import de.fraunhofer.aisec.cpg.persistence.DoNotPersist
+import org.neo4j.ogm.annotation.Relationship
+
+/**
+ * The minimal properties to identify an assumption, is the assumption type and either node, edge,
+ * or reference Class. Only one of the three should be provided.
```
oxisto, line 37: reference class?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,144 @@
...
+ */
+class Assumption(
+    val assumptionType: AssumptionType,
+    var message: String,
+    // If the assumption was made the CPG sourcecode, the cpg sourcecode locations is used
+    var assumptionLocation: String,
+    node: Node? = null,
+    @DoNotPersist var edge: Edge<*>? = null,
+    var aID: String? = null,
```
oxisto, line 46: what's aID?!

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,144 @@
...
+import org.neo4j.ogm.annotation.Relationship
+
+/**
+ * The minimal properties to identify an assumption, is the assumption type and either node, edge,
+ * or reference Class. Only one of the three should be provided.
+ */
+class Assumption(
+    val assumptionType: AssumptionType,
+    var message: String,
+    // If the assumption was made the CPG sourcecode, the cpg sourcecode locations is used
```
oxisto, line 42: should be a docstring. What kind of location is that? Why is it a string and not a (SARIF) location?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,144 @@
...
+    assumptionScope: Node? = node,
+    var status: AssumptionStatus = AssumptionStatus.Ignored,
+) : OverlayNode() {
+
+    @Relationship(value = "ASSUMPTION_SCOPE", direction = Relationship.Direction.OUTGOING)
+    /**
+     * The Scope in which this assumption has validity. It is equal to [underlyingNode] or can be
+     * connected to its ancestor nodes in the AST if the scope is wider.
+     */
+    var assumptionScope = assumptionScope
```
oxisto, line 56: move to constructor?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,144 @@
...
+    node: Node? = null,
+    @DoNotPersist var edge: Edge<*>? = null,
+    var aID: String? = null,
+    assumptionScope: Node? = node,
+    var status: AssumptionStatus = AssumptionStatus.Ignored,
+) : OverlayNode() {
+
+    @Relationship(value = "ASSUMPTION_SCOPE", direction = Relationship.Direction.OUTGOING)
+    /**
+     * The Scope in which this assumption has validity. It is equal to [underlyingNode] or can be
```
oxisto, line 53: Scope -> scope

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,144 @@
...
+        if (assumptionScope == null) {
+            this@Assumption.assumptionScope = super.underlyingNode
+        }
+        name = Name(assumptionType.name)
+        location = node?.location
+    }
+}
+
+enum class AssumptionStatus {
+    Undecided, // Default state: Patterns(Paths) depending on the assumption are returned, result
```
oxisto, line 69: Please use proper kdoc to annotate the individual items, not line comments.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,144 @@
...
+    // has an undecided state, assumptions are shown.
+    Rejected, // User or Algorithm: Patterns(Paths) depending on the assumption are not returned,
+    // results have a decided state, assumptions are (not?) shown.
+    Confirmed, // User or Algorithm: Patterns(Paths) depending on the assumption are returned,
```
oxisto, line 73: Maybe Accepted instead of Confirmed?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,144 @@
...
+    Assumption(
+        assumptionType,
+        message,
+        getCurrentFileAndLine(),
+        edge = edge,
+        assumptionScope = scope,
+    )
+}
+
+fun getCurrentFileAndLine(): String {
```
oxisto, line 141: missing doc. Why `stackTrace[4]`?! Will this crash if the stack trace is different to your expectations?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluator.kt
```diff
@@ -426,6 +426,7 @@ open class ValueEvaluator(
         var condition = expr.condition
 
         // Assume that condition is a binary operator
+        // Todo assume
```
oxisto, line 429: rather than writing `// TODO assume´ I would appreciate if we would actually already use `assume` here and set the message to "TODO", but this way we can at least search for all assume functions

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,144 @@
...
+ * @param assumptionType The type of assumption used to differentiate between assumptions and group
+ *   similar assumptions.
+ * @param edge The edge that cause the assumption to be necessary, even if the assumption has
+ *   validity for the entire program, the node that the edge starts at serves as location to be
+ *   reported to the user.
+ * @param scope The scope that the assumption has validity for, here the scope is a node, because
+ *   the assumption is valid for every node in its ast subtree.
+ * @param message The message describing the assumption that was taken.
+ */
+fun assume(assumptionType: AssumptionType, message: String, edge: Edge<*>?, scope: Node? = null) {
```
oxisto, line 130: As an idea you could use a common base class of `Node` and `Edge` to avoid a lot of duplicate code.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -85,6 +88,24 @@ inline fun <reified T> Node?.allChildrenWithOverlays(
     }
 }
 
+/**
+ * Searches for all Overlay nodes that are connected to a node in this nodes sub AST and returns all
+ * nodes of type [T]. For convenience, an optional predicate function [predicate] can be supplied,
+ * which will be applied via [Collection.filter].
+ */
+@JvmOverloads
+inline fun <reified T> Node?.allOverlays(noinline predicate: ((T) -> Boolean)? = null): List<T> {
```
oxisto, line 97: There is already `allChildrenWithOverlays` which I think does essentially the same thing.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -229,6 +250,36 @@ inline fun <reified T : Statement> FunctionDeclaration.body(n: Int = 0): T {
...
+    /**
+     * @param assumptionType The type of assumption used to differentiate between assumptions and group
+     *   similar assumptions.
+     * @param aID an ID chosen by the caller to identify the assumption across translation, e.g.
+     *   filename or classname, with a function name and a potential counter.
+     * @param scope The scope that the assumption has validity for, here the scope is a node, because
+     *   the assumption is valid for every node in its ast subtree.
+     * @param message The message describing the assumption that was taken.
+     */
+    fun <T> NodePath.assume(
```
oxisto, line 263: My gut feeling says that you want to make this an extension function of `NodePath`  and move it to the other assume functions so that they are all in the same place.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -229,6 +250,36 @@ inline fun <reified T : Statement> FunctionDeclaration.body(n: Int = 0): T {
 
 class StatementNotFound : Exception()
 
+class NodePath(var nodes: List<Node>, var assumptions: MutableList<Assumption>) {
```
oxisto, line 253: could be a data class, also documentation missing

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
```diff
@@ -336,6 +345,34 @@ fun QueryTree<*>.successfulLastNodes(): List<Node> {
...
+/**
+ * @param assumptionType The type of assumption used to differentiate between assumptions and group
+ *   similar assumptions.
+ * @param aID an ID chosen by the caller to identify the assumption across translation, e.g.
+ *   filename or classname, with a function name and a potential counter.
+ * @param scope The scope that the assumption has validity for, here the scope is a node, because
+ *   the assumption is valid for every node in its ast subtree.
+ * @param message The message describing the assumption that was taken.
+ */
+fun <T> QueryTree<T>.assume(
```
oxisto, line 357: Please move all of the assume functions into one file

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/memory/cxx/CXXDynamicLoadingPass.kt
```diff
@@ -97,10 +97,15 @@ class CXXDynamicLoadingPass(ctx: TranslationContext) : ConceptPass(ctx) {
         val path =
             call.arguments.getOrNull(0)?.followPrevDFG {
                 it is CallExpression && it.operationNodes.any { it is LoadLibrary }
-            }
+            } // Todo Propagate assumptions
```
oxisto, line 100: I have no idea what I need to do here? Do I, as a someone who writes a pass, need to propagate the assumptions? That would not be something that I expect

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/AssignExpression.kt
```diff
@@ -161,6 +161,7 @@ class AssignExpression :
             // We need to see if there is enough room on the left side. Currently, we only support
             // languages that do not allow to mix tuple and non-tuple types luckily, so we can just
             // assume that all arguments on the left side are assignment targets
+            // Todo assume
```
oxisto, line 164: replace with assume("TODO) call

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
```diff
@@ -306,6 +308,11 @@ class DFGPass(ctx: TranslationContext) : ComponentPass(ctx) {
                 node.variable.variables.lastOrNull()?.prevDFGEdges += iterable
             }
         }
+        assume(
+            AssumptionType.AmbiguityAssumption,
+            "If this is not the case, we assume that the last VariableDeclaration in the statement is the one we care about.",
```
oxisto, line 313: Not sure if this is a good example. What's an ambiguity assumption? The ambiguity is there, that's not an assumption. The assumption would be that because in 90 % of the cases we only have one variable here, we assume that only one variable is enough to process, or something like that.

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/GoEvaluationOrderGraphPass.kt
```diff
@@ -114,7 +114,8 @@ class GoEvaluationOrderGraphPass(ctx: TranslationContext) : EvaluationOrderGraph
                 // It is a bit philosophical whether the deferred call happens before or after the
                 // return statement in the EOG. For now, it is easier to have it as the last node
                 // AFTER the return statement
-                addEOGEdge(path.last(), defer.input)
+                // Todo Propagate AssumptionsEdge
```
oxisto, line 117: I as the author of this pass have no idea what to do here

oxisto (commented):
I would REALLY prefer to only have the 3-4 `assume` functions in a separate PR, as this already introduces too much logic into the system, some of which we need to probably have some discussion about. But let's see how/if we can resolve the comments.

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/GoEvaluationOrderGraphPass.kt
konradweiss: This is a note to me, to not forget the propagation of assumptions. I will remove all todos of this type and note this places down on my machine.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
konradweiss: That should not work if want to have this type of syntax, in that case i would not be allowed to use querytree.assume(...) but would have to write it down like this every time: `queryTree.also {it.assumptions.add(assume(...))}`. This is because the core module wants to have assumptions, when they are used in passes, but the core module has no dependency to cpg-analysis.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
konradweiss: I tried, before, but it did not allow me to use the @Relationship annotation there. In that case we would not persist the edge to neo4j, correct?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
konradweiss: Forgot to change the documentation when i changed the function.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
oxisto: What I would recommend is to have a common interface like "Assumable" (bad wording) and have the extension function based on that. You can then even have Node and Edge derive from this interface and reduce the set of functions from 4 to possibly 1 or 2.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
oxisto: Not sure what exactly you tried, but this seems to work fine for me 🤷 
```diff
diff --git a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
index 180d2c288ce..e18c4e002a4 100644
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
@@ -44,16 +44,16 @@ class Assumption(
     node: Node? = null,
     @DoNotPersist var edge: Edge<*>? = null,
     var aID: String? = null,
-    assumptionScope: Node? = node,
-    var status: AssumptionStatus = AssumptionStatus.Ignored,
-) : OverlayNode() {
-
-    @Relationship(value = "ASSUMPTION_SCOPE", direction = Relationship.Direction.OUTGOING)
     /**
      * The Scope in which this assumption has validity. It is equal to [underlyingNode] or can be
      * connected to its ancestor nodes in the AST if the scope is wider.
      */
-    var assumptionScope = assumptionScope
+    @Relationship(
+        value = "ASSUMPTION_SCOPE",
+        direction = Relationship.Direction.OUTGOING
+    ) var assumptionScope: Node? = node,
+    var status: AssumptionStatus = AssumptionStatus.Ignored,
+) : OverlayNode() {
 
     init {
         super.underlyingNode = node ?: edge?.start
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
konradweiss: An assumptionID that has to be provided when the assumption is not on a node or edge to identify the assumption across cpg runs. I changed the name and added more documentation there.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
konradweiss: Also works for me, I have no idea why it showed me an error message telling me that that annotation cannot be used on parameters. Maybe I did not mark it field by adding `val`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
konradweiss: Renamed

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
konradweiss: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
konradweiss: It was in the past, before i added the assume function. Now, with the interface i may be able to remove it again and make it a dataclass.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt
konradweiss: The types of assumptions we have flexible and can be adapted to your needs, and what seems adequate naming in the moment.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/memory/cxx/CXXDynamicLoadingPass.kt
konradweiss: Removed todo comment for now

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/expressions/AssignExpression.kt
konradweiss: Removed todo comment for now

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
konradweiss: I made an interface `HasAssumptions` out of it with only one assume function

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
konradweiss: This should never crash, especially after i made it private. To make it more transparent how this work, and not work off of a stack trace offset that should never change. I now look at the file names to be more resilient.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
konradweiss: After the introduction of the `HasAssumptions` interface, that is the base class we use

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
konradweiss: I removed NodePath again. If we properly use List<Edge> instead of List<Node> the information is already reachable over the edge.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
konradweiss: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
konradweiss: Removed this function

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,261 @@
...
+    NetworkAvailableAssumption,
+    ResourceExistsAssumption,
+    ServiceReachableAssumption,
+    AtomicExecutionAssumption,
+    TrustBoundaryAssumption,
+    DataRangeAssumption,
+    TrustedInputAssumption,
+}
+
+interface HasAssumptions {
```
oxisto, line 157: interface needs doc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,261 @@
...
+     *   name and a potential counter. The assumptionID is needed if the assumption is neither on a
+     *   node, nor on an edge.
+     * @param scope The scope that the assumption has validity for, here the scope is a node,
+     *   because the assumption is valid for every node in its ast subtree.
+     * @param message The message describing the assumption that was taken.
+     */
+    fun HasAssumptions.assume(
+        assumptionType: AssumptionType,
+        message: String,
+        assumptionID: String? = null,
```
oxisto, line 178: as disussed, we want to make sure the the node ID is deterministic in the future, so we should use it instead

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,261 @@
...
+        if (assumptionScope == null) {
+            this@Assumption.assumptionScope = super.underlyingNode
+        }
+        name = Name(assumptionType.name)
+        location = node?.location
+    }
+}
+
+/**
+ * the [AssumptionStatus] that an assumption can have, default being set to
```
oxisto, line 99: Spelling -> T

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -0,0 +1,261 @@
...
+    ResourceExistsAssumption,
+    ServiceReachableAssumption,
+    AtomicExecutionAssumption,
+    TrustBoundaryAssumption,
+    DataRangeAssumption,
+    TrustedInputAssumption,
+}
+
+interface HasAssumptions {
+    val assumptionNodes: MutableList<Assumption>
```
oxisto, line 158: as discussed offline, rename to `assumptions`, we will rename/remove/convert the assumptions in the security goal

oxisto (approved):
Minor things, can be merged once the discussed things are implemented

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/config/python/PythonStdLibConfigurationPass.kt
```diff
@@ -94,7 +94,7 @@ class PythonStdLibConfigurationPass(ctx: TranslationContext) : ConceptPass(ctx)
                 ?.mapNotNull { it.lastOrNull() as? Configuration }
                 ?.toSet()
                 ?.forEach { conf ->
-                    newLoadConfiguration(
+                    newLoadConfiguration( // Todo Propagate Assumptions to creation of new node
```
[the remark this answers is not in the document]
konradweiss: Removed this one too, it was an oversight from my side

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
oxisto: Rename the `assumptions` of Security Goal to `legacyAssumptions`. We will then later adjust the YAML files.

PR #2256, opened 2025-04-24
Added more tests - especially in console

oxisto:


codyze-console/src/test/kotlin/de/fraunhofer/aisec/codyze/console/ApplicationTest.kt
```diff
@@ -31,40 +31,52 @@ import de.fraunhofer.aisec.cpg.TranslationConfiguration
...
+import kotlin.test.*
 
 /** A mock configuration for the translation manager. */
-val mockConfig = TranslationConfiguration.builder().sourceLocations(File("some/path")).build()
+val mockConfig =
+    TranslationConfiguration.builder().sourceLocations(java.io.File("some/path")).build()
+
+/** A mock translation unit. */
+val mockTu =
+    TranslationUnitDeclaration().apply {
```
KuechA, line 59: Is there a specific reason why you neither use the `NodeBuilder` nor the Fluent-DSL?

KuechA (approved):

codyze-console/src/test/kotlin/de/fraunhofer/aisec/codyze/console/ApplicationTest.kt
oxisto: Sort of. I don't really have a translation context at this stage (and the service does not need one, only the finished nodes), so its hard to invoke them at this point.

PR #2261, opened 2025-04-30
Assumptions on splitting up the DFG which has to follow different copies of an object

KuechA:
The extension function `Node.generatesNewData()` can be used to identify different "copies" of an object which have to be handled separately. To do so, it applies some heuristics which affect the result of an analysis/query. This also means that we have to label each node on the path/result with an assumption. Validating this assumption can be done in two ways:
1. Checking that the result for each node on the path is correct.
2. Globally checking that the heuristics are correct and adequately model the program behavior.

Currently, we label each node with a new instance of the assumption and the text contains the description of option 1 (including start and end nodes) and the description of the heuristics from 2. This is likely to generate many assumptions, however, it should be easier to "approve" them without the risk of missing a result with an incorrect heuristics/assumption. If we have a better approach for such issues, I'd be interested in how this would be implemented @konradweiss. Also, if there's a chance that we visit some nodes multiple times (e.g. when evaluating different queries). Will we add similar/the same assumptions multiple times or is this already handled appropriately by the `assume` function?


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -264,43 +265,62 @@ fun dataFlowWithValidator(
...
+                this in (this.astParent as AssignExpression).rhs &&
+                !this.type.isMutable -> {
+                (this.astParent as AssignExpression).findTargets(this)
+            }
+            else -> emptySet()
         }
-        else -> emptySet()
-    }
+
+    this.assume(
```
oxisto, line 310: Maybe the master of assumptions (@konradweiss) needs to have a word here too, but I would assume that we would split up this list that you have and only put the matching assumptions in the "when" branches? If that is even possible?

KuechA: Sounds plausible. However, for the `else` branch, I guess we need this whole text as we would have to validate that the current node did not miss any "DFG-splitting" in the behavior? +1 for a check by @konradweiss that we exploit all capabilities of the API and what is the anticipated usage

konradweiss (approved):
With the introduction of the `NodeWithAssumption` and `NodeCollectionWithAssumption` this PR LGTM. We will later adapt the two data classes or the `NodePath` to be consistent with each other

PR #2267, opened 2025-05-02
Use `ConcurrentHashMap` to mark dirty and clean nodes

KuechA:
I somehow experienced `ConcurrentModificationException`s in `MermaidPrinterTest.testPrintEOG()` which seem to come from the newly introduced dirty nodes system. This is somehow weird and I would question why this can happen in the first place. Any idea @oxisto ? Can you replicate this problem? To fix it (if we agree that it actually exists), I'd use a `ConcurrentHashMap` instead of a normal `HashMap` when marking dirty/clean nodes. I'm wondering if we also have to synchronize access to the list of passes for a dirty node


oxisto (approved):
I *think* synchronised is not necessary if we have a concurrent hash map. It would have been necessary without it.

PR #2270, opened 2025-05-05
Use NodePath to return a path

konradweiss:
In the past, a `List<Node>` was used to return paths. This has the issue that edge information is lost, and assumptions on the path could not be propagated adequately. This PR introduced the `NodePath` object that contains a `List<Node>` and a `List<Assumption>`. `NodePath` and `Context` now are `HasAssumptions` to carry assumptions. The initial idea to use `List<Edge<*>>` as a better representation of a path had to be abandoned as some paths are found with a transition function on nodes and not explicit edge traversal that has one clear `Edge` element to follow.
The changes in this PR can be summarized by:

- Introducing `NodePath` to handle path information
- Add `NodePath` and `Context` to the implementations for `HasAssumptions`
- Adapt handling of return values where needed
- Forward assumptions with `addAssumptionDependences` where assumptions would be lost due to the conditional creation of new nodes.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -237,13 +239,21 @@ enum class FailureReason {
     HIT_EARLY_TERMINATION,
 }
 
+/**
+ * This data class represents a path of nodes as some paths are not just lists of edges, but instead
+ * can be traversed by a predicate.
+ */
+data class NodePath(val nodes: List<Node>) : HasAssumptions {
+    override val assumptions: MutableList<Assumption> = mutableListOf()
```
KuechA, line 247: Can you move this directly to the constructor?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -781,20 +794,29 @@ fun Node.followXUntilHit(
         if (nextNodes.isEmpty() && collectFailedPaths) {
             // TODO: How to determine if this path is really at the end or if it exceeded the number
             // of steps?
-            failedPaths.add(FailureReason.PATH_ENDED to currentPath.map { it.first })
+            failedPaths.add(
+                FailureReason.PATH_ENDED to
+                    NodePath(currentPath.map { it.first })
+                        .addAssumptionDependences(currentPath.map { it.second }.toList())
```
KuechA, line 800: Why this unintuitive interface of generating a `NodePath` and then calling `addAssumptionDependences` instead of directly adding them to the constructor?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -363,7 +373,10 @@ class Context(
     val indexStack: SimpleStack<IndexedDataflowGranularity> = SimpleStack(),
     val callStack: SimpleStack<CallExpression> = SimpleStack(),
     var steps: Int,
-) {
+) : HasAssumptions {
+
+    override val assumptions: MutableList<Assumption> = mutableListOf()
```
KuechA, line 378: The context is mostly used by the sensitivity classes to check if a step is valid. Can they also modify the assumptions? E.g. if an invokes edge has an assumption? Is there a system to "strip" assumptions again (e.g. if we enter a function and then exit it again, the subsequent nodes may not need that assumption any more if there was no significant impact by the function.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/AnalysisConfiguration.kt
```diff
@@ -285,7 +288,9 @@ class Forward(graphToFollow: GraphToFollow) : AnalysisDirection(graphToFollow) {
                         scope = scope,
                         sensitivities = sensitivities,
                     )
-                    .map { (edge, newCtx) -> this.unwrapNextStepFromEdge(edge) to newCtx }
+                    .map { (edge, newCtx) ->
+                        this.unwrapNextStepFromEdge(edge) to newCtx.addAssumptionDependence(edge)
```
KuechA, line 292: as this code repeats and is quite verbose, can we somehow simplify that or move it to a common function?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: Why do we only add `it.second` as assumption dependences? Would it make sense to refactor the worklist instead and directly collect `NodePath` objects there rather than adding assumptions to a buch of different objects and copying them over and over again?

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -430,12 +432,12 @@ fun Node.alwaysFlowsTo(
...
                         children = mutableListOf(QueryTree(value = it)),
                         stringRepresentation =
-                            "The EOG path reached the node ${it.lastOrNull()} matching the required predicate" +
+                            "The EOG path reached the node ${it.nodes.lastOrNull()} matching the required predicate" +
                                 if (earlyTermination != null)
                                     " before reaching a node matching the early termination predicate"
                                 else "",
                         node = this,
-                        terminationReason = Success(it.last()),
+                        terminationReason = Success(it.nodes.last()),
```
KuechA, line 440: Shouldn't we add the assumptions to the QueryTree? That should be quite easy

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -55,26 +55,26 @@ fun FulfilledAndFailedPaths.toQueryTree(
         SinglePathResult(
```
KuechA, line 55: Can we directly extend this `QueryTree` implementation to add the assumptions? Can we somehow unify it with the `NodePath` class?

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Sarif.kt
```diff
@@ -104,6 +102,8 @@ fun QueryTree<*>.getCodeflow(): List<List<Node>> {
         this.children.flatMap { it.getCodeflow() }
     } else if (this.value is List<*>) {
         listOf((this.value as Iterable<*>).filterIsInstance<Node>())
+    } else if (this.value is NodePath) {
```
KuechA, line 105: Can/should this happen?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
konradweiss: Done

KuechA: Wait. As you didn't want to use this in the other PR, it doesn't make sense to introduce it here either

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Sarif.kt
konradweiss: Yes and it did, in the past here some results were returned as List<Node> and handled properly, now, in case a NodePath is returned we have to unwrap.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/AnalysisConfiguration.kt
konradweiss: I implemented an unwrapper function for this.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
konradweiss: No there is no such functionality, if a transition function declares an assumption on the context, all paths containing that node due to the transition, the assumption is relevant for the path.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
konradweiss: Yes, I added it to the QueryTree in the `SinglePathResult` normally I would have expected the NodePath be a subgraph and therefore child of the QueryTree, but it is part of the value apparently

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
konradweiss: I think that would need some larger rework where everywhere we are currently returning NodePath, we return SinglePathResult, or we make NodePath itself a QueryTree and swap out `SinglePathResults` for it. But I would leave that for the future.

codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Sarif.kt
konradweiss: This may be the result of me accidentally adding a NodePath to the value of a QueryTree where before we had a List<Node>

KuechA (approved):

PR #2271, opened 2025-05-05
Tagging API to add concept and operations programmatically

oxisto:
This PR adds a "tagging API" which can be used to assign concepts and operations programmatically through the use of a `ConceptTagPass`. The desired tagging can either be directly assigned to the pass or the more conformable way is to use a Kotlin script with the file ending `.concept.kts` to do so. The latter is also useful if custom concepts want to be added.

### DIFF
--- a/cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/Extensions.kt
+++ b/cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/Extensions.kt
@@ -30,72 +30,9 @@ package de.fraunhofer.aisec.cpg.passes.concepts
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.OverlayNode
 import de.fraunhofer.aisec.cpg.graph.concepts.Concept
-import de.fraunhofer.aisec.cpg.graph.concepts.Operation
-import de.fraunhofer.aisec.cpg.graph.edges.flows.insertNodeAfterwardInEOGPath
-import de.fraunhofer.aisec.cpg.passes.Pass.Companion.log
-import kotlin.collections.set
 import kotlin.reflect.KClass
 
-class ConceptAssignmentContext {
-
-    internal val assignments: MutableList<OverlayAssignment<*>> = mutableListOf()
-
-    fun <T : Concept> ops(
-        assign: OverlayAssignment<T>,
-        block: OperationAssignmentContext<T>.(T) -> Unit,
-    ) {
-        assign.assigned.forEach { concept ->
-            val ctx = OperationAssignmentContext(concept)
-            block(ctx, concept)
-        }
-    }
-
-    internal fun build() {}
-}
-
-class OperationAssignmentContext<ConceptClass : Concept>(val concept: ConceptClass) {}
-
-/**  */
-class OverlayAssignment<T : OverlayNode>(val assigned: List<T>) {
-
-    companion object {
-        fun <T : OverlayNode> fromBuilder(
-            builder: () -> T,
-            nodes: List<Node>,
-        ): OverlayAssignment<T> {
-            return OverlayAssignment(
-                nodes.map { node ->
-                    // Create the overlay from the constructor
-                    val overlay = builder()
-
-                    log.debug(
-                        "Added overlay of type {} to {} '{}'",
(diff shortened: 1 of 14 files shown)


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/Extensions.kt
```diff
@@ -0,0 +1,180 @@
...
+ * The [TaggingContext] is also returned, which can then be used by the [TagOverlaysPass].
+ */
+fun tag(body: TaggingContext.() -> Unit): TaggingContext {
+    val ctx = TaggingContext()
+    body(ctx)
+    return ctx
+}
+
+/**
+ * each: Applies the specified overlay nodes to "each" of the nodes selected. A node is first of all
```
KuechA, line 64: Shouldn't this simply be the method name? If so, I wouldn't specify the name as string as refactoring will cause confusion

```suggestion
 * Applies the specified overlay nodes to "each" of the nodes selected. A node is first of all
```

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/Extensions.kt
```diff
@@ -0,0 +1,180 @@
...
+    val selector: Selector<T>,
+    val builder: (BuilderContext<T>) -> List<OverlayNode>,
+) {
+    fun collect(
+        lattice: NodeToOverlayState,
+        state: NodeToOverlayStateElement,
+        node: Node,
+    ): List<OverlayNode> {
+        // Check, if our selector matches
+        return if (selector(node as T)) {
```
KuechA, line 139: Can we get rid of this cast and move `T` to the signature without too much effort? Or should we actually make the cast with a check (`as?` and return the empty list if `null`?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/Extensions.kt
```diff
@@ -0,0 +1,180 @@
...
+) {
+    fun build(): List<OverlayNode> {
+        return builder(this)
+    }
+}
+
+/**
+ * A selector that describes a possible selection of a CPG [Node] by the following properties:
+ * - its [KClass] (mandatory, see [klass])
+ * - either its [Node.name] (see [namePredicate]) or any other property (see [predicate])
```
KuechA, line 166: Interesting choice to connect the `name` and `predicate` check with a logical OR and not an AND (which I would somehow have used/expected). Since it's documented correctly, it's ok.

KuechA (approved):
based on my comments, I think it's overall ok and just some minor aspects that i wanted to clarify/share my thoughts

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/TagOverlaysPass.kt
```diff
@@ -0,0 +1,58 @@
...
+import de.fraunhofer.aisec.cpg.passes.configuration.DependsOn
+
+/**
+ * This pass can be used to "tag" [OverlayNode]s to a specific "underlying" [Node] using the
+ * [EOGConceptPass]. It basically uses the [handleNode] function to introduce the individual items
+ * of a [TaggingContext] into the state of the [EOGConceptPass].
+ */
+@DependsOn(SymbolResolver::class)
+@DependsOn(ControlFlowSensitiveDFGPass::class)
+open class TagOverlaysPass(ctx: TranslationContext) : EOGConceptPass(ctx) {
```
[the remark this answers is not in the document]
oxisto: It's basically open because I extend it in `ConceptScriptPass`. I wanted to separate the tagging itself (which lives in cpg-concepts) from the more heavier "kotlin script" related stuff (which lives in codyze-core)

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/Extensions.kt
oxisto: Hmm good point. Might make more sense to combine with an AND to have something like `each<CallExpression>("foo") { it.bar }`

PR #2277, opened 2025-05-06
Update README.md with generated contributor images

oxisto:


KuechA (approved):
It's ok but I'm not sure if it's necessary as the same info is in Github's view on the right side of the screen

PR #2284, opened 2025-05-09
PrevPDG accesses prevDFG

KuechA:


konradweiss (approved):
Definitely the correct change here

PR #2285, opened 2025-05-09
Lattice Redesign fixes and performance

KuechA:
Two things happen in this PR:

* Return `Order.EQUAL` if two lattices are identical to save heavy computations
* Use `IdentitySet` and `List` when calculating key sets because anything else would lose equal but non-identical keys.


morbitzer (approved):
looks good

PR #2288, opened 2025-05-12
Make assumptions a set and add collectAssumptions 

konradweiss:
Sets are more adequate for holding assumptions to avoid storing the same assumptions multiple times at a node or in a result object.

The `collectAssumptions` function allows for retrieval of all contained assumptions in the final result when the assumption holding object is a composite type, e.g., QueryTree that holds QueryTree, and does not force us to rewrite setters to copy assumptions every time we add a `HasAssumptions` implementing object to another `HasAssumptions` implementing object. The use of `collectAssumptions` in `addAssumptionDependences` helps with the propagation of assumptions in composite `HasAssumptions` types, e.g., QueryTree, NodeWithAssumption.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
@@ -290,7 +290,7 @@ enum class AssumptionType {
  * [HasAssumptions.addAssumptionDependences] or [HasAssumptions.addAssumptionDependence].
  */
 interface HasAssumptions {
-    val assumptions: MutableList<Assumption>
+    val assumptions: MutableSet<Assumption>
 
     /**
      * This function adds a new assumption to the object it is called on. If the object is a node or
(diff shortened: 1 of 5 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -322,6 +322,15 @@ interface HasAssumptions {
         return this
     }
 
+    /**
+     * This is function is used to collect all assumptions stored in the [HasAssumptions] object and
+     * its contained objects that can have assumptions. While this default implementation is simple,
+     * composite [HasAssumptions] objects should return the assumptions of the contained objects.
+     */
+    fun collectAssumptions(): Set<Assumption> {
+        return assumptions.toSet()
```
KuechA, line 331: Do we need some loop detection in the implementations of this method? We seem to traverse some sub-graphs sometimes, which could result in problems, can't it? Currently, probably not a big deal though

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
konradweiss: That could in theory happen if for example a QueryTree contains itself as subtree, or a composite result contains itself as a result. However, that is only possible if a result is dependent on itself, which clearly a problem in the implementation, not a case that we have to handle during runtime. I would, therefore, not implement a loop detection or worklist. I could be convince do consider a sanity check in the future.

PR #2291, opened 2025-05-13
Let ScopedWalker.replace also consider BinaryOperators

morbitzer:
The current version of ScopedWalker.replace breaks the EOG if the node to replace is a BinaryOperator. This PR fixes this. 
Note: We may have to do something similar with other nodes


oxisto (approved):
I was thinking whether an `ArgumentHolder` could help there, but for some reason the argument holder interface only has function to interact with the arguments, but no iteration over all arguments, which we would use here to disconnect the EOG of the arguments :( So need to keep this this specific way for now.

morbitzer:
I introduced an additional function into the ArgumentHolder interface to make sure this works correctly over all nodes that may need this extra functionality. However, I'm a bit lost with all the classes that now need to introduce this new function, can someone support me here?

PR #2293, opened 2025-05-15
add List.toQueryTree

morbitzer:


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
```diff
@@ -369,3 +369,20 @@ class SinglePathResult(
...
 
 class QueryException(override val message: String) : Exception(message)
+
+/** Merges a List<QueryTree<Boolean>> into a single QueryTree<Boolean> */
+fun List<QueryTree<Boolean>>.toQueryTree(
+    node: Node? = null,
+    stringRepresentation: String = "",
+    assumptions: MutableSet<Assumption> = mutableSetOf(),
+): QueryTree<Boolean> {
+    val value = this.all { it.value }
```
KuechA, line 379: I'd suggest the following changes:

* Provide a version for `all` and `any`
* Collect the assumptions from the `QueryTree`s in `this`
* Maybe also generate the `stringRepresentation` directly in this method

KuechA (approved):

PR #2295, opened 2025-05-20
Test and update the function `dataFlowWithValidator`

### LINKED ISSUE #2292: `dataFlowWithValidator` may not behave as expected
I'd expect that the query function `dataFlowWithValidator` behaves as follows:

* Any path from a source to a sink requires passing through a validator.
* If no sink is reached on a path, we do not need to pass through a validator.

I think that the current implementation fails to handle the second case correctly and would return `false` but I have to double-check this.

KuechA:
Fixes #2292

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
@@ -246,7 +246,7 @@ fun dataFlowWithValidator(
     scope: AnalysisScope,
     vararg sensitivities: AnalysisSensitivity,
 ): QueryTree<Boolean> {
-    return source.alwaysFlowsToPrivate(
+    return source.alwaysFlowsToInternal(
         allowOverwritingValue = true,
         noSinkIsGood = true,
         earlyTermination = sinkPredicate,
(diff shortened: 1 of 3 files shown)


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -494,7 +532,16 @@ fun Node.alwaysFlowsTo(
                         terminationReason = Success(it.nodes.last()),
                     )
                 }
-        nothingFailed = nothingFailed && nextEOGEvaluation.failed.isEmpty()
+        nothingFailed =
+            nothingFailed &&
+                nextEOGEvaluation.failed.none {
```
konradweiss, line 537: can we extract the negation making it .all (noSinkIsGood && it.first == FailureReason.PATH_ENDED). That would IMO make it a lot more readable and AFAIK it should evaluate to the same expression

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -406,6 +409,41 @@ fun Node.alwaysFlowsTo(
...
+ * [allowOverwritingValue] can be used to configure if overwriting the value (or part of it) results
+ * in a failure of the requirement (if `false`) or if it does not affect the evaluation. The
+ * analysis can be configured with [scope] and [sensitivities]. [stopIfImpossible] enables the
+ * option to stop iterating through the EOG if we already left the function where we started and
+ * none of the nodes reaching by the DFG is in the new scope or one of its parents (i.e., the
+ * condition cannot be fulfilled anymore).
+ *
+ * If [noSinkIsGood] is set to `true`, all results with "ended path" are considered as fulfilled.
+ */
+private fun Node.alwaysFlowsToPrivate(
```
konradweiss, line 437: IMO we should name this Internal and not Private because I think to remember that we did so in other places and that would be more consistent. But the author is allowed to merge if he does not agree.

konradweiss (approved):
Two minor adjustments that could be made for readability and consistency

PR #2307, opened 2025-05-23
Fixes slow CPG performance

### LINKED ISSUE #2306: CPG translation is extremely slow
I am comparing an analysis of the cinder project. In v10.0.0 the whole analysis takes about 60s. In current main I aborted after 3+ min, one of the early passes (ambiguity resolver) took 40s alone vs. 1-2s in v10.0.0 :(

oxisto:
Introduced by #2250, the problem was a hashcode implementation for unwrapped edge collections
Fixes #2306


KuechA (approved):
Quite weird but should most likely not harm anything important

PR #2308, opened 2025-05-23
Fixed assumption debug spam

oxisto:
Also we will now hard-crash if we violate the assumption's assumption about having either a node or edge

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
@@ -25,6 +25,7 @@
  */
 package de.fraunhofer.aisec.cpg.assumptions
 
+import de.fraunhofer.aisec.cpg.frontends.TranslationException
 import de.fraunhofer.aisec.cpg.graph.Name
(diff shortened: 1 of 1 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/assumptions/Assumption.kt
```diff
@@ -83,19 +84,9 @@ class Assumption(
...
-            )
-        }
-
-        @Suppress("SENSELESS_COMPARISON")
-        if (listOf(node != null, edge != null, id != null).filter { it }.size > 1) {
-            log.warn(
-                "An assumption should be created with only one of the following arguments/properties: id, node or edge. But multiple of those are provided"
+        if (listOf(node != null, edge != null).filter { it }.size > 1) {
+            throw TranslationException(
+                "An assumption should be created with only one of the following arguments/properties: node or edge. But multiple of those are provided"
```
KuechA, line 89:
```suggestion
                "An assumption must be created with only one of the following arguments/properties: node or edge. But multiple of those are provided"
```

KuechA (approved):
I think that the text is no longer accurate

PR #2310, opened 2025-05-23
Allow nested tagging logic

KuechA:
Initial suggestion to implement nested tagging logic.


cpg-concepts/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/TagOverlaysPassTest.kt
```diff
@@ -30,19 +30,38 @@ import de.fraunhofer.aisec.cpg.TranslationContext
...
+import de.fraunhofer.aisec.cpg.graph.statements.expressions.Reference
+import java.util.Objects
 import kotlin.test.Test
 import kotlin.test.assertIs
 import kotlin.test.assertNotNull
 import kotlin.test.assertSame
 
 class TagOverlaysPassTest {
 
+    class CryptoArgument(underlyingNode: Node? = null) : Concept(underlyingNode) {
```
oxisto, line 49: I wonder if we could somehow manage to make those classes data classes. Then we would not need to manually specify equals / hashcode. Might also help some of our users ;)

oxisto (approved):
Better!

PR #2311, opened 2025-05-23
Implement more operators returning QueryTrees

KuechA:


cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/query/QueryTest.kt
```diff
@@ -318,7 +318,7 @@ class QueryTest {
         val queryTreeResult3 =
             result.allExtended<CallExpression>(
                 { it.name.localName == "memcpy" },
-                { it.arguments[2].intValue!! ne const(11) },
+                { it.arguments[2].intValue!! ne 11 },
```
oxisto, line 321: Can we make the receiver of our infix nullable? Then we could probably get rid of these annoying !! as well

oxisto (approved):
Very nice!

PR #2313, opened 2025-05-23
New implementation of OverlayNode.equals

### LINKED ISSUE #2312: `OverlayNode.equals` is wrong
The implementation of `OverlayNode.equals` calls `Node.equals` which returns that the nodes are not equal if they have no `location` and are not identical. As `OverlayNode`s do not have a `location`, this cannot work. Detected when implementing #2310

KuechA:
Fixes #2312

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/OverlayNode.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/OverlayNode.kt
@@ -54,9 +54,21 @@ abstract class OverlayNode() : Node() {
      * have different underlying nodes.
      */
     override fun equals(other: Any?): Boolean {
+        if (this === other) {
+            return true
+        }
         return other is OverlayNode &&
-            super.equals(other) &&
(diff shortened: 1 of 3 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/OverlayNode.kt
```diff
@@ -54,9 +54,21 @@ abstract class OverlayNode() : Node() {
...
-            super.equals(other) &&
-            this.underlyingNode == other.underlyingNode
+            name == other.name &&
+            code == other.code &&
+            comment == other.comment &&
+            location == other.location &&
+            isImplicit == other.isImplicit &&
+            // If the underlying node is null, we don't care about it. We also cannot compare the
+            // location in this case.
+            (this.underlyingNode == null ||
```
oxisto, line 68: Hmm I am pretty sure I had proposed similar code in the past and I was widely criticized for that :D I thought we don't need this special handling of an empty underlying node.

Maybe I misunderstood what the original problem is?

PR #2319, opened 2025-05-26
Introduce 4 state result

### LINKED ISSUE #2318: Introduce 4-State "Boolean"
We need a 4-state "Boolean" instead of the regular boolean as the final result of a query:
- Proven, which means the statement is "proven" under the (accepted) assumptions
- Violated, which means the statement is "violated" under the (accepted) assumptions
- Undecided, which means the statement cannot be decided properly because there are unaccepted assumptions left
- NotYetEvaluated, which means that it has not been evaluated yet, mainly for manual assessment.

My suggestions would be a sealed class + data object.

We probably want to have a function that converts a QueryTree<Boolean> into this 4-state because we probably want to continue using true/false internally in the queries but use the 4-state in the fulfilment of the requirement.

KuechA:
Fixes #2318

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2025, Fraunhofer AISEC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *                    $$$$$$\  $$$$$$$\   $$$$$$\
+ *                   $$  __$$\ $$  __$$\ $$  __$$\
+ *                   $$ /  \__|$$ |  $$ |$$ /  \__|
+ *                   $$ |      $$$$$$$  |$$ |$$$$\
+ *                   $$ |      $$  ____/ $$ |\_$$ |
+ *                   $$ |  $$\ $$ |      $$ |  $$ |
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.query
+
+import de.fraunhofer.aisec.cpg.assumptions.AssumptionStatus
+
+sealed class RequirementState
+
+/**
+ * Represents a query that has been evaluated and the result is known to be `false` or some
+ * assumptions are rejected.
+ */
+data object Failed : RequirementState()
+
+/**
+ * Represents a query that has been evaluated and the result is known to be `true` and all
+ * assumptions are accepted.
+ */
+data object Succeeded : RequirementState()
+
+/**
+ * Represents a query that has been evaluated but the result is not yet known because some
+ * assumptions have to be accepted (or rejected).
+ */
+data object IsUndecided : RequirementState()
+
+/**
+ * Represents a query that has not yet been evaluated. Will be most likely used in the context of
+ * manual assessments which have to be conducted.
+ */
+data object IsNotYetEvaluated : RequirementState()
+
+fun QueryTree<Boolean>.toQueryState(): QueryTree<RequirementState> {
+    val (newValue, stringInfo) =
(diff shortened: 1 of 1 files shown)


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
```diff
@@ -0,0 +1,236 @@
...
+package de.fraunhofer.aisec.cpg.query
+
+import de.fraunhofer.aisec.cpg.assumptions.AssumptionStatus
+import de.fraunhofer.aisec.cpg.evaluation.compareTo
+
+sealed class RequirementState
+
+/**
+ * Represents a query that has been evaluated and the result is known to be `false` or some
+ * assumptions are rejected.
```
oxisto, line 35: Please link to the [assumptions] variable in this kdoc. Same for all.

oxisto: Ideally, we should also link to the Rejected enum / object in case we ever rename them to keep the wording consistent.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
```diff
@@ -0,0 +1,236 @@
...
+ * Represents a query that has been evaluated and the result is known to be `true` and all
+ * assumptions are accepted.
+ */
+data object Succeeded : RequirementState()
+
+/**
+ * Represents a query that has been evaluated but the result is not yet known because some
+ * assumptions have to be accepted (or rejected).
+ */
+data object IsUndecided : RequirementState()
```
oxisto, line 49: Why prefixed with `Is`?

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * Represents a query that has not yet been evaluated. Will be most likely used in the context of
+ * manual assessments which have to be conducted.
+ */
+data object IsNotYetEvaluated : RequirementState()
+
+fun QueryTree<Boolean>.toQueryState(): QueryTree<RequirementState> {
```
oxisto, line 57: Would it make sense to supply a type alias like `RequirementEvaluation` for `QueryTree<RequirementState>`?

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
```diff
@@ -0,0 +1,236 @@
...
+        value = newValue,
+        children = mutableListOf(this),
+        stringRepresentation = "The requirement ${ newValue::class.simpleName } because $stringInfo",
+    )
+}
+
+/**
+ * Performs a logical and (&&) operation between the values and creates and returns [QueryTree]s.
+ */
+infix fun RequirementState.and(other: RequirementState): QueryTree<RequirementState> {
```
oxisto, line 86: I am confused. Why are all of these functions below necessary? Shouldn't it behave just like any other query tree?

KuechA: Because these functions are only available for `QueryTree<Boolean>`, the rest don't have that. So, if you want some logic (which is going to be the case based on the anticipated use-case) , you need this

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
KuechA: removed

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
oxisto: They way I am imagining it is the following: user-facing we will try to work with QueryTree<Boolean> and it the logic-operators should only be there for Boolean. Then once we execute the queries inside codyze, we convert the Boolean result + assumptions into the QueryTree<RequirementState>. At this state, we will only have the (final) result. We will not do any more computations with it.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
```diff
@@ -0,0 +1,225 @@
...
+ */
+data object Undecided : DecisionState()
+
+/**
+ * Represents a query that has not yet been evaluated. Will be most likely used in the context of
+ * manual assessments which have to be conducted.
+ */
+data object NotYetEvaluated : DecisionState()
+
+fun QueryTree<Boolean>.decide(): Decision {
```
oxisto, line 61: Since this will a public facing API, please generate a short docstring for this.

oxisto (approved):

PR #2323, opened 2025-05-27
Cleanup state on finalCleanup & documentation

maximiliankaul:
The `intermediateState` object must be cleared once the pass is done or it will interfere with other passes.


oxisto:
> The `intermediateState` object must be cleared once the pass is done or it will interfere with other passes.

Don't you mean the `finalState`?

PR #2329, opened 2025-05-28
Disconnect OverlayNodes from the graph

maximiliankaul:


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -307,6 +307,20 @@ abstract class Node() :
...
         nextEOGEdges.clear()
         prevEOGEdges.clear()
+
+        if (this is OverlayNode) {
+            underlyingNode?.overlayEdges?.removeAll { edge ->
+                edge.end === this || edge.start === this
+            }
+            underlyingNodeEdge.clear()
+        }
+        this.overlays.forEach { overlay ->
```
oxisto, line 317: Also here, clearing overlay edges should be enough

oxisto:
I added a test - since there was none. After actually testing this with a test, I discovered that additional removal/clear functionality is needed in the singleton edge list.

PR #2330, opened 2025-05-28
Keep order when creating nodes EOGConceptPass

maximiliankaul:
Currently, when creating nodes in an EOGConceptPass they are first added to a set and then later created via some pass specific internal mechanisms. This means that the order when creating multiple nodes in the pass and the order when they are actually created internally might differ, resulting in unexpected EOG results for example.

- [ ] Do we need to change more sets to ordered sets?
- [ ] Do we want to have this change?

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
@@ -31,12 +31,7 @@ import de.fraunhofer.aisec.cpg.helpers.functional.PowersetLattice.Element
 import de.fraunhofer.aisec.cpg.helpers.identitySetOf
 import de.fraunhofer.aisec.cpg.helpers.toIdentitySet
 import java.io.Serializable
-import java.util.IdentityHashMap
-import kotlin.collections.component1
-import kotlin.collections.component2
-import kotlin.collections.fold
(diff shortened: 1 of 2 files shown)


KuechA:
Actually, I'm pretty sure that these changes do not have the intended effect. The PR changes the order in which a `Lattice` would keep its elements. However, based on the description of the PR, I think that the intention is to change the implementation of `PowersetLattice.Element` (which currently inherits from `IdentitySet`) to a `LinkedHashSet`. That's where we keep the `OverlayNodes` which are created by the `EOGConceptPass`es.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
```diff
@@ -208,7 +203,7 @@ interface Lattice<T : Lattice.Element> {
 
 /** Implements a [Lattice] whose elements are the powerset of a given set of values. */
 class PowersetLattice<T>() : Lattice<PowersetLattice.Element<T>> {
-    override lateinit var elements: Set<Element<T>>
+    override lateinit var elements: LinkedHashSet<Element<T>>
 
     class Element<T>(expectedMaxSize: Int) : IdentitySet<T>(expectedMaxSize), Lattice.Element {
```
KuechA, line 208: Probably you want to change this line of code. We need a very fast and scalable implementation of a `Set` which keeps the order and compares based on the identity, not equality.

PR #2333, opened 2025-05-30
Update AssumptionStatus and better assumption collection before decide

konradweiss:
This PR does the following:

- Uses the assumption statuses in the translation results to update the assumption status of a `QueryTree` before the decision process.
- Fixes a mistake, at least one IMO, in the decision logic.
- Collects all assumptions related to a `QueryTree`.
- Adds global assumptions that are attached to the `TranslationResult` and `Component` assumptions in the `collectAssumptions` of the Node class.

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
@@ -265,12 +265,11 @@ fun dataFlowWithValidator(
  * computing this node as the result of a function.
  */
 data class NodeWithAssumption(val node: Node) : HasAssumptions {
-    /** See [HasAssumptions.assumptions]. */
     override val assumptions: MutableSet<Assumption> = mutableSetOf()
 
     /**
      * Adds the [assumptions] of the current [NodeCollectionWithAssumption] and the assumptions of
-     * the node that is the result. See [HasAssumptions.collectAssumptions].
+     * the node that is the result.
      */
     override fun collectAssumptions(): Set<Assumption> {
         return super.collectAssumptions() + node.collectAssumptions()
@@ -283,12 +282,11 @@ data class NodeWithAssumption(val node: Node) : HasAssumptions {
  * taken when computing this collection of nodes as the result of a function.
  */
 data class NodeCollectionWithAssumption(val nodes: Collection<Node>) : HasAssumptions {
-    /** See [HasAssumptions.assumptions]. */
     override val assumptions: MutableSet<Assumption> = mutableSetOf()
 
     /**
      * Adds the [assumptions] of the current [NodeCollectionWithAssumption] and the assumptions of
-     * all nodes contained in the object. See [HasAssumptions.collectAssumptions].
+     * all nodes contained in the object.
      */
     override fun collectAssumptions(): Set<Assumption> {
         return super.collectAssumptions() + nodes.flatMap { it.collectAssumptions() }
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
@@ -265,8 +265,13 @@ fun dataFlowWithValidator(
  * computing this node as the result of a function.
  */
 data class NodeWithAssumption(val node: Node) : HasAssumptions {
+    /** See [HasAssumptions.assumptions]. */
     override val assumptions: MutableSet<Assumption> = mutableSetOf()
 
+    /**
+     * Adds the [assumptions] of the current [NodeCollectionWithAssumption] and the assumptions of
+     * the node that is the result. See [HasAssumptions.collectAssumptions].
+     */
     override fun collectAssumptions(): Set<Assumption> {
         return super.collectAssumptions() + node.collectAssumptions()
     }
@@ -278,8 +283,13 @@ data class NodeWithAssumption(val node: Node) : HasAssumptions {
  * taken when computing this collection of nodes as the result of a function.
  */
 data class NodeCollectionWithAssumption(val nodes: Collection<Node>) : HasAssumptions {
+    /** See [HasAssumptions.assumptions]. */
     override val assumptions: MutableSet<Assumption> = mutableSetOf()
 
+    /**
+     * Adds the [assumptions] of the current [NodeCollectionWithAssumption] and the assumptions of
+     * all nodes contained in the object. See [HasAssumptions.collectAssumptions].
+     */
     override fun collectAssumptions(): Set<Assumption> {
         return super.collectAssumptions() + nodes.flatMap { it.collectAssumptions() }
     }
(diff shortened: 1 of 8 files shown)


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
```diff
@@ -64,29 +64,43 @@ data object NotYetEvaluated : DecisionState()
...
- * [AssumptionStatus.Rejected] or [AssumptionStatus.Undecided]).
+ * [QueryTree.collectAssumptions] (i.e., it checks if all are [AssumptionStatus.Accepted] or if some
+ * are [AssumptionStatus.Rejected] or [AssumptionStatus.Undecided]).
  */
 context(TranslationResult)
 fun QueryTree<Boolean>.decide(): Decision {
     val statues = this@TranslationResult.assumptionStatuses
-    // TODO(kweiss): Use status to set status
+    // The assumptions need to be collected, as they are located at the respective construct they
+    // are placed on and only
```
KuechA, line 74: odd formatting

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
```diff
@@ -286,6 +286,10 @@ abstract class Node() :
         Overlays(this, mirrorProperty = OverlayNode::underlyingNodeEdge, outgoing = true)
     var overlays by unwrapping(Node::overlayEdges)
 
+    override fun collectAssumptions(): Set<Assumption> {
+        return super.collectAssumptions() + (component?.assumptions ?: emptySet())
```
KuechA, line 290: Should it be `assumptions` or `collectAssumptions`?

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
```diff
@@ -64,29 +64,43 @@ data object NotYetEvaluated : DecisionState()
...
+    // forwarded on evaluation. Accepting or rejecting an assumption has a different impact on query
+    // evaluation depending
+    // on the sup-query tree the assumption is placed on.
+    // Global assumptions are also included below, component wide assumptions are included with
+    // collectAssumptions() in
+    // the individual nodes.
+    val assumptions = this.collectAssumptions() + this@TranslationResult.collectAssumptions()
```
KuechA, line 81: Shouldn't the assumptions of the `TranslationResult` affect every query and already be included in the result of `this.collectAssumptions()`? I'm somewhat lost when to use which of these options (accessing `assumptions` or calling `collectAssumptions`) as it's not very well documented

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
konradweiss: Spotless. Apparently somethings one has to do manually.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
konradweiss: I had collectAssumptions before, which creates a loop. Here I knew that the component would be a simple node and `.assumptions` would be enough. Else, I would have to override the ´collectAssumptions´ specifically for `Component`s to not recursively call `collectAssumptions` because a components component is itself.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
konradweiss: Whether or not we include all Assumptions on a TranslationResult on every node in the graph is somewhat of a difficult question. It is obviously enough to add them at the end, when doing the final decision, because they will always be relevant. But philosophically we could add them at every node in the graph and then collect them when we call `collectAssumptions`, bot would result in the same assumptions at the end. The only reason why I did not do so with the `Components` is that nodes can be part of different components and I did not find a way to get all contained `Nodes` in a QueryTree.

There is sufficient documentation on HasAssumptions.collectAssumptions in the interface. I can add some more documentation on the overriding functions to clarify why they do shift away from that default behavior.

One should always use collectAssumptions when accessing all relevant assumptions for an object during final evaluation, the field `assumptions` only contains the assumptions attached to that specific option by either calling `assume()` or explicitly collect and store all assumptions of a different `HasAssumptions` object by calling `addAssumptionDependences()`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Node.kt
KuechA: I'd suggest to override the implementation for `Component` in this case because this system sounds very error-prone if the user just has to known this or will create loops or not receive any meaningful result.

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
```diff
@@ -70,10 +70,7 @@ open class QueryTree<T>(
      */
     open var node: Node? = null,
 
-    /**
-     * Assumptions can be created in the QueryTree object with the [assume] function ore by adding
-     * an assumption manually.
-     */
+    /** See [HasAssumptions.assumptions]. */
```
KuechA, line 73: Isn't there a link by default to the parent implementation's documentation?

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/FlowQueries.kt
```diff
@@ -265,8 +265,13 @@ fun dataFlowWithValidator(
  * computing this node as the result of a function.
  */
 data class NodeWithAssumption(val node: Node) : HasAssumptions {
+    /** See [HasAssumptions.assumptions]. */
```
KuechA, line 268: Isn't there a link by default to the parent implementation's documentation? -> can be left out

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -261,8 +261,13 @@ enum class FailureReason {
  */
 data class NodePath(
     val nodes: List<Node>,
+    /** See [HasAssumptions.assumptions]. */
```
KuechA, line 264: see above

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -414,6 +419,7 @@ class Context(
     var steps: Int,
 ) : HasAssumptions {
 
+    /** See [HasAssumptions.assumptions]. */
```
KuechA, line 422: see above

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryStates.kt
KuechA: ok. maybe I didn't read the documentation close enough then or it was not clear to me because of the `Component` loop thing which should be fixed now.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
konradweiss: If there is no previous doc on the function, there is a way to show the doc of the parent function, so I will remove this. In cases where the overriding field/method adds Doc, the parent Doc is not shown, so here there may be a benefit in having the reference. On the other hand one can use navigation shortcuts to traverse the parent implementations/declarations until finding the Documentation.

PR #2337, opened 2025-06-03
Minor fix in map lattice

KuechA:


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/EOGConceptPass.kt
```diff
@@ -72,10 +72,6 @@ open class EOGConceptPass(ctx: TranslationContext) :
     var currentComponent: Component? = null
 
     override fun cleanup() {
-        // Nothing to do
-    }
-
-    override fun finalCleanup() {
         val finalState = intermediateState ?: return
```
oxisto: Let's remove intermediate state altogether

PR #2341, opened 2025-06-04
Computes the Decision Lazily and for nested QueryTree<Boolean>

konradweiss:
This PR:

- Implements the decision function for nested `QueryTree<Boolean>`
- Tests the results 
- Fixes an Issue with propagation of `Undecided`

Note: The computation has to be done lazily by providing a lambda for each query tree that will later compute the decide() as the assumption status is set after the construction of `QueryTree<Boolean>`

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
@@ -107,7 +107,7 @@ open class QueryTree<T>(
                     else
                         "the assumptions ${assumptions.filter { it.status == AssumptionStatus.Rejected }.map { it.id.toHexDashString() }.joinToString(", ") } were rejected")
 
-            assumptions.any { it.status == AssumptionStatus.Undecided } ->
+            this == Undecided || assumptions.any { it.status == AssumptionStatus.Undecided } ->
                 Undecided to
                     "the assumptions ${assumptions.filter { it.status == AssumptionStatus.Undecided }.map { it.id.toHexDashString() }.joinToString(", ")} are not yet decided"
 
@@ -217,7 +217,7 @@ infix fun <T, S> T.IN(other: S): QueryTree<Boolean> {
             ?: throw IllegalArgumentException(
                 "Cannot check if ${thisQt.value} is of type ${otherQt.value}. The other value must be a Collection<*>."
             )
-    
+
     return QueryTree(result, mutableListOf(thisQt, otherQt), "${thisQt.value} in ${otherQt.value}")
 }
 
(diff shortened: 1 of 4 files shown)


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
```diff
@@ -527,17 +624,22 @@ fun List<QueryTree<Boolean>>.mergeWithAll(
...
+                    "All elements has value true"
+                } else {
+                    "At least one of the elements has false"
+                },
+            node = node,
+            assumptions = assumptions,
+        )
+        .registerLazyDecision {
+            this.drop(1).fold(this.first().lazyDecision.value) { acc, qt ->
+                acc and qt.lazyDecision.value
```
KuechA: Can you rename the variables? These lines of code are a bit hard to understand, some comments would be good

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
```diff
@@ -71,6 +72,72 @@ open class QueryTree<T>(
     open var node: Node? = null,
     override var assumptions: MutableSet<Assumption> = mutableSetOf(),
 ) : Comparable<QueryTree<T>>, HasAssumptions {
+
+    /**
+     * The purpose of [lazyDecision] is to evaluate the decision after all post-processing
+     * information is applied e.g. after setting the [AssumptionStatus] of the [Assumption]s. This
```
KuechA:
```suggestion
     * information is applied, e.g., after setting the [AssumptionStatus] of the [Assumption]s. This
```

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
```diff
@@ -71,6 +72,72 @@ open class QueryTree<T>(
...
+     * default implementation will simply consider the value of the [QueryTree], if the value is a
+     * [Boolean], it will return [Failed] and [Succeeded] respectively. If the value is not a
+     * [Boolean], it will return [Succeeded] for now as the value was simply determined.
+     *
+     * Functions on QueryTree<Boolean> that will restrict, whether all assumptions in the query tree
```
KuechA: This paragraph is hard to read. Maybe something like: This function should be overridden for [QueryTree]s and certain operations to steer how assumptions affect the decision-making process. As an example, for logic operations "and", "or", and "implies", the decision should first be made for the left and right hand side of the operation because invalidating a sub-tree has different effects for each operation.

KuechA (commented):
Some nits

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
konradweiss: renamed vars, simplified expression added some doc

KuechA (approved):

PR #2348, opened 2025-06-06
Improved loop detection

### LINKED ISSUE #2289: Potentially to sensitive loop detection
This piece of code is supposed to detect if a node was already visited over a call context and is used for loop detection:
```
/**
 * Checks if the [node] is already in the [path] and if the last element on the call stack reaching
 * it is the same in [context] and in the [path]. This serves as an indication of a loop. We need
 * this to differentiate that a node (e.g., function declaration) can be visited through multiple
 * paths, e.g., if there are subsequent calls to the same function without indicating a loop.
 */
fun isNodeWithCallStackInPath(
    node: Node,
    context: Context,
    path: Collection<Pair<Node, Context>>,
): Boolean {
    return path.any {
        it.first === node &&
            context.callStack.top?.let { top -> top in it.second.callStack } != false
    }
}
```
Given the following code I think that the function returns true although the node is visited the first time over that path:
```
1: main() {
2:     for i in range(1)
3:        a(i)
4: }
5: 
6: fun a(int value){
7:     if(value == 1){
8:         print("Here a loop is detected when coming from the call site in line 3")
9:     }else{
10:       a(1) 
11:    }
12:}

```
the loop should be erroneously detected when visiting 8 with callstack(3) because the entire path at this point contains a visit to 8 with callstack(3:10).

I think this may be solved by comparing tops, e.g. `context.callStack.top === it.second.callStack.top ` but I am not sure.

KuechA:
Fixes #2289

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
@@ -856,7 +856,7 @@ fun Node.followXUntilHit(
         }
 
         for ((next, newContext) in nextNodes) {
-            // Copy the path for each outgoing CDG edge and add the next node
+            // Copy the path for each outgoing edge and add the next node
             if (predicate(next)) {
                 // We ended up in the node fulfilling "predicate", so we're done for this path. Add
                 // the path to the results.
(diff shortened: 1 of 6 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -925,8 +925,8 @@ fun isNodeWithCallStackInPath(
     path: Collection<Pair<Node, Context>>,
 ): Boolean {
     return path.any {
-        it.first == node &&
-            context.callStack.top?.let { top -> top in it.second.callStack } != false
+        it.first == node && context.callStack == it.second.callStack
+        // context.callStack.top?.let { top -> top in it.second.callStack } != false
```
oxisto: Remove commented out code

PR #2358, opened 2025-06-12
Setting assumption status in `Assumption` creation

oxisto:
Via companion object


codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/Project.kt
```diff
@@ -146,16 +146,16 @@ class AnalysisProject(
 
     /** Analyzes the project and returns the result. */
     fun analyze(): AnalysisResult {
-        val tr = TranslationManager.builder().config(config).build().analyze().get()
-
         // Propagate assumption status into a translation result
         assumptionStatusFunctions.forEach { (uuid, func) ->
-            tr.assumptionStates[Uuid.parse(uuid)] = func(tr)
+            Assumption.statues[Uuid.parse(uuid)] = func()
```
KuechA: Is this name `statues` intended?

oxisto: No

PR #2366, opened 2025-06-18
More UI fixes

oxisto:
- **Loading children with assumptions 10 at a time**
- **Filtering out duplicates assumptions**
- **More loading fixes**


codyze-console/src/main/webapp/src/lib/utils/itemLimiting.ts
KuechA, line 1: Is this required?

KuechA (approved):

PR #2376, opened 2025-06-24
Upgrade SootUp to 2.0.0

KuechA:
We're currently stuck in SootUp version 1.3.0 due to some breaking API changes in subsequent versions. This PR upgrades to the most recent version 2.0.0 and also tries to enable analysis of apk files. However, the latter does not work properly yet and we fail with an OutOfMemoryException. This problem needs further investigation.

### DIFF
--- a/cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/JVMLanguageFrontend.kt
+++ b/cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/JVMLanguageFrontend.kt
@@ -35,18 +35,27 @@ import de.fraunhofer.aisec.cpg.graph.declarations.TranslationUnitDeclaration
 import de.fraunhofer.aisec.cpg.graph.types.Type
 import de.fraunhofer.aisec.cpg.sarif.PhysicalLocation
 import java.io.File
+import sootup.apk.frontend.ApkAnalysisInputLocation
+import sootup.apk.frontend.DexBodyInterceptors
 import sootup.core.model.Body
 import sootup.core.model.SootMethod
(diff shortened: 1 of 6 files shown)


cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/JVMLanguageFrontend.kt
```diff
@@ -173,8 +180,13 @@ class JVMLanguageFrontend(
     }
 
     override fun codeOf(astNode: Any): String? {
+
         if (astNode is SootMethod && astNode.isConcrete) {
-            return astNode.body.toString()
+            try {
+                return astNode.body.toString()
+            } catch (_: IllegalArgumentException) {
+                // Do nothing
```
konradweiss: I am confused, I would log something if we eat an exception, or can this never happen and catching it is enforced somehow?

konradweiss (commented):
One minor remark and I am wondering if we should investigate the OutOfMemoryException when analyzing an APK before merging this. Is it a less important side feature?

KuechA:
> One minor remark and I am wondering if we should investigate the OutOfMemoryException when analyzing an APK before merging this. Is it a less important side feature?

The feature would be relevant. However, I don't see how this PR benefits from debugging SootUp's code. It does not happen in our code and we do follow the documentation

cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/JVMLanguageFrontend.kt
KuechA: No idea. We log something now.

PR #2377, opened 2025-06-25
Faster fixpoint iteration

KuechA:
This PR aims to speed up the fixpoint iteration by improving the creation, comparison and computation of the least upper bound of the lattices. These are a bottleneck in many passes when dealing with more complex functions.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
@@ -139,16 +139,44 @@ interface Lattice<T : Lattice.Element> {
         transformation: (Lattice<T>, EvaluationOrder, T) -> T,
     ): T {
         val globalState = IdentityHashMap<EvaluationOrder, T>()
-        val finalState: T = this.bottom
+        var finalState: T = this.bottom
         for (startEdge in startEdges) {
             globalState[startEdge] = startState
         }
-        val edgesList = mutableListOf<EvaluationOrder>()
-        startEdges.forEach { edgesList.add(it) }
+        // This list contains the edge(s) (probably only one unless we made a mistake) of the
+        // current basic block that we are currently processing. We select this one with priority
+        // over the other options.
+        val currentBBEdgesList = mutableListOf<EvaluationOrder>()
+        // This list contains the edge(s) that are the next branch(es) to process. We process these
+        // after the current basic block has been processed.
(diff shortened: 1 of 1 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/IdentitySet.kt
```diff
@@ -80,6 +80,30 @@ open class IdentitySet<T>(expectedMaxSize: Int = 16) : MutableSet<T> {
...
+        for (element in elements) {
+            map[element] = counter.addAndGet(1)
+        }
+    }
+
+    /**
+     * Adds all [elements] to this [IdentitySet] without checking if they are already present. This
+     * should only be used if this set is empty!
+     */
+    fun addAllWithoutCheck(elements: IdentitySet<T>) {
```
konradweiss, line 87: Do we need both functions here? They are both sets so I would have hoped that the above is usable with all instances where this function is used.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
```diff
@@ -209,13 +210,13 @@ class PowersetLattice<T>() : Lattice<PowersetLattice.Element<T>> {
     override lateinit var elements: Set<Element<T>>
 
     class Element<T>(expectedMaxSize: Int) : IdentitySet<T>(expectedMaxSize), Lattice.Element {
-        constructor(set: Set<T>) : this(set.size) {
-            addAll(set)
+        constructor(set: Set<T>) : this(ceil(set.size * 1.5).toInt()) {
```
konradweiss, line 266: I would like a small comment here why we make this set bigger. I assume it is so that future additions are faster?

konradweiss:
I ran our current main and this branch against barbican several times and did not notice any noticeable speedup. However, if the changes are overall good, I would still merge them even without doing more benchmarking.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/IdentitySet.kt
KuechA: Removed the upper one

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
KuechA: correct. added a comment

konradweiss (approved):

PR #2378, opened 2025-06-25
Faster computation of `hashCode` of nodes.

KuechA:
We compute the `hashCode` of nodes in many occasions. Several nodes may hold a lot of sub-statements which makes the computation expensive if it is based on all of the sub-statements. Having a faster computation thereof can be beneficial to improve the performance of passes. The biggest effect is probably introduced by removing the `statements` from a `Block`'s `hashCode` since this can easily add up to thousands of nodes whose hashCode has to be created then.


oxisto:
> Someone should double-check if this has no unintended side-effects.

Do we have any numbers on the performance improvement this brings - since we are not sure about the side-effects?

konradweiss:
I don't see this change producing a problematic side effect unless we are using hashCode for something it is not supposed to be used. The only negative impact would be performance itself if we get to many hashCode collisions that the linear search in the hash bucket takes to long. Or am I missing something?

I tried doing small performance tests but they are not conclusive IMO, here we clearly would need something more sophisticated.

konradweiss (approved):

PR #2379, opened 2025-06-27
Construct CDG based on basic blocks

KuechA:


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
```diff
@@ -64,6 +70,169 @@ open class ControlDependenceGraphPass(ctx: TranslationContext) : EOGStarterPass(
...
+                this.startNode == other.startNode &&
+                this.endNode == other.endNode
+        }
+
+        override fun toString(): String {
+            return "$startNode - $endNode"
+        }
+    }
+
+    fun collectBasicBlocks(
```
konradweiss: An overall Description on how basic blocks are identified or how they should look like would be nice here. Also, the return type is a triple, so some description on what the single elements are is necessary IMO.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
```diff
@@ -64,6 +70,169 @@ open class ControlDependenceGraphPass(ctx: TranslationContext) : EOGStarterPass(
...
+                allBasicBlocks.add(basicBlock)
+            }
+            // Add the basic block to the already seen map for this node
+            alreadySeen[currentStartNode] = basicBlock
+
+            basicBlock.nodes.add(currentStartNode)
+
+            val shortCircuit = currentStartNode.astParent as? ShortCircuitOperator
+            val nextRelevantEOGEdges =
+                if (currentStartNode.nextEOGEdges.size > 1 && shortCircuit != null) {
```
konradweiss: Here I am not certain if we want to ignore this type of branching just because it is not a statement.

konradweiss: Do we need an earlier check `language is HasShortCircuitOperators` here?

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPassTest.kt
```diff
@@ -105,11 +105,12 @@ class ControlDependenceGraphPassTest {
...
             }
         assertNotNull(printAfterLoop)
-        assertEquals(2, printAfterLoop.prevCDG.size)
+        assertEquals(1, printAfterLoop.prevCDG.size)
         assertTrue(main in printAfterLoop.prevCDG)
-        assertTrue(
+        /*assertTrue(
             forEachStmt in printAfterLoop.prevCDG
-        ) // TODO: Is this really correct or should it be filtered out in the pass?
+        )*/
```
konradweiss: Can we answer this in this PR?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
oxisto: Are these basic blocks somehow really "added" to the graph? would it make sense to do so? and if yes probably in an extra pass beforehand?

KuechA: Currently, they are not added to the graph. I'd be open to add them and move this code into an extra pass which is executed after the EOG pass and before the CDG pass. We had some discussion on how to add the basic blocks and one option to do so was by adding them as overlay nodes. Most importantly, as there are EOG edges between the basic blocks, we have to ensure that this does not disturb any other pass. This may require us to properly differentiate between AST-nodes (which we cannot do yet) and other nodes when iterating through sub-graphs but I'm not sure what exactly are the implications of adding such nodes before all passes have terminated. Maybe, we could implement this easier after #2399

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
KuechA: Added a check

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
KuechA: Added some documentation

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPassTest.kt
KuechA: Well, it's sort of a philosophical question and depends if we consider that we may have endless loops or not. If we do not consider this, it's not in the CDG (that's also the current result).

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
KuechA: I added a switch and disable it for now but we could still add this feature easily later-on.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPass.kt
oxisto: I think adding them as an overlay graph would be a good idea. Can we already "prepare" this by at least moving them to a separate file and letting in inherit from `OverlayNode`. The actual adding to the graph we can do later.

KuechA: Documented this in #2400

PR #2384, opened 2025-07-07
Use three worklists to prioritize what to do in the fixpoint iteration

KuechA:
We currently perform some sort of depth-first traversal when iterating the EOG and computing fixpoints. However, a different scheduling might be more beneficial to reduce the number of merges and possible re-iterating over the same blocks of code many times. This PR tries to improve the runtime of the fixpoint iteration with minimal overhead by having the following prioritization:
1. Iterate through all statements of a basic block
2. Iterate over a basic block from a branch we've seen (idea: explore all branches)
3. Continue at a merge point after having seen all branches (currently, this is simplified by exploring all branches, could be improved by allowing to process a merge point if all branches leading there have already been explored)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/functional/BasicLatticesRedesign.kt
```diff
@@ -138,16 +138,36 @@ interface Lattice<T : Lattice.Element> {
...
             globalState[startEdge] = startState
         }
-        val edgesList = mutableListOf<EvaluationOrder>()
-        startEdges.forEach { edgesList.add(it) }
-
-        while (edgesList.isNotEmpty()) {
-            val nextEdge = edgesList.first()
-            edgesList.removeFirst()
+        val currentBBEdgesList = mutableListOf<EvaluationOrder>()
+        val potentialNextBBEdgesList = mutableListOf<EvaluationOrder>()
```
konradweiss: I think the naming here could be improved. I would rename `potentialNextBBEdgesList` to `nextBranchEdgeList` or `nextBBEdgesList`. Additionally I could see `mergePointsEdgesList` to be better named `edgesAfterMergePoint`.

konradweiss (approved):
I have some renaming suggestions but functionally I consider this an approvable change. Intuitively I would agree that this should be improving the performance.

PR #2390, opened 2025-07-15
Fixing inconsistency with empty `statements` when `statementEdge` is a getter

konradweiss:
This PR solves the issue that when calling `statements` on certain `StatementHolder`s the list would be empty. This happened in the case where the `statementEdges` had a getter that created and returned a new list. `statements` uses `statementEdges` as a delegate, but the delegate functions are still bound to the old list created during initialization.

This PR adds a backing field to `statementEdges` and initializes it with a list. The getter itself does not not create a new list, but resets the content. Additionally this PR solves an issue withe the fluent DSL that did not appear due to incorrect testing.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
@@ -61,14 +61,14 @@ class ForEachStatement : LoopStatement(), BranchingNode, StatementHolder {
     override val branchedBy: Node?
         get() = iterable
 
-    override var statementEdges: AstEdges<Statement, AstEdge<Statement>> = astEdgesOf<Statement>()
+    override var statementEdges: AstEdges<Statement, AstEdge<Statement>>
         get() {
-            field.clear()
-            field += variableEdge
-            field += iterableEdge
-            field += statementEdge
-            field += elseStatementEdge
-            return field
+            val statements = astEdgesOf<Statement>()
+            statements += variableEdge
+            statements += iterableEdge
+            statements += statementEdge
+            statements += elseStatementEdge
+            return statements
         }
         set(_) {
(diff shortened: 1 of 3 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/statements/ForEachStatement.kt
```diff
@@ -61,14 +61,14 @@ class ForEachStatement : LoopStatement(), BranchingNode, StatementHolder {
...
-    override var statementEdges: AstEdges<Statement, AstEdge<Statement>>
+    override var statementEdges: AstEdges<Statement, AstEdge<Statement>> = astEdgesOf<Statement>()
         get() {
-            val statements = astEdgesOf<Statement>()
-            statements += variableEdge
-            statements += iterableEdge
-            statements += statementEdge
-            statements += elseStatementEdge
-            return statements
+            field.clear()
```
oxisto: A getter calling *clear* on a backing field sounds very dangerous to me. Getters should be side-effect free. I am failing to understand how this new logic is different from the old logic? Can you provide a unit test that demonstrate this (e.g. which fails on current main and is fixed in this branch)

konradweiss: It is more about the way a delegate works in the background when it references a property. Especially when that property has a getter that instantiates a new field. At least from the behavior I was observing.

The behavior I was observing was that the delegate-property `statements` is always empty and my understanding is that when we delegate with `by unwrapping(ForEachStatement::statementEdges)` it inits once by calling the getter of `statementEdges` which at first object construction is an empty list. Then, on further access of `statements` it retriggers the get of `statementEdges` but the value returned by `statements` is bound to the list that was returned during initialization.

I will provide a test that will show the size of `statements` is not correct and that it does not contain all the `end` nodes reachable through `statementEdges`.

PR #2391, opened 2025-07-16
Renaming proposal for subtree searching extension functions

konradweiss:
Renaming extension functions that search in the sub tree or ancestor chain to differentiate them from normal functions on those graph objects.

This PR renames extensions functions to more clearly differentiate them from several properties on nodes as there was a naming ambiguity. Sometimes, depending on the imports, it was unclear that all nodes of a certain type in the AST-Subtree were collected instead of simply accessing the direct ast children, e.g. `statements`.

This PR refactors `allChildren` to `allDescendants` and all functions using `allChildren` and then accessing a fraction of the nodes in the subtree by filtering for the type or some condition. The extension gets the prefix `all` and the previous name is capitalized. An Example: `statements` is now `allStatements` and therefore users will know that if they use `statements` they access the children and if they use `allStatements` they will access all descendant statements. While users will not know of the existence of `allStatements` and the prefix `all` it should be discoverable through any IDE and then we clearly avoid the ambiguity.


KuechA:
This breaks the API, so it will probably be part of CPG version 11. Also, the changes require updating the documentation.

PR #2399, opened 2025-07-30
Introduction of an `AstNode`

oxisto:
This PR introduces an `AstNode`, in order to (later) restrict certain extension functions that only make sense on AST nodes.

### DIFF
--- a/cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extension.kt
+++ b/cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extension.kt
@@ -31,27 +31,37 @@ import org.reflections.Reflections
 import org.reflections.scanners.Scanners.SubTypes
 
 /**
- * Retrieves a set of all [Concept] nodes associated with this [Node] and its AST children
- * ([Node.nodes]).
+ * Retrieves a set of all [Concept] nodes associated with this [Node] and its AST children (if this
+ * is an [AstNode], see [AstNode.nodes]).
  *
- * @return A set containing all [Concept] nodes found in the overlays of the [Node] and its
- *   children.
+ * @return A set containing all [Concept] nodes found in the overlays of the [Node] and its eventual
+ *   AST children.
  */
-val AstNode.conceptNodes: Set<Concept>
-    get() = this.nodes.flatMapTo(mutableSetOf()) { it.overlays.filterIsInstance<Concept>() }
+val Node.conceptNodes: Set<Concept>
+    get() {
+        return allOverlays<Concept>()
+    }
 
 /**
- * Retrieves a set of all [Operation] nodes associated with this [Node] and its AST children
- * ([Node.nodes]).
+ * Retrieves a set of all [Operation] nodes associated with this [Node] and its AST children (if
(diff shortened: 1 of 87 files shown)


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extension.kt
```diff
@@ -37,7 +37,7 @@ import org.reflections.scanners.Scanners.SubTypes
  * @return A set containing all [Concept] nodes found in the overlays of the [Node] and its
  *   children.
  */
-val Node.conceptNodes: Set<Concept>
```
konradweiss, line 40: Can we get the conceptNodes of non-AST nodes? I ma thinking of an implementation that returns the concepts attached to this, and if this is an ASTNode, traverses the AST tree an collects more.

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extension.kt
```diff
@@ -48,7 +48,10 @@ val Node.conceptNodes: Set<Concept>
  *   children.
  */
 val Node.operationNodes: Set<Operation>
-    get() = this.nodes.flatMapTo(mutableSetOf()) { it.overlays.filterIsInstance<Operation>() }
+    get() =
```
konradweiss: Same here

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/AstNode.kt
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * Note: This only returns the *direct* children of this node. If you want to have *all*
+     * children, e.g., a flattened AST, you need to call [AstNode.allChildren].
+     *
+     * For Neo4J OGM, this relationship will be automatically filled by a pre-save event before OGM
+     * persistence. Therefore, this property is a `var` and not a `val`.
+     */
+    @Relationship("AST")
+    @JsonIgnore
+    var astChildren: List<AstNode> = listOf()
```
konradweiss, line 56: The ast Parent back reference, is that something we should/can put here?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extension.kt
oxisto: moved it back to Node

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extension.kt
oxisto: moved it back to Node

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/AstNode.kt
oxisto: `astParent` is (currently) part of `id` so this needs to stay in `Node` (for now).

konradweiss (approved):

PR #2403, opened 2025-08-05
Rename fields/methods/... in `RecordDeclaration` to `innerFields`, ...

oxisto:
This should make it clear that it only contains those that are directly embedded into the AST structure.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/RecordDeclaration.kt
```diff
@@ -50,20 +51,44 @@ open class RecordDeclaration :
...
 
-    @Relationship(value = "FIELDS", direction = Relationship.Direction.OUTGOING)
-    var fieldEdges = astEdgesOf<FieldDeclaration>()
-    var fields by unwrapping(RecordDeclaration::fieldEdges)
+    /**
+     * The [FieldDeclaration]s that are directly contained in this record declaration's AST
+     * structure. This does not include any fields that might be declared in a base class or
+     * interface or fields that are declared outside the AST structure.
+     */
+    @Relationship(value = "INNER_FIELDS", direction = Relationship.Direction.OUTGOING)
```
KuechA: Should be renamed back to `FIELDS` and `fieldEdges`.

PR #2413, opened 2025-08-26
Dokka v2

lshala:
Migration to Dokka V2 following the migration guide: https://kotlinlang.org/docs/dokka-migration.html.

Note: As is it noted in the ["Share Dokka configuration across modules"](https://kotlinlang.org/docs/dokka-migration.html#share-dokka-configuration-across-modules) section, the `subprojects {}` or `allprojects {}` using will be deprecated in the future. Instead, it is recommended to create a dedicated Dokka convention file, which can then be reused by other conventions or modules.


KuechA:
Is this supposed to create some output? I somehow didn't manage to create the dokka html files on my laptop

build.gradle.kts
```diff
@@ -40,42 +40,30 @@ repositories {
...
+        subprojects.forEach {
+            dependencies {
+                dokka(project(":${it.name}"))
+            }
+        }
+    }
+    pluginsConfiguration {
+        versioning {
+            version.set(tag)
+            olderVersionsDir.set(projectDir.resolve("previousDocs"))
```
KuechA: We will probably have to test this in the CI/CD pipeline in the future. On my laptop, I receive an exception if this directory is not present. Since we do not make use of this, it's probably not a big deal but we could probably also delete this line.

lshala: Removed this line

PR #2417, opened 2025-09-03
Timeout for CDG and CFSensitiveDFG pass

KuechA:


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/UnreachableEOGPass.kt
```diff
@@ -75,7 +75,15 @@ open class UnreachableEOGPass(ctx: TranslationContext) : EOGStarterPass(ctx) {
         }
 
         val nextEog = node.nextEOGEdges.toList()
-        val finalStateNew = unreachabilityState.iterateEOG(nextEog, startState, ::transfer)
+        val finalStateNew =
+            unreachabilityState.iterateEOG(nextEog, startState, ::transfer)
+                ?: run {
+                    log.warn(
+                        "Could not compute unreachability of EOG edges for {}, reached a timeout",
```
maximiliankaul, line 82: The logging should happen in the iterateEOG (or wherever the timeout triggers) and not here. Additionally, can this function only fail for timeouts? Or could this cause misleading error messages here?

KuechA: This function returns `null` on timeouts (and only in this case). In this case, I want to stop running the pass for this function and return here immediately. I cannot do this when handling the stuff in `iterateEOG`, so, in any case I need some handling of the `null` values here. Also note that I won't figure out which pass timed out in `iterateEOG` when logging it there.

KuechA:
Could we continue with the review process please @konradweiss @maximiliankaul ?

konradweiss:
> Could we continue with the review process please @konradweiss @maximiliankaul ?

I am currently working on tests for this PR and will merge when I am finished.

konradweiss (approved):

PR #2425, opened 2025-09-09
Update CODEOWNERS

konradweiss:
The Code owners are updated with specific responsibilities for passes.


.github/CODEOWNERS
```diff
@@ -1,3 +1,61 @@
...
+cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ImportResolver.kt @oxisto @konradweiss @maximiliankaul
+cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeResolver.kt @oxisto @konradweiss
+cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeHierarchyResolver.kt @oxisto @konradweiss
+cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DynamicInvokeResolver.kt @oxisto @peckto
+cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ResolveCallExpressionAmbiguityPass.kt @oxisto @peckto
+cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/ResolveMemberExpressionAmbiguityPass.kt @oxisto @peckto
+
+# Language Specific Passes
+
+cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PythonAddDeclarationsPass.kt @maximiliankaul @KuechA
```
oxisto: I would skip those, these are already assigned through the ownership of the language frontend

.github/CODEOWNERS
```diff
@@ -1,3 +1,61 @@
+##### Here start the pass specific ownerships, new passes should be added here #####
+
+# Core Semantic Passes
+
+cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt @konradweiss @KuechA
```
KuechA: How did I end up with this pass? 😨

konradweiss: You are the only person I would assign as the second responsible.

.github/CODEOWNERS
konradweiss: The purpose of adding the passes to the codeowners file does not only serve the purpose of getting them covered during code reviews. They are more specific than modules, including the language specific modules and the expertise on some passes don't necessarily match the maintainers of a frontend (which I know sound like it should be the case but is not when looking at some passes), for example the PythonAddDeclarationPass. This assignments purpose is to have two responsibles for a pass in case questions or problems come up and the placement in the codeowners was then decided as the best place to put it while having some functionality. Assigning the second person to the entire frontend would also mean that we would assign some people more responsibilities then initially planned, but it would also mean that we get the same reliability for language frontends that we want for passes.

The reason why I would not skip the language specific passes that are exactly matched to the language frontend is that then the assignment is implicit. However, we could assume that people that would look in the codeowner files to find someone to ask questions, would also continue looking for the responsibles of the frontend.

We could remove the pass assignments for language specific frontends if they exactly match the frontend responsibles and in case where we have only one fronted responsilbe, ask the second if he is also willing to look at the frontend if necessary?

oxisto: Yeah I suggest to not overcomplicate things, we can just add a second person to the language frontends. Personally I am not a big fan of so many very specific paths in the CODEOWNERS file because this will get out of sync very fast. I am not sure if GitHub warns us if we rename or delete a pass path.

konradweiss: I agree on that point, we did not find a better place to put these in though. I will keep an eye on it.

.github/CODEOWNERS
```diff
@@ -1,3 +1,61 @@
...
+cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt @oxisto @konradweiss
+cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/DFGPass.kt @KuechA @konradweiss @oxisto
```
oxisto: In general, we should probably use wildcards here instead of specifying the exact path, such as `*DFGPass*` or something. Please have a look at https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax

This way we are at least robust to smaller changes such as moving passes between packages or even modules and maybe also automatically include `DFGPassTest.kt` or something.

konradweiss: That is an Idea, I was slightly leaning away from that idea, to not hit other files with similar names, but hitting the test file itself is a benefit. I will use a pattern of the form `*/Passname*.kt` such that the pattern for `DFGPass` will not match `ControlFlowSensitiveDFGPass`. Basically fixing the Prefix, keeping the suffix variable.

PR #2414, opened 2025-08-29
Generate concepts and operations from cloud ontology

KuechA:
This PR aims to generate `Concept`s and `Operation`s from the ontology of security relevant (cloud) aspects. This will add several auto-generated classes

PR #2434, opened 2025-09-24
DO NOT MERGE!! Bundle open confirmate PRs

KuechA:
#2431 and #2414 are contained in this snapshot building PR.


KuechA:
Most of the branches are now merged to main, so this PR is no longer required

PR #2439, opened 2025-10-02
Implementing a simple running counter for our legacy ID 

konradweiss:
The running counter is necessary for the fluent dsl to allow for testing.


konradweiss:
This should not be necessary anymore with the new pseudo location. The regular ID should now be unique and deterministic in the Fluent DSL if it is in the regular CPG analysis

PR #2445, opened 2025-10-08
Transition to integration tests for MCP

fwendland:
MCP unit tests depend on the Python language frontend, because the code snippets for testing are in Python. If Python language frontend is unavailable, unit tests fail.

This PR moves most of the tests into integration test and adjusts build accordingly.


cpg-mcp/src/integrationTest/kotlin/mcp/ListCommandsTest.kt
```diff
@@ -87,6 +88,7 @@ class ListCommandsTest {
     }
 
     @Test
+    @EnabledIf(value = "hasPythonFrontend", disabledReason = "missing Python language frontend")
```
oxisto: I would just remove this here if we already have the build grade solution

fwendland: Done.

PR #2449, opened 2025-10-14
Inferring the location of nodes in the fluent dsl 

konradweiss:
This PR inspects the call stack on creation to determine in which file and startLine a node was defined. The other location fields: startColumn, endLine, endColumn are inferred from the surrounding and contained nodes. These fields are not accurate as the information cannot be determined from the code location in the call stack, but adhere to the invariant of code locations in a tree not overlapping and a parent nodes location containing the child node locations.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
@@ -71,7 +71,7 @@ fun LanguageFrontend<*, *>.translationResult(
 
     // Start pseudo location inference for the root node of translation, propagating to its
     // descendents.
-    node.inferPseudolocations()
+    node.inferPseudoLocations()
 
     return node
 }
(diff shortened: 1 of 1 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
```diff
@@ -63,6 +69,8 @@ fun LanguageFrontend<*, *>.translationResult(
         executePassesSequentially(ctx, node, mutableSetOf())
     }
 
+    node.inferrPseudolocations()
```
KuechA:
```suggestion
    node.inferPseudolocations()
```
What does this do? Do we need it here as well? There should only be one `TranslationResult`

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
konradweiss: The inference of a pseudo location is started at the root node of (most) translations, which is the Translation result. The call for infering could be also placed at construction of nodes that need a location and propagated from there, e.g. the translation unit. For now, I would keep it there. 

I added further doc at the definition and the call site.

PR #2455, opened 2025-10-23
Codyze requirements catalogs

KuechA:


codyze-core/src/main/kotlin/de/fraunhofer/aisec/codyze/catalogs/german/BSI_TR02102.kt
```diff
@@ -0,0 +1,1002 @@
...
+ * - [BSI TR 02102-2 version
+ *   2025-01](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR02102/BSI-TR-02102-2.pdf?__blob=publicationFile&v=11)
+ *   regarding TLS
+ */
+class BSI_TR02102(override val requirePQC: Boolean = false) :
+    SotaRequirementsCatalog(), CryptoCatalog, TLSCatalog {
+
+    context(kdf: KeyDerivationFunction)
+    override fun checkKDF(): QueryTree<Boolean> {
+        return kdf.type IN listOf("CatKDF")
```
KuechA, line 63: Is this really what we want? Seems very specific

PR #2463, opened 2025-11-06
Fix Python import alias parsing by handling missing lineno attributes

allsmog:
## Summary
Fixes the JepException when parsing Python import statements with aliases due to missing `lineno` attribute handling.

## Bug Description
CPG's Python frontend crashes when parsing Python import statements with aliases (e.g., `from typing import Optional as Opt, List as L`) because Python AST `alias` objects don't always have location attributes (`lineno`, `col_offset`, `end_lineno`, `end_col_offset`). These attributes were only added in Python 3.10+.

## Error
```
jep.JepException: <class 'AttributeError'>: 'alias' object has no attribute 'lineno'
```

## Stack Trace
- at de.fraunhofer.aisec.cpg.frontends.python.Python$AST$WithLocation$DefaultImpls.getLineno(Python.kt:80)
- at de.fraunhofer.aisec.cpg.frontends.python.Python$AST$alias.getLineno(Python.kt:1291)
- at de.fraunhofer.aisec.cpg.frontends.python.PythonLanguageFrontend.locationOf(PythonLanguageFrontend.kt:288)
- at de.fraunhofer.aisec.cpg.frontends.python.StatementHandler.handleImportFrom(StatementHandler.kt:635)

## Solution
Modified the `WithLocation` interface in `Python.kt` to wrap each location attribute getter in try-catch blocks, returning `0` as a safe default value instead of calling `TODO()`. This allows parsing to succeed even when location attributes are missing.

## Changes
- **File**: `cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/Python.kt`
- **Lines**: 76-115 (WithLocation interface)
- Added null-safety checks for `lineno`, `col_offset`, `end_lineno`, and `end_col_offset` attributes

## Testing
- Tested with various Python import alias patterns
- Existing test files already exercise this functionality:
  - `cpg-language-python/src/test/resources/python/import_vs_member.py`
  - `cpg-language-python/src/test/resources/python/superclasses/superclass_import_module_alias.py`
- Test `testImportVsMember` in `PythonFrontendTest.kt` validates import alias behavior

## Side Effects
- Location information for AST nodes without location attributes will return (0, 0) coordinates
- This is acceptable as it's better than crashing, and import functionality still works correctly

## Backward Compatibility
✅ No breaking changes - fully backward compatible

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "external/security-metrics"]
+	path = external/security-metrics
+	url = https://github.com/Cybersecurity-Certification-Hub/security-metrics.git
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,3 +10,5 @@ coverage:
         target: 75%
 comment:
     layout: "condensed_header, files, condensed_footer"
+ignore:
+  - "**/ontology/**"
--- a/codyze-console/build.gradle.kts
+++ b/codyze-console/build.gradle.kts
@@ -65,3 +65,5 @@ tasks.processResources { dependsOn(pnpmBuild) }
 var jarTasks = tasks.withType<Jar>()
 
 jarTasks.forEach { it.dependsOn(pnpmBuild) }
+
+tasks.shadowJar { setProperty("zip64", true) }
--- a/codyze-console/src/main/webapp/package.json
+++ b/codyze-console/src/main/webapp/package.json
@@ -23,7 +23,7 @@
     "@tailwindcss/forms": "^0.5.10",
     "@tailwindcss/typography": "^0.5.15",
     "@tailwindcss/vite": "^4.0.0",
-    "cookie": "1.0.2",
+    "cookie": "1.1.1",
     "eslint": "^9.18.0",
     "eslint-config-prettier": "^10.0.1",
     "eslint-plugin-svelte": "^3.0.0",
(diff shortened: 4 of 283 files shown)


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/Python.kt
```diff
@@ -77,25 +77,41 @@ interface Python {
             /** Maps to the `lineno` filed from Python's ast. */
             val lineno: Int
                 get() {
-                    return (pyObject.getAttr("lineno") as? Long)?.toInt() ?: TODO()
+                    return try {
+                        (pyObject.getAttr("lineno") as? Long)?.toInt() ?: 0
+                    } catch (e: Exception) {
+                        0
```
oxisto: Hmm somehow this is not a very clean solution. If the field is sometimes null, we should declare the field as a nullable `Int?` instead.

If this is not possible somehow (@maximiliankaul), we should at least return -1 and not 0

maximiliankaul: I agree `-1` is better here than `0`.
I don't have a preference regarding `Int` or `Int?`. I assume `Int?` is more "Kotlin"? then we'd also need to change `locationOf` in `PythonLanguageFrontend` (to fallback to `-1`) but nothing else seems to break. I suggest to go with that soulution.

maximiliankaul: Changed

PR #2480, opened 2025-12-05
Make the Python logging pass use EOG order

### LINKED ISSUE #2479: `LogWrite` operations not created for logger calls inside functions
The PythonLoggingConceptPass creates `Log` concepts and operations. In the following example we would expect that a `LogWrite` operation is created for the call `logger.info` . However, this only happens when `logger.info` is not called inside another function:

```
import logging

logger = logging.getLogger(__name__)

def foo():
      logger.info("bar")
```

maximiliankaul:
Previously, we did not enforce handling nodes in EOG order. This could lead to a `getLogger` call in global scope to be handled after the log usage in a function scope. This broke log resolution and is no fixed by handling nodes in EOG order.

Fixes #2479


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/logging/python/PythonLoggingConceptPass.kt
```diff
@@ -78,15 +81,32 @@ class PythonLoggingConceptPass(ctx: TranslationContext) : ComponentPass(ctx) {
...
-     * relevant parts of the Python code for logging.
+     * This pass needs to handle nodes in EOG order to make sure that log creation is handled before
+     * log usage. The actual handling is done in [handleNode].
      */
     override fun accept(comp: Component) {
+        walker = SubgraphWalker.ScopedWalker(ctx.scopeManager, Strategy::EOG_FORWARD)
+        walker.registerHandler { node -> handleNode(node) }
+        // Gather all resolution EOG starters; and make sure they really do not have a
+        // predecessor, otherwise we might analyze a node multiple times
+        val nodes = comp.allEOGStarters.filter { it.prevEOGEdges.isEmpty() }
```
KuechA, line 92: Does this really ensure the order of the test case or if the logger is declared in an EOG-connected code fragment and used in another one? They may not always be traversed in the required order by the SubgraphWalker (or any other option that we have), correct?

KuechA: Fixing this would probably need some sort of the "mark dirty" logic and would probably be required in several other passes, too. Not sure how to handle this, a respective system should probably be implemented as part of a separate PR

maximiliankaul: This is a wide-spread problem. I copy&pasted this code from other passes. I agree that we should come up with a clean solution.

KuechA: Should be discussed separately see #2514

PR #2483, opened 2025-12-08
Manual extension of concepts

fwendland:
We've cherry-picked manual concept extensions from #2454 after the merge of the generated ontology concepts.

- Adjusting for new folder for concepts from ontology
- Fixing name and property changes in ontology


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/manualExtensions/ConceptContainer.kt
```diff
@@ -0,0 +1,85 @@
...
+
+class InputValidation(
+    val input: Node?,
+    val validatedOutput: Node?,
+    concept: Concept,
+    underlyingNode: Node?,
+) : Operation(concept, underlyingNode)
+
+// Deprecated? property already defined in [HttpClient]
+// val HttpClient.isTLS: Boolean
```
KuechA: Remove this commented out code

PR #2487, opened 2025-12-12
Include property `id` without name collisions

fwendland:
We've ignored property `id` from the ontology due to naming collisions with class properties in the CPG. However, we need to include the property and prevent name collisions at the same time.

We propose renaming property `id` from the ontology with a name based on its associated entity. Doing so also resolves a bug, where we used `id` as part of `equals()` and `hashCode()` causing stack overflow exceptions.

Changes to the ontology based concept classes are added too.

### DIFF
--- a/cpg-language-typescript/build.gradle.kts
+++ b/cpg-language-typescript/build.gradle.kts
@@ -38,6 +38,8 @@ mavenPublishing {
     }
 }
 
+deno { version("v2.5.6") }
+
 val compileWindowsX8664 =
     tasks.register<RunDenoTask>("compileWindowsX8664") {
         dependsOn(tasks.installDeno)


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/generator/KotlinCodeGeneration.kt
```diff
@@ -343,9 +339,15 @@ private fun addPropertiesToClass(
         }
 
         // Enforce camel case: convert first letter to lowercase
-        val propertyName = dataProp.propertyName.replaceFirstChar { it.lowercase() }
-
-        if (propertyName == "id") continue
+        var propertyName = dataProp.propertyName.replaceFirstChar { it.lowercase() }
+
+        // need to rename property `id` to prevent naming collisions
```
interruptedHandshake, line 344: In the ontology it was named like this: ``id``. The new name resourceId is pretty fine I guess

PR #2502, opened 2025-12-16
Fix CDG for ShortCircuitOperators

### LINKED ISSUE #2486: CDG and ShortCircuitOperators
```
foo() && bar()
```

one would expect `prevCDG(bar())` to contain `foo()` but this is not the case.
(Tested with a language that has `HasShortCircuitOperator`. The `&&` correctly translates to a `ShortCircuitOpertor` node.)

Problem identified by @lshala and @morbitzer

maximiliankaul:
Fixes #2486

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/BasicBlockCollectorPass.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/BasicBlockCollectorPass.kt
@@ -36,7 +36,12 @@ import de.fraunhofer.aisec.cpg.graph.statements.expressions.ShortCircuitOperator
 import de.fraunhofer.aisec.cpg.passes.configuration.DependsOn
 import java.util.*
 
+/**
+ * This pass collects [BasicBlock]s by iterating through the EOG of [EOGStarterHolder]s. A basic
+ * block is defined as a sequence of nodes with a single entry point and a single exit point.
+ */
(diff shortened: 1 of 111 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/BasicBlockCollectorPass.kt
```diff
@@ -95,9 +103,30 @@ class BasicBlockCollectorPass(ctx: TranslationContext) : EOGStarterPass(ctx) {
                         }
                     }
             }
+
             basicBlock.nodes.add(currentStartNode)
 
-            val nextRelevantEOGEdges = currentStartNode.nextEOGEdges
+            // Note: This may not identify all cases correctly, e.g., if a ShortCircuitOperator is
+            // not the direct astParent.
+            val shortCircuit = currentStartNode.astParent as? ShortCircuitOperator
```
KuechA, line 116: Would it be better to make the look-ahead in the EOG here instead of checking the `astParent`?

maximiliankaul: I'll try that.

maximiliankaul: Merging this now as is. Future work in https://github.com/Fraunhofer-AISEC/cpg/issues/2539

PR #1331, opened 2023-10-10
More improvements to the Go frontend

oxisto:
This includes better handling of (underlying) types as well as better handling of packages.

PR #2522, opened 2025-12-23
Apply quick-fixes and fix typos

fwendland:
A bit of clean up for code quality and reduced compiler warnings.

### DIFF
--- a/cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/GoExtraPass.kt
+++ b/cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/GoExtraPass.kt
@@ -104,8 +104,8 @@ class GoExtraPass(ctx: TranslationContext) : ComponentPass(ctx) {
 
     private lateinit var walker: SubgraphWalker.ScopedWalker<AstNode>
 
-    // Note: Code analysis suggests that this property is non-nullable, but there is a complex reason, why it has to be
-    // nullable.
+    // Note: Code analysis suggests that this property is non-nullable, but there is a complex
+    // reason, why it has to be nullable.
     override val scope: Scope?
         get() = scopeManager.currentScope
 


cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/SpecificationHandler.kt
```diff
@@ -299,6 +301,7 @@ class SpecificationHandler(frontend: GoLanguageFrontend) :
         }
     }
 
+    // TODO Never used -> missing implementation?
```
oxisto: Im pretty sure it is used 🤔

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/GoExtraPass.kt
```diff
@@ -101,6 +101,7 @@ class GoExtraPass(ctx: TranslationContext) : ComponentPass(ctx) {
 
     private lateinit var walker: SubgraphWalker.ScopedWalker<AstNode>
 
+    // Note: Code analysis suggests that this property is non-nullable.
```
oxisto: that's a complex issue. I tried that before but there was a reason it was not working, but I forget what is what 🤣

cpg-language-go/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontendTest.kt
```diff
@@ -221,7 +221,7 @@ class GoLanguageFrontendTest : BaseTest() {
 
         val n = p.variables["n"]
         assertNotNull(n)
-        with(tu) { assertEquals(tu.primitiveType("int").pointer(), n.type) }
+        assertEquals(tu.primitiveType("int").pointer(), n.type)
```
oxisto, line 224: don't we have like an assert type or something.

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/SpecificationHandler.kt
fwendland: It was used until PR #1331, where the handler for `FuncType` (cf. [L77/78](https://github.com/Fraunhofer-AISEC/cpg/pull/1331/files#diff-26b4b5e93c9af911fc44b00b3a63bc6984dc4507b6f7cbe572d457a3df974792)) was changed. 

I'll remove the comment and suppress the warning in case we need the function.

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/GoExtraPass.kt
fwendland: I'll updated the comment and mention a complex reason to keep it nullable. Should prevent us from following compiler advice in the future.

cpg-language-go/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguageFrontendTest.kt
fwendland: Maybe :thinking: Couldn't find something appropriate though :( I've came across [`assertIs`](https://kotlinlang.org/api/core/kotlin-test/kotlin.test/assert-is.html), but we would need to rework the assert expression. 

I've just removed the unnecessary receiver. Now, the assert also matches the previous asserts more closely.

PR #2537, opened 2026-01-19
Try to fix QueryTree IDs

KuechA:
Currently, the ID of a QueryTree is based on the node it relates to, the QueryTree's value and the IDs of the children. This means that, if multiple QueryTrees relate to the same node and have equal children (e.g. `false` or no children at all), they all share the same ID. In the Codyze Web UI, there is no way to distinguish them and results may be displayed in a weird way (i.e., one of these nodes will be displayed multiple times, instead of displaying all of the nodes).

I'm trying to make this a bit less likely by including the string representation into the ID as well. Are there any other fields that we can/should include here? If I remember correctly, using the parent node was somewhat problematic due to our logic suppressing results.

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
@@ -64,6 +64,10 @@ import kotlin.uuid.Uuid
 open class QueryTree<T>(
     value: T,
     var children: List<QueryTree<*>> = emptyList(),
+    /**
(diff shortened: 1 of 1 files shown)


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/query/QueryTree.kt
```diff
@@ -97,6 +97,14 @@ open class QueryTree<T>(
             return calculateConfidence()
         }
 
+    var stringRepresentation: String = stringRepresentation
```
konradweiss, line 108: Some doc describing if this holds a description or the syntactic representation of the query tree is needed here IMO

PR #2540, opened 2026-01-23
Make Neo4j `pushToNeo4j` more easily available.

maximiliankaul:
This PR moves connection default values to a new file and provides a new function `pushToNeo4jWithNewSession` to persist a `TranslationResult` without already having a session available.

This is intended as a convenience over having to create a session manually in other modules.

### DIFF
--- a/cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/Neo4J.kt
+++ b/cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/Neo4J.kt
@@ -71,9 +71,21 @@ internal typealias Relationship = Map<String, Any?>
  *
  * @param noPurgeDb A boolean flag indicating whether to skip the database purge step. If set to
  *   true, the existing data in the database will not be deleted before persisting the new data.
+ * @param protocol The protocol to use for connecting to the Neo4j database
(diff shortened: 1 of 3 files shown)


cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/Neo4J.kt
```diff
@@ -62,6 +64,23 @@ internal val log = LoggerFactory.getLogger("Persistence")
...
 
+/**
+ * This function creates a new Neo4j session, optionally purges the database, and persists the
+ * current [TranslationResult] into the database using the [persist] function (which requires a
+ * session context often not available at the call-site).
+ *
+ * @param noPurgeDb A boolean flag indicating whether to skip the database purge step. If set to
+ *   true, the existing data in the database will not be deleted before persisting the new data.
+ */
+fun TranslationResult.pushToNeo4jWithNewSession(noPurgeDb: Boolean = false) {
```
oxisto: Just `pushToNeo4j`? you can also remove the `pushToNeo4j` from the `Application` then use this function

maximiliankaul: Ok, I'll update this.

PR #2544, opened 2026-01-27
Python: Add implicit constructor return and fix receiver scoping

lshala:
When an instance is created via a constructor, we can't follow the DFG edges from the arguments to the returned instance. This PR aims to fix that by adding an implicit `ReturnStatement` at the end of an `__init__` function. 
It also fixes:
- related issues with how the CFG Pass tracks state changes.
- receiver scoping: The receiver was only added to the scope when it had a default value. This is now fixed, allowing the symbol resolver to handle self references correctly.

### DIFF
--- a/cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/ArgumentsHandlerTest.kt
+++ b/cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/ArgumentsHandlerTest.kt
@@ -230,21 +230,4 @@ class ArgumentsHandlerTest {
         assertNotNull(defaultArg4)
         assertContains(defaultArg4.nextDFG, funcDefault)
     }
-
-    @Test
-    fun testReceiverInScope() {
-        val method = result.methods["method_using_self"]
-        assertNotNull(method)
-
-        val receiver = method.receiver
-        assertNotNull(receiver, "Method should have a receiver")
-
-        val selfRef = method.refs.singleOrNull() { it.name.localName == "self" }
-        assertNotNull(selfRef, "Expected one reference to 'self'")
-        assertEquals(
-            receiver,
-            selfRef.refersTo,
-            "Reference to 'self' should resolve to the receiver",
-        )
-    }
 }
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
@@ -26,7 +26,6 @@
 package de.fraunhofer.aisec.cpg.frontends.python
 
 import de.fraunhofer.aisec.cpg.graph.*
-import de.fraunhofer.aisec.cpg.graph.declarations.MethodDeclaration
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.*
 import jep.python.PyObject
 
@@ -543,23 +542,6 @@ class ExpressionHandler(frontend: PythonLanguageFrontend) :
 
     private fun handleName(node: Python.AST.Name): Expression {
         val r = newReference(name = node.id, rawNode = node)
-
-        /*
-         * TODO: this is not nice... :(
-         *
-         * Take a little shortcut and set refersTo, in case this is a method receiver. This allows us to play more
-         * nicely with member (call) expressions on the current class, since then their base type is known.
-         */
-        val currentFunction = frontend.scopeManager.currentFunction
-        if (currentFunction is MethodDeclaration) {
-            val recv = currentFunction.receiver
-            recv.let {
-                if (node.id == it?.name?.localName) {
-                    r.refersTo = it
-                    r.type = it.type
-                }
-            }
-        }
         return r
     }
 
--- a/cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt
+++ b/cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt
@@ -548,6 +548,7 @@ class PythonFrontendTest : BaseTest() {
         assertEquals(1, fooMemCall.invokes.size)
         assertInvokes(fooMemCall, bar)
         assertLocalName("self", fooMemCall.base)
+        assertRefersTo(fooMemCall.base, foo.receiver)
     }
 
     @Test
--- a/cpg-language-python/src/test/resources/python/arguments.py
+++ b/cpg-language-python/src/test/resources/python/arguments.py
@@ -45,6 +45,3 @@ class MyClass:
     def call5(self, a):
         kw_args_and_default(b=True, a=3)
 
-    def method_using_self(self, x):
-        self.x = x
-


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/DeclarationHandler.kt
```diff
@@ -163,6 +164,22 @@ class DeclarationHandler(frontend: PythonLanguageFrontend) :
...
+                block +=
+                    newReturnStatement().apply {
+                        this.isImplicit = true
+                        this.returnValue =
+                            newReference("self").apply {
+                                this.isImplicit = true
+                                this.access = AccessValues.READ
+                                this.refersTo = func.receiver
+                            }
+                        this.codeAndLocationFrom(func)
```
KuechA, line 177: Probably not necessary but I think this is inconsistent in the frontends.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/DeclarationHandler.kt
```diff
@@ -163,6 +164,22 @@ class DeclarationHandler(frontend: PythonLanguageFrontend) :
...
+        if (func is ConstructorDeclaration) {
+            (func.body as? Block)?.let { block ->
+                block +=
+                    newReturnStatement().apply {
+                        this.isImplicit = true
+                        this.returnValue =
+                            newReference("self").apply {
+                                this.isImplicit = true
+                                this.access = AccessValues.READ
+                                this.refersTo = func.receiver
```
KuechA: This should be done by the symbol resolver.
```suggestion
```

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/DeclarationHandler.kt
lshala: I would keep this as it is

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/DeclarationHandler.kt
```diff
@@ -268,11 +284,12 @@ class DeclarationHandler(frontend: PythonLanguageFrontend) :
                 val defaultValue =
                     args.defaults.getOrNull(0)?.let { frontend.expressionHandler.handle(it) }
                 defaultValue?.let {
-                    frontend.scopeManager.addDeclaration(recvNode)
                     result.additionalProblems +=
                         newProblemExpression("Receiver with default value", rawNode = args)
                 }
             }
+            // Add the receiver to the scope so that references to it can be resolved
+            frontend.scopeManager.addDeclaration(recvNode)
```
lshala, line 292: @maximiliankaul Can you please take look and check if this change leads to any side effects

oxisto: So what's weird is that, even if we remove this line completely, resolving to `self` still works - even though it is not in the symbol list. So it seems we are doing some (unnecessary) magic in some other parts of the code

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/ArgumentsHandlerTest.kt
```diff
@@ -230,4 +230,21 @@ class ArgumentsHandlerTest {
...
+    fun testReceiverInScope() {
+        val method = result.methods["method_using_self"]
+        assertNotNull(method)
+
+        val receiver = method.receiver
+        assertNotNull(receiver, "Method should have a receiver")
+
+        val selfRef = method.refs.singleOrNull() { it.name.localName == "self" }
+        assertNotNull(selfRef, "Expected one reference to 'self'")
+        assertEquals(
```
oxisto: use `assertRefersTo` please

There is already a `testSelf` test in PythonFrontendTest. We should probably just extend that one to actually include some assertRefersTo

oxisto: like `assertRefersTo(fooMemCall.base, foo.receiver)`

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/DeclarationHandler.kt
oxisto: Ok the magic is here yuck!

https://github.com/Fraunhofer-AISEC/cpg/blob/f018867920fcb154fa535052a9da696a24590736/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt#L547-L563

So changing this here definitely makes sense but we should probably also try to remove this ugly workaround

oxisto: After some initial testing that workaround could be completely removed when adding self to the symbols like you did:

```patch
diff --git a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
index 37cc7cad3e1..f183ca9b0f9 100644
--- a/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
+++ b/cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/ExpressionHandler.kt
@@ -26,7 +26,6 @@
 package de.fraunhofer.aisec.cpg.frontends.python
 
 import de.fraunhofer.aisec.cpg.graph.*
-import de.fraunhofer.aisec.cpg.graph.declarations.MethodDeclaration
 import de.fraunhofer.aisec.cpg.graph.statements.expressions.*
 import jep.python.PyObject
 
@@ -543,23 +542,6 @@ class ExpressionHandler(frontend: PythonLanguageFrontend) :
 
     private fun handleName(node: Python.AST.Name): Expression {
         val r = newReference(name = node.id, rawNode = node)
-
-        /*
-         * TODO: this is not nice... :(
-         *
-         * Take a little shortcut and set refersTo, in case this is a method receiver. This allows us to play more
-         * nicely with member (call) expressions on the current class, since then their base type is known.
-         */
-        val currentFunction = frontend.scopeManager.currentFunction
-        if (currentFunction is MethodDeclaration) {
-            val recv = currentFunction.receiver
-            recv.let {
-                if (node.id == it?.name?.localName) {
-                    r.refersTo = it
-                    r.type = it.type
-                }
-            }
-        }
         return r
     }
 
diff --git a/cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt b/cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt
index 86a855730e1..20f47ab2f0b 100644
--- a/cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt
+++ b/cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonFrontendTest.kt
@@ -548,6 +548,7 @@ class PythonFrontendTest : BaseTest() {
         assertEquals(1, fooMemCall.invokes.size)
         assertInvokes(fooMemCall, bar)
         assertLocalName("self", fooMemCall.base)
+        assertRefersTo(fooMemCall.base, foo.receiver)
     }
 
     @Test
```

lshala: Nice! Thanks for helping! 😄

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/statementHandler/ArgumentsHandlerTest.kt
lshala: Perfect. I extended the 'testSelf' with the `assertRefersTo`.

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/DeclarationHandler.kt
oxisto: I would keep it. its more expressive of what is actually happening.

PR #2545, opened 2026-01-27
Try to get apks working in the jvm frontend - Part 2

KuechA:
We cannot analyze apk files. First, there are several exceptions. Second, there seems to be an issue with memory consumption as the apk files have a bunch of library functions. This PR provides a way to filter such functions (based on a configuration) and exclude their function bodies from the analysis to save memory.

### DIFF
--- a/cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoFrontendConfiguration.kt
+++ b/cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoFrontendConfiguration.kt
@@ -31,7 +31,7 @@ import de.fraunhofer.aisec.cpg.graph.declarations.FunctionDeclaration
 
 abstract class GoFrontendConfiguration() : FrontendConfiguration<GoLanguageFrontend>()
 
-class NoDependenciesGoFrontendConfiguration : FrontendConfiguration<GoLanguageFrontend>() {
+class NoDependenciesGoFrontendConfiguration : GoFrontendConfiguration() {
     /**
      * By default, we do not parse the body of functions in dependencies, as they are often large
      * and not relevant for analysis. This can be overridden by providing a custom


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/collections/EdgeSet.kt
```diff
@@ -39,7 +39,7 @@ abstract class EdgeSet<NodeType : Node, EdgeType : Edge<NodeType>>(
     override var outgoing: Boolean = true,
     override var onAdd: ((EdgeType) -> Unit)? = null,
     override var onRemove: ((EdgeType) -> Unit)? = null,
-) : HashSet<EdgeType>(), EdgeCollection<NodeType, EdgeType> {
+) : HashSet<EdgeType>(1), EdgeCollection<NodeType, EdgeType> {
```
oxisto, line 45: What's the reasoning behind this? Can we have a comment on this?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/flows/ControlDependence.kt
```diff
@@ -41,7 +41,7 @@ class ControlDependence(
     start: Node,
     end: Node,
     /** A set of [EvaluationOrder.branch] values. */
-    var branches: Set<Boolean> = setOf(),
+    var branches: Set<Boolean> = HashSet(1),
```
oxisto: What's the reasoning behind this? Can we have a comment on this?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -471,6 +482,20 @@ private constructor(
             return this
         }
 
+        fun <T : LanguageFrontend<*, *>> configureFrontend(
```
oxisto, line 495: add kdoc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -471,6 +482,20 @@ private constructor(
...
+            clazz: KClass<T>,
+            config: FrontendConfiguration<T>,
+        ): Builder {
+            this.frontendConfigurations[clazz] = config
+            return this
+        }
+
+        inline fun <reified T : LanguageFrontend<*, *>> configureFrontend(
```
oxisto, line 511: add kdoc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/edges/collections/EdgeSet.kt
KuechA: Collections are allocated for `n` elements (I think 12 is the default). However, I think that we almost never have that many edges, so it's a waste of memory. I'll add a comment on this. We may want to change this value depending on the type of edge to avoid copy operations.

PR #2559, opened 2026-02-10
---

SomeUserName1:


cpg-language-rust/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/RustControlFlowTest.kt
```diff
@@ -0,0 +1,204 @@
...
+                it.registerLanguage<RustLanguage>()
+            }
+        assertNotNull(tu)
+
+        val ifLetFunc = tu.functions["if_let"]
+        assertNotNull(ifLetFunc)
+        val body = ifLetFunc.body as? Block
+        assertNotNull(body)
+
+        val ifStmt = body.statements[1] as? IfStatement
```
oxisto: you should use getOrNull here, otherwise the as? cast will not succeed, since this it will already crash with a NPE if the statement is less than 2. That happens multiple times here in the tests.

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
```diff
@@ -0,0 +1,415 @@
...
+        frontend.scopeManager.enterScope(record)
+
+        val body = node.getChildByFieldName("body")
+        if (body != null) {
+            for (i in 0 until body.childCount) {
+                val child = body.getChild(i)
+                if (child.isNamed) {
+                    val decl = handle(child)
+                    if (decl != null) {
+                        record.addDeclaration(decl)
```
oxisto: we are debating this in other frontends, so the problem is that you are adding the function to the AST of the class (using `addDeclaration`), but in reality that function s NOT part of the AST of the class, but only of its scope. So in general we try to avoid that, only add the symbol but not the actual AST. That leaves us with a problem where to place the AST node of the function, especially if we want to keep some kind of `impl` AST structure.

We have been discussing something like a `PartialRecordDeclaration` for those AST declarations outside of the class itself.

cc @konradweiss

cpg-language-rust/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/RustControlFlowTest.kt
SomeUserName1: fixed with the next push

SomeUserName1:
Yeah sorry for the volume. if you see unneccessary code or inflations, please tell me, s.t. i can cut it.

Are those changes within the reasonable limits of what a frontend should have?

not all features are supported yet, so more will come. 

Regarding tree-sitter vs. rust-analyzer: the analyser is richer of course when it comes to annotations and the like- I havent looked under the hood so far. Assuming it also produces an AST (or an annotated AST), unifying the code, i.e. using rust-analyzer as ground truth instead of tree-sitter should be conceptually easy (not saying it will be easy to do though). Do you see clear up/down sides of one over the other?

konradweiss:
> Thank you for this contribution. This needs a larger code review as this is quite a substantial amount of (AI generated) code. @konradweiss also has been working on a Rust frontend, based on the rust-analyser framework. We need to see how we can progress here towards a common goal.

@SomeUserName1 Thank you for your interest in the CPG! All help is welcome. We are already implementing a Rust frontend in this PR: https://github.com/Fraunhofer-AISEC/cpg/pull/2560
. It was originally just a branch, so I opened it as a draft PR for better visibility.

You are welcome to use this version of our language frontend. Please be aware, however, that it is a work in progress and things may change. We also appreciate your feedback on implementation decisions and on language constructs that you think are necessary but currently unsupported, especially as we approach a stage where the frontend is considered “feature-sufficient.”

We have leaned away from Tree-sitter because we faced some issues in the past. Rust-analyzer is well supported, and although it was hard to interface it with our kotlin code, it now works.

gradle.properties.example
```diff
@@ -10,6 +10,7 @@ enableTypeScriptFrontend=true
 enableRubyFrontend=true
 enableJVMFrontend=true
 enableINIFrontend=true
+enableRustFrontend=false
```
oxisto: you need to enable this in order to enable the CI test runs of the frontend

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
oxisto: We will have a `ExtensionDeclaration` very soon, which will solve this problem, this is blocked until then though.

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/experimental/rust/DeclarationHandler.kt
```diff
@@ -0,0 +1,692 @@
+/*
```
oxisto, line 1: Can we please move `cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/` to `cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/experimental/rust/`, then it won't interfere with our new frontend

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/experimental/rust/DeclarationHandler.kt
```diff
@@ -0,0 +1,692 @@
...
+            "const_item" -> handleConstItem(node)
+            "static_item" -> handleStaticItem(node)
+            "use_declaration" -> handleUseDeclaration(node)
+            else -> {
+                ProblemDeclaration("Unknown declaration type: ${node.type}")
+            }
+        }
+    }
+
+    private fun handleFunctionItem(node: TSNode): Declaration {
```
oxisto, line 80: can you please add a short javadoc to those functions. even though they are internal, it would be nice because these are the "meat" of these handlers.

Sort of a gold-standard to do that is in https://github.com/Fraunhofer-AISEC/cpg/blob/a65cb7e69d04adde21bfc9abdc6de55f683d623b/cpg-language-llvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/llvm/ExpressionHandler.kt#L190-L196

or

https://github.com/Fraunhofer-AISEC/cpg/blob/a65cb7e69d04adde21bfc9abdc6de55f683d623b/cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/ExpressionHandler.kt#L95-L101

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
```diff
@@ -0,0 +1,692 @@
...
+                ProblemDeclaration("Unknown declaration type: ${node.type}")
+            }
+        }
+    }
+
+    private fun handleFunctionItem(node: TSNode): Declaration {
+        val nameNode = node.getChildByFieldName("name")
+        val name = nameNode?.let { frontend.codeOf(it) } ?: ""
+
+        val recordDeclaration =
```
oxisto:
```suggestion
        val recordDeclaration = frontend.scopeManager.currentRecord
```

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
```diff
@@ -0,0 +1,677 @@
...
+                type.superTypes.add(boundType)
+            }
+        }
+    }
+
+    private fun handleParameters(node: TSNode, func: FunctionDeclaration) {
+        for (child in node.children) {
+            if (child.type == "parameter") {
+                val pattern = child["pattern"]
+                var name = pattern?.let { frontend.codeOf(it) } ?: ""
```
oxisto: child.text()

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
```diff
@@ -0,0 +1,677 @@
...
+                frontend.scopeManager.addDeclaration(param)
+                func.parameters += param
+            } else if (child.type == "self_parameter") {
+                handleSelfParameter(child, func)
+            }
+        }
+    }
+
+    private fun handleSelfParameter(node: TSNode, func: FunctionDeclaration) {
+        val recordDeclaration =
```
oxisto: frontend.currentRecord

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
```diff
@@ -0,0 +1,677 @@
...
+                            fieldName,
+                            frontend.typeOf(fieldTypeNode),
+                            rawNode = child,
+                        )
+
+                    // Check for visibility modifier (e.g. pub)
+                    for (fieldChild in child.children) {
+                        if (fieldChild.type == "visibility_modifier") {
+                            val vis = frontend.codeOf(fieldChild) ?: "pub"
+                            field.annotations += newAnnotation(vis, rawNode = fieldChild)
```
oxisto: FieldDeclaration has `modifiers`, better to use that.

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
```diff
@@ -0,0 +1,677 @@
...
+    }
+
+    private fun handleFunctionSignatureItem(node: TSNode): MethodDeclaration {
+        val nameNode = node["name"]
+        val name = nameNode?.let { frontend.codeOf(it) } ?: ""
+
+        val recordDeclaration =
+            (frontend.scopeManager.currentScope as? RecordScope)?.astNode as? RecordDeclaration
+
+        val func =
```
oxisto: is that always a method?

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/experimental/rust/StatementHandler.kt
```diff
@@ -0,0 +1,603 @@
...
+            if (child.isNamed && child.type != "return") {
+                ret.returnValue = frontend.expressionHandler.handle(child) as? Expression
+                break
+            }
+        }
+        return ret
+    }
+
+    private fun handleIfExpression(node: TSNode): IfStatement {
+        val ifStmt = newIfStatement(rawNode = node)
```
oxisto: this should probably be a conditional expression instead. we are running into the problem that apparently rust has no "statements", but all statements are expressions; but the CPG differentiates between statements and expressions :(

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
```diff
@@ -0,0 +1,675 @@
...
+                newFunctionDeclaration(name, rawNode = node)
+            }
+
+        // Check for async modifier
+        for (child in node.children) {
+            if (child.type == "function_modifiers") {
+                for (modifier in child.children) {
+                    if (modifier.type == "async") {
+                        // We don't fully support async modifiers yet, so we just add an annotation
+                        val annotation = newAnnotation("Async", rawNode = modifier)
```
oxisto: This should use `modifiers` once https://github.com/Fraunhofer-AISEC/cpg/pull/2562 is merged in

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/experimental/rust/DeclarationHandler.kt
SomeUserName1: done

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/experimental/rust/StatementHandler.kt
SomeUserName1: changed to conditional expression

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
SomeUserName1: use modifiers now

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/experimental/rust/DeclarationHandler.kt
SomeUserName1: done

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
SomeUserName1: done

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
SomeUserName1: done & replaced in other places as well

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
SomeUserName1: get recordDeclaration from func if func is a MethodDeclaration

cpg-language-rust/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/rust/DeclarationHandler.kt
SomeUserName1: Creates FunctionDeclaration now, when not in record scope, otherwise MethodDeclaration

oxisto:
> Mh that broke the build. Handy that you are a co creator of all commits now 😄 What's the coverage goal? 80%? Overall or file-per-file. The open gaps are mostly error handling branches for scenarios that do not happen if tree sitter doesn't do wild things.
> 
> So should I rather remove those branches or add tests even for the most exotic branches (tree sitter parser version bump with breaking changes)

Yeah I'm fixing it now. We currently aim 75-80%, I am trying to rework the tests, they currently aiming at reaching all the coverage, which is ok but we usually want tests categorised by language functions rather than have tests like `testComplex1` or `TestBranchingCoverage` ;)

PR #2562, opened 2026-02-12
Adding `modifiers` to all `Declaration` nodes

oxisto:
This PRs adds a `modifiers` property to all declaration nodes and tries to fill this information in our supported languages.

### DIFF
--- a/cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/DeclarationHandler.kt
+++ b/cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/DeclarationHandler.kt
@@ -28,7 +28,6 @@ package de.fraunhofer.aisec.cpg.frontends.jvm
 import de.fraunhofer.aisec.cpg.frontends.Handler
 import de.fraunhofer.aisec.cpg.graph.*
 import de.fraunhofer.aisec.cpg.graph.declarations.*
-import de.fraunhofer.aisec.cpg.helpers.map
 import sootup.core.jimple.basic.Local
 import sootup.core.model.SootClass
 import sootup.core.model.SootField
@@ -150,7 +149,7 @@ class DeclarationHandler(frontend: JVMLanguageFrontend) :
         return newFieldDeclaration(
             field.name,
             frontend.typeOf(field.type),
-            field.modifiers.map { it.name.lowercase() },
+            field.modifiers.map { it.name.lowercase() }.toSet(),
             rawNode = field,
         )
     }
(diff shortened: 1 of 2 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
```diff
@@ -553,3 +553,17 @@ object Util {
...
+ * Applies the given predicate to each element in this set and returns a new set containing the
+ * results.
+ *
+ * @param T the type of the elements in the original set
+ * @param R the type of the elements in the resulting set
+ * @param predicate a function that maps an element of type T to type R
+ * @return a new set with elements of type R, resulting from applying the predicate to each element
+ *   of the original set
+ */
+fun <T, R> Set<T>.map(predicate: (T) -> R): Set<R> {
```
KuechA: Is this function really necessary? What's the difference to the standard lib one?

cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/DeclarationHandler.kt
```diff
@@ -28,6 +28,7 @@ package de.fraunhofer.aisec.cpg.frontends.jvm
 import de.fraunhofer.aisec.cpg.frontends.Handler
 import de.fraunhofer.aisec.cpg.graph.*
 import de.fraunhofer.aisec.cpg.graph.declarations.*
+import de.fraunhofer.aisec.cpg.helpers.map
```
KuechA: unnecessary?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
oxisto: the standard lib one maps everything to a List :(

cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/DeclarationHandler.kt
oxisto: its using the one above

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
KuechA: I see... And what prevents us from calling `toSet()` on this? It's a bit longer but you ensure you get a Set as well. Also, I guess that we will just forget the import oftentimes

cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/DeclarationHandler.kt
KuechA: Yes, I got that already but it's the only place in the whole codebase were we use it, see suggestion for another "fix" above.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/helpers/Util.kt
oxisto: It’s basically faster and uses less memory because it will map directly into a Set, whereas in the other case we will map into a new list, then iterate that list into a set. So we are iterating over the values twice and we are allocating another list. Unless the compiler optimizes all of that away magically.

oxisto: See https://github.com/Fraunhofer-AISEC/cpg/issues/2567

oxisto: I'll leave this as map().toSet() and will tackle this in a separate PR

PR #2564, opened 2026-02-12
Better comparison of property edge list

oxisto:
This PR does a cleanup of comparing property edge lists, mainly getting rid of the `propertyEqualsList` functions and directly using their logic in regular equals.

### DIFF
--- a/cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/NodeTest.kt
+++ b/cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/NodeTest.kt
@@ -125,4 +125,56 @@ class NodeTest {
             block1.statementEdges == block2.statementEdges
         }
     }
+
+    @Test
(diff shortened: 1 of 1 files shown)


cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/NodeTest.kt
```diff
@@ -39,4 +39,142 @@ class NodeTest {
...
+            block1.statements += block2
+
+            block2.statements += stmt1
+            block2.statements += block1
+
+            // Test that comparison completes without stack overflow
+            // If Edge.equals used structural equality instead of reference equality,
+            // this would cause infinite recursion: block1 -> block2 -> block1 -> ...
+            // This should complete without throwing StackOverflowError
+            block1.statementEdges == block2.statementEdges
```
KuechA, line 125: Should there be an `assertFalse`?

PR #2392, opened 2025-07-17
Use the new OGM to export cpg as json

peckto:
Transition from the Neo4j OGM to the cpg internal OGM to persist cpg as json, analog to the Neo4j persistence.

One problem I'm currently facing with this approach is, that in some cases nodes with duplicate id (uuid) occur.
Not sure if those nodes are actually the same, or if there is a collision with the uuid.

Example:
https://github.com/pq-crystals/dilithium
```json
    {
      "id": "ffffffff-ad71-abff-ffff-ffffcb38da74",
      "labels": [
        "Node",
        "Declaration",
        "ValueDeclaration",
        "VariableDeclaration"
      ],
      "properties": {
        "isArray": false,
        "isGlobal": true,
        "isImplicitInitializerAllowed": false,
        "argumentIndex": 0,
        "isImplicit": false,
        "isInferred": true,
        "fullName": "pqcrystals_dilithium2_ref_idxlut",
        "localName": "pqcrystals_dilithium2_ref_idxlut",
        "nameDelimiter": "::",
        "name": "pqcrystals_dilithium2_ref_idxlut"
      }
    },
    {
      "id": "ffffffff-ad71-abff-ffff-ffffcb38da74",
      "labels": [
        "Node",
        "Declaration",
        "ValueDeclaration",
        "VariableDeclaration"
      ],
      "properties": {
        "isArray": false,
        "isGlobal": true,
        "isImplicitInitializerAllowed": false,
        "argumentIndex": 0,
        "isImplicit": false,
        "isInferred": true,
        "fullName": "pqcrystals_dilithium2_ref_idxlut",
        "localName": "pqcrystals_dilithium2_ref_idxlut",
        "nameDelimiter": "::",
        "name": "pqcrystals_dilithium2_ref_idxlut"
      }
    },
    {
      "id": "ffffffff-ad71-abff-ffff-ffffcb38da74",
      "labels": [
        "Node",
        "Declaration",
        "ValueDeclaration",
        "VariableDeclaration"
      ],
      "properties": {
        "isArray": false,
        "isGlobal": true,
        "isImplicitInitializerAllowed": false,
        "argumentIndex": 0,
        "isImplicit": false,
        "isInferred": true,
        "fullName": "pqcrystals_dilithium2_ref_idxlut",
        "localName": "pqcrystals_dilithium2_ref_idxlut",
        "nameDelimiter": "::",
        "name": "pqcrystals_dilithium2_ref_idxlut"
      }
    },
[...]
```

PR #2565, opened 2026-02-12
Trying to get rid of Neo4J OGM

oxisto:
This PR **finally** removes the Neo4J OGM (as follow up to #2392). It converts the schema generator to use our new persistence API and removes all usage of Neo4J OGM annotations.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,3 +1,12 @@
+# This workflow runs on every push.
+# Besides the actual build, it contains several additional steps:
+#
+# - Run unit, integration and performance tests
+# - Check code formatting (spotless)
+# - Generate and upload code coverage reports (Kover + Codecov)
+# - Publish artifacts to Maven Central (snapshots)
+# - Create a GitHub release (on tags)
+# - Generate and publish JavaDoc via Dokka (currently disabled)
 name: build
 
 on:
@@ -67,7 +76,8 @@ jobs:
         uses: gradle/actions/setup-gradle@v5
       - name: Build ${{ env.version }}
         run: |
-          ./gradlew --parallel spotlessCheck -x spotlessApply build -x distZip -x distTar koverXmlReport koverHtmlReport performanceTest integrationTest
+          # The distZip/distTar and shadowJar/shadowDistZip/shadowDistTar slow down the build and are not needed in CI
+          ./gradlew --parallel spotlessCheck -x spotlessApply build -x distZip -x distTar -x shadowJar -x shadowDistZip -x shadowDistTar koverXmlReport koverHtmlReport performanceTest integrationTest
         id: build
         env:
           ORG_GRADLE_PROJECT_version: ${{ env.version }}
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,143 @@
+# AGENTS.md
+
+This file provides guidance to AI coding agents (Claude Code, Cursor, Codex, Gemini CLI, etc.) working in this repository.
+
+## Project Overview
+
+**CPG** is a code property graph library and analysis platform.
+
+Key modules:
+- **cpg-core** – Core library: AST nodes, graph structures, passes, type system
+- **cpg-analysis** – Higher-level analyses built on top of cpg-core (dataflow, control flow, call graphs, etc.)
+- **cpg-language-\*** – Language frontends, one module per language (e.g., `cpg-language-go`, `cpg-language-python`)
+- **cpg-concepts** – Concept and operation definitions
+- **cpg-mcp** – MCP server exposing CPG analysis tools (dataflow, symbol analysis, concept application) to LLMs via streamable HTTP
+- **codyze-console** – Web-based analysis UI with AI agent chat (see [Architecture](#codyze-console-architecture) below)
+
+## Technology Stack
+
+### Backend
+- **Language**: Kotlin (requires Java 21)
+- **Build**: Gradle with Kotlin DSL
+- **Testing**: JUnit 5, kotlin.test
+- **Formatting**: Google Java Style via spotless plugin
+
+### Frontend (`codyze-console/src/main/webapp`)
+- **Framework**: Svelte 5 with SvelteKit
+- **Styling**: Tailwind CSS
+- **Package Manager**: pnpm (not npm)
+
+## Development Commands
+
+### Build
+
+```bash
+# Full build (format + test + publish locally)
+./gradlew clean spotlessApply build publishToMavenLocal
+
+# Quick build
+./gradlew build
+
+# Build a single module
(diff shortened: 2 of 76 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/converters/NameConverter.kt
```diff
@@ -76,7 +76,10 @@ class NameConverter : CpgCompositeConverter<Name?> {
                 Pair("String", FIELD_NAME_DELIMITER),
             )
 
-    override fun toEntityAttribute(value: MutableMap<String, *>): Name {
-        return parseName(value[FIELD_FULL_NAME].toString(), value[FIELD_NAME_DELIMITER].toString())
+    override fun toEntityAttribute(value: Map<String, *>?): Name {
+        return parseName(
+            value?.get(FIELD_FULL_NAME).toString(),
```
KuechA: If `value` is `null`, this will create the Name `null.null`. Is this really intended? In `SimpleNameConverter`, we set it to an empty string.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/converters/TranslationStatsConverter.kt
```diff
@@ -55,9 +55,9 @@ class TranslationStatsConverter : CpgCompositeConverter<TranslationStats?> {
     override val graphSchema: List<Pair<String, String>>
         get() = listOf(Pair("String", FIELD_TOTAL_LINES_OF_CODE))
 
-    override fun toEntityAttribute(value: MutableMap<String, *>): TranslationStats {
+    override fun toEntityAttribute(value: Map<String, *>?): TranslationStats {
         return TranslationStats().also {
-            it.totalLinesOfCode = Integer.parseInt(value[FIELD_TOTAL_LINES_OF_CODE] as String)
+            it.totalLinesOfCode = Integer.parseInt(value?.get(FIELD_TOTAL_LINES_OF_CODE) as String)
```
KuechA: This cast could fail. We should call `toString()` and have a common logic for `value` being `null` - this is the third different variant (and so far the worst) I've seen until now (of 3!).

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/Common.kt
```diff
@@ -49,11 +50,66 @@ import kotlin.reflect.full.withNullability
...
+    /** Converts a graph database value back to a property value. */
+    fun toEntityAttribute(value: G?): T
+}
+
+/**
+ * Interface for converters that convert an object property to multiple graph database properties.
+ *
+ * @param T The object property type
+ */
+interface CompositeAttributeConverter<T> {
```
KuechA: I'm not sure if this interface is really necessary. The same could be modeled with the normal `AttributeConverter`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/Common.kt
```diff
@@ -49,11 +50,66 @@ import kotlin.reflect.full.withNullability
...
+}
+
+/**
+ * Annotation to specify a converter for a property.
+ *
+ * @param value The converter class to use for this property.
+ */
+@Target(AnnotationTarget.FIELD, AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Convert(val value: KClass<out Any>)
```
KuechA: `KClass<Any>` is very generic. I'd prefer to ensure that it's somehow a converter class from above.

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Schema.kt
```diff
@@ -137,70 +145,101 @@ class Schema {
...
+     * Extracts information on the nodes and edges that can be persisted to the neo4 database using
+     * the new persistence schema.
      */
     fun extractSchema() {
-        val meta = MetaData(Node.javaClass.packageName)
-        val nodeClassInfo =
-            meta.persistentEntities().first { it.underlyingClass == Node::class.java }
+        // Use ClassGraph to find all Node subclasses
+        val scanResult =
+            ClassGraph().acceptPackages(Node::class.java.packageName).enableClassInfo().scan()
```
KuechA, line 159: I think we have some logic like this in other packages, e.g. for overlays. We should have a common one instead of a bunch of different ways to identify subclasses of some node class.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/converters/NameConverter.kt
konradweiss: Solved with default values if null

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/converters/TranslationStatsConverter.kt
konradweiss: Solved with default values if null

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/Common.kt
konradweiss: I kept this but expressed it as a subinterface of AttributeConverter

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/persistence/Common.kt
konradweiss: Restricted to KClass<out AttributeConverter<*, *>> because we now have it as common root attribute converter

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Schema.kt
konradweiss: This is not just for identifying subclasses of Node, but to then further get information on relationships etc. But maybe I am just not understanding what exactly you mean. Can you pinpoint me to where we do the same/something overlapping?

oxisto: I think it does not exist, because I had to import ClassGraph as a new concept, the only implementation that existed before was a method from the old OGM

PR #2566, opened 2026-02-13
Adding ExtensionDeclaration

konradweiss:
The ExtensionDeclaration is an AST container that holds additional declarations that are not added as a subdeclarations to any current AST element. Instead, the frontend can parse additional information from the parser AST element to determine which scope to retrieve and add the contained declarations too.

Examples of such containers are the `impl` block in Rust or the partial classes in C#. The main purpose of not just adding the contained declarations to the AST of the record it extends is to maintain a clean AST.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/ExtensionDeclaration.kt
```diff
@@ -0,0 +1,82 @@
...
+     * namespace.
+     */
+    val declarationEdges = astEdgesOf<Declaration>()
+    override val declarations by unwrapping(ExtensionDeclaration::declarationEdges)
+
+    /**
+     * In some languages, there is a relationship between paths / directories and the package
+     * structure. Therefore, we need to be aware of the path this namespace / package is in.
+     */
+    var path: String? = null
```
oxisto: I think this is a left over from copying it from the namespace declaration. Should be removed

PR #2586, opened 2026-02-19
Add some language traits to JVM language

KuechA:


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/TypeResolver.kt
```diff
@@ -61,6 +61,9 @@ open class TypeResolver(ctx: TranslationContext) : ComponentPass(ctx) {
         walker = SubgraphWalker.ScopedWalker(scopeManager, strategy = Strategy::AST_FORWARD)
         walker.registerHandler { handleNode(it) }
         walker.iterate(component)
+
+        // Hacky chackykc hacky
+        typeManager.lookAlsoAtThis.forEach { handleType(it) }
```
oxisto: I think this MAY require a better solution 🤣

PR #2594, opened 2026-02-20
Remove parallel pass processing

oxisto:
This PR removes parallel pass processing as this was never really fully functional and it restricted us to use concurrent hash maps for certain things (such as the clean/dirty system) which were incredibly slow.

The use-case for parallel pass processing is very niche, as its usually the case that passes modify certain nodes and therefore parallel processing is seldom "safe" to do.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
@@ -27,7 +27,9 @@
 
 package de.fraunhofer.aisec.cpg.graph.builder
 
-import de.fraunhofer.aisec.cpg.*
+import de.fraunhofer.aisec.cpg.ScopeManager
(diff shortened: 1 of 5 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/builder/Fluent.kt
```diff
@@ -37,16 +37,33 @@ import de.fraunhofer.aisec.cpg.graph.declarations.Function
...
+import kotlin.Suppress
+import kotlin.Unit
+import kotlin.apply
+import kotlin.collections.List
+import kotlin.collections.filter
+import kotlin.collections.firstOrNull
+import kotlin.collections.forEach
+import kotlin.collections.listOf
+import kotlin.collections.maxOfOrNull
+import kotlin.collections.mutableListOf
```
KuechA: Why do we need all these imports if all we do in this file is remove stuff?

PR #2598, opened 2026-02-23
Use simple content-based file type identification im JVM frontend

KuechA:
Currently, the JVM frontend relies on file names. However, we have use-cases where we do not have any file ending. To handle this, we could just check what is included in the file and identify jars or apk files based on the contents in a zip file. This is a very simple heuristics but should usually work.

Note: A file without the ending apk/jar won't be identified by the `JVMLanguage` as this would require overriding the `handlesFile` method. #2597 enables applications using the CPG as library to do this.


cpg-language-jvm/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/JVMLanguageFrontendTest.kt
```diff
@@ -224,6 +289,7 @@ class JVMLanguageFrontendTest {
                 )
             }
         assertNotNull(tu)
+        assertNotNull(tu)
```
lshala: duplicated assert

lshala:
```suggestion
```

lshala (approved):
LGTM, but it would be nice to have a test for files not ending with .apk/.jar

PR #2600, opened 2026-02-23
Improve inference (for JVM frontend)

KuechA:
1. We account for package names which are nested structures, i.e., we have to split them up and go in/out of the namespaces properly.
2. We have types which may not be declared and are not part of the AST, this means they will be inferred but without being part of the AST, this will happen in the SymbolResolver and less performant. We therefore already add such types in the frontend, so that the `TypeResolver` can start working on them

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Component.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Component.kt
@@ -72,7 +72,7 @@ open class Component : AstNode() {
      * Stores types that are not directly connected to a node, but are still relevant for the type
      * resolution of this component.
      */
-    val lookAlsoAtThis = IdentityHashMap<String, Type>()
+    val additionalTypes = IdentityHashMap<String, Type>()
 
     /**
      * In contrast to other Nodes we do not add the assumptions collected over the component because
--- a/cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/ExpressionHandler.kt
+++ b/cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/ExpressionHandler.kt
@@ -377,7 +377,10 @@ class ExpressionHandler(frontend: JVMLanguageFrontend) :
         val currentComponent =
             this@ExpressionHandler.frontend.currentTU?.firstParentOrNull<Component>()
 
-        currentComponent?.lookAlsoAtThis?.computeIfAbsent(parentTypeName) {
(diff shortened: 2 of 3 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TypeManager.kt
```diff
@@ -66,6 +67,12 @@ class TypeManager {
      */
     val resolvedTypes = identitySetOf<Type>()
 
+    /**
+     * Stores types that are not directly connected to a node, but are still relevant for the type
+     * resolution of a component.
+     */
+    val lookAlsoAtThis = IdentityHashMap<Component, IdentityHashMap<String, Type>>()
```
oxisto: I think you suggest moving this into `Component`, that might make a good idea.

KuechA: done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Component.kt
```diff
@@ -66,6 +68,12 @@ open class Component : AstNode() {
         translationUnits.add(tu)
     }
 
+    /**
+     * Stores types that are not directly connected to a node, but are still relevant for the type
+     * resolution of this component.
+     */
+    val lookAlsoAtThis = IdentityHashMap<String, Type>()
```
oxisto: name it `additionalTypes`

cpg-language-jvm/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/ExpressionHandler.kt
```diff
@@ -372,7 +372,16 @@ class ExpressionHandler(frontend: JVMLanguageFrontend) :
 
     private fun SootClassMemberSignature<*>.toStaticRef(): Reference {
         // First, construct the name using <parent-type>.<fun>
-        val ref = newReference("${this.declClassType.fullyQualifiedName}.${this.name}")
+        val parentTypeName = "${this.declClassType.fullyQualifiedName}"
+        val ref = newReference("$parentTypeName.${this.name}")
+        val currentComponent =
+            this@ExpressionHandler.frontend.currentTU?.firstParentOrNull<Component>()
+
+        currentComponent?.lookAlsoAtThis?.computeIfAbsent(parentTypeName) {
```
oxisto: maybe add a commetn saying that we want to make the parent "type" available in the type manager, so that we can deduce inference later correctly

PR #2602, opened 2026-02-23
Python: Fix self-assigned fields and or-expression type propagation

lshala:
Two related fixes:
- Instance/self-assigned fields (e.g. self.foo = Foo()) were not being declared because `self` has an `UnknownType`. This prevented calls on those fields from being resolved.
- There are also cases where the type can either be passed via a parameter or it will be declared with the default constructor (e.g. self.foo = foo or Foo())

Partially fixes #2575

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
@@ -41,6 +41,7 @@ import de.fraunhofer.aisec.cpg.graph.ContextProvider
 import de.fraunhofer.aisec.cpg.graph.Name
 import de.fraunhofer.aisec.cpg.graph.Node
 import de.fraunhofer.aisec.cpg.graph.OverlayNode
+import de.fraunhofer.aisec.cpg.graph.builder.assign
 import de.fraunhofer.aisec.cpg.graph.component
 import de.fraunhofer.aisec.cpg.graph.declarations.Declaration
 import de.fraunhofer.aisec.cpg.graph.declarations.Function
@@ -274,6 +275,9 @@ abstract class Language<T : LanguageFrontend<*, *>>() : Node() {
      * or the right operand can be the result at runtime), or an empty set if no special handling is
      * needed.
      *
+     * If both [lhs] and [rhs] have only one assigned type, the type will be determined through
+     * [propagateTypeOfBinaryOperation].
+     *
      * Languages can override this to handle language-specific operators.
      */
     open fun propagateAssignedTypesOfBinaryOperation(
@@ -281,6 +285,19 @@ abstract class Language<T : LanguageFrontend<*, *>>() : Node() {
         lhs: Expression,
         rhs: Expression,
     ): Set<Type> {
+        val lhsAssigned = lhs.assignedTypes
+        val rhsAssigned = rhs.assignedTypes
+        if (lhsAssigned.size == 1 && rhsAssigned.size == 1) {
+            val typeResult =
+                propagateTypeOfBinaryOperation(
+                    operatorCode,
+                    lhsAssigned.single(),
+                    rhsAssigned.single(),
+                )
+            if (typeResult !is UnknownType) {
+                return setOf(typeResult)
+            }
+        }
         return emptySet()
     }
 


cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
```diff
@@ -248,6 +248,11 @@ open class PythonLanguage :
                     primitiveType("float")
                 }
             }
+            "or" -> {
+                // In Python, `a or b` returns `a` if truthy, else `b`. If the lhs type is
+                // unknown, the fallback is the rhs type.
+                return if (lhsType is UnknownType) rhsType else lhsType
```
maximiliankaul: I don't think this is how we should implement this. `lhsType` being `UnknownType` doesn't indicate anything regarding the boolean value of `lhs`. My suggestion would be:
- add a reference to https://docs.python.org/3/reference/expressions.html#boolean-operations
- also implement `and` and `not`
- check whether the value of `lhs` is `False`. Then use `rhsType`. (However, this looks difficult here as we don't have access to `lhs` only to `lhsType` 🤔

maximiliankaul: As I'm not supposed to "Request changes" and the PR is now ready: I think this should be addressed before merging!

cpg-language-python/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/python/PythonLanguage.kt
```diff
@@ -248,10 +249,25 @@ open class PythonLanguage :
...
+             * side is dynamic, `DynamicType` is returned, and otherwise fall back to
+             * `unknownType()`.
+             *
+             * See https://docs.python.org/3/reference/expressions.html#boolean-operations
+             * *
+             */
+            "or",
+            "and" -> {
+                return when {
+                    lhsType == rhsType -> lhsType
```
maximiliankaul: I like the lhs == rhs check. I don't get the `DynamicType` check. We should probably introduce a union of the two possible types?

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/SymbolResolverTest.kt
```diff
@@ -125,4 +128,64 @@ class SymbolResolverTest {
...
+        assertNotNull(clientClass)
+
+        val serviceClass = result.records["Service"]
+        assertNotNull(serviceClass)
+
+        val clientField = serviceClass.fields["client"]
+        assertIs<Field>(clientField)
+
+        val fieldType = clientField.type
+        assertIs<ObjectType>(fieldType)
```
maximiliankaul, line 150: Shouldn't this then check for `Client`?

lshala: Done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -267,6 +268,22 @@ abstract class Language<T : LanguageFrontend<*, *>>() : Node() {
...
+     * needed.
+     *
+     * Languages can override this to handle language-specific operators.
+     */
+    open fun propagateAssignedTypesOfBinaryOperation(
+        operatorCode: String?,
+        lhs: Expression,
+        rhs: Expression,
+    ): Set<Type> {
+        return emptySet()
```
KuechA, line 301: I'm not sure if that's a good default implementation. At least if `lhs` and `rhs` have a single assigned type, we could use the same logic as for the `type` field.

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
lshala: Please double check again if that's what you had in mind

KuechA: yes, I think that's better

PR #2603, opened 2026-02-24
Upgrade to mcp kotlin-sdk version 0.8.4 and provide http server

KuechA:


cpg-mcp/src/main/kotlin/de/fraunhofer/aisec/cpg/mcp/Application.kt
```diff
@@ -85,3 +104,12 @@ fun runMcpServerUsingStdio() {
 fun runSseMcpServerUsingKtorPlugin(port: Int, server: Server) = runBlocking {
     embeddedServer(CIO, host = "0.0.0.0", port = port) { mcp { server } }.start(wait = true)
 }
+
+fun runHttpMcpServerUsingKtorPlugin(port: Int, server: Server) {
+    runBlocking {
+        embeddedServer(factory = CIO, host = "0.0.0.0", port = port) {
```
maximiliankaul: I know this is the same as the `sse` option, but I'd be in favor to restrict this to `127.0.0.1`

PR #2623, opened 2026-03-10
Upgrades the Kotlin MCP SDK from 0.8.4 to 0.9.0 

lshala:
Since version 0.9.0 (#2620), there is a breaking change regarding the "handler" function, which we use in almost every test. It now accepts a `ClientConnection` as receiver, which for our tests would mean creating a dummy class with some boilerplate code.
Instead, the new `kotlin-sdk-testing` module (see [test example](https://github.com/modelcontextprotocol/kotlin-sdk/blob/5b58814613069babb1bbd5cab4572331eaed4b45/kotlin-sdk-testing/Module.md)) is used, which provides a way to test the client-server functionalities.

Closes #2620


KuechA:
I guess the changes are ok but the tests don't run

PR #2643, opened 2026-03-20
Add functionality for callbacks

KuechA:
This PR adds a way to register callbacks that are called when `Pass`es or `LanguageFrontend`s have terminated. It's useful to provide the user with some progress. It extends `consumeTargets`, `analyze`, and `executePass` with an additional argument for the callback.


cpg-ai/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/RunPassForNodeTest.kt
```diff
@@ -42,6 +44,36 @@ import kotlin.test.assertTrue
 
 class RunPassForNodeTest {
 
+    @Test
+    fun runEOGPass() {
```
maximiliankaul, line 48: I don't get how this tests something with callbacks (as in the PR description) or the EOG as the test name suggests :shrug: Other than that, I'm ok with this test ;)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationProgressCallback.kt
```diff
@@ -0,0 +1,53 @@
...
+    /** Called after all configured frontends finished running. */
+    fun afterFrontends(
+        ctx: TranslationContext,
+        result: TranslationResult,
+        executedFrontends: Set<LanguageFrontend<*, *>>,
+    ) {}
+
+    /** Called after a pass has finished execution. */
+    fun afterPass(
+        pass: KClass<out Pass<out Node>>,
```
maximiliankaul, line 64: Can you write some doc regarding the parameters, please? For example: Is the pass a filter to only trigger the cb on a specific task?

cpg-ai/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/RunPassForNodeTest.kt
```diff
@@ -79,7 +118,14 @@ class RunPassForNodeTest {
         assertNotNull(ctx)
 
         // Use a concrete TU-related pass on the parent
-        val result = runPassForNode(globalAnalysisResult, ResolveMemberAmbiguityPass::class, ctx)
+        val result =
+            runPassForNode(
+                nodeToPass,
+                globalAnalysisResult,
```
maximiliankaul, line 124: globalAnalysisResult twice is intended (also below)?

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/TranslationProgressCallbackTest.kt
```diff
@@ -0,0 +1,197 @@
...
+    override fun accept(t: TranslationResult) {
+        executionCount++
+    }
+
+    companion object {
+        var executionCount: Int = 0
+    }
+}
+
+class TranslationProgressCallbackTest {
```
maximiliankaul, line 91: Maybe also a test with more than 1 callback :thinking: ?

maximiliankaul (approved):
LGTM

cpg-ai/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/RunPassForNodeTest.kt
KuechA: I think this was mostly to reach sufficient code coverage. It's not related to the main changes but mostly to the changed signature in the MCP tool

cpg-ai/src/integrationTest/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/RunPassForNodeTest.kt
KuechA: It is indeed intended. I have to pass two nodes: The `TranslationResult` and the node which I want to analyze (or from which `ruinPassForNode` searches for the node to feed into the `Pass`. In this case, the test tested the technology to identify the correct node if we pass the AST-parent (which was the behavior before already, I just had to add this map and the additional translationResult which is now required for the analysis).

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationProgressCallback.kt
KuechA: it's the pass which has just terminated. Added documentation for everything in the class

PR #2661, opened 2026-03-27
Update CONTRIBUTING.md

lshala:
A few outdated parts in the `Contributing.md` file were identified in #2646 and will be updated in this PR


CONTRIBUTING.md
```diff
@@ -73,19 +73,18 @@ Most of our code is written in Kotlin and if you develop new nodes, one should f
...
-var parameters by PropertyEdgeDelegate(FunctionDeclaration::parameterEdges)
+var parameters by unwrapping(Function::parameterEdges)
 ```
 
-Note: We actually want list property to be immutable so that they can only be modified by the node class itself. However, it is currently not possible to have them immutable on the public getter, but mutable for the class itself. There is a Kotlin issue tracking this feature request. Once https://youtrack.jetbrains.com/issue/KT-14663 is resolved, we should set the public type for all those lists to `List` instead of `MutableList`. Properties delegated by `PropertyEdgeDelegate` are already immutable.
+Note: We actually want list property to be immutable so that they can only be modified by the node class itself. However, it is currently not possible to have them immutable on the public getter, but mutable for the class itself. There is a Kotlin issue tracking this feature request. Once https://youtrack.jetbrains.com/issue/KT-14663 is resolved, we should set the public type for all those lists to `List` instead of `MutableList`. Properties delegated by `unwrapping` are already immutable.
```
lshala, line 88: @konradweiss As discussed, I made a draft change. Please double check this and possibly any other outdated parts which I might have missed

Issue #2732, opened 2026-04-21
SymbolResolver too strict language comparison


KuechA:
The SymbolResolver removes symbols from a list of candidates if the languages do not equal. However, we support languages which are not equal but yet symbols can be reused e.g. I can define a function in C and use it in CPP. This leads to cases where the symbol resolution does not work. The problem is in lines 202 - 205:
```kotlin
// Filter according to the language
if (languageOnly != null) {
    list.removeIf { it.language != languageOnly }
}
```

oxisto:
Yeah it should probably be extended to !it.language.isAssignedFrom(languageOnly) or something like that.

oxisto:
```patch
diff --git a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
index eb74ca4001d..52ecb8caa89 100644
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/scopes/Scope.kt
@@ -201,7 +201,10 @@ sealed class Scope(
 
             // Filter according to the language
             if (languageOnly != null) {
-                list.removeIf { it.language != languageOnly }
+                list.removeIf {
+                    !languageOnly::class.java.isInstance(it.language) &&
+                            !it.language::class.java.isInstance(languageOnly)
+                }
             }
 
             // Filter the list according to the predicate, if we have any
```

PR #2734, opened 2026-04-22
A very simple way to specify interactions between two programming languages

KuechA:
In case of incomplete code, it is possible that a function is declared in some language and then used in another programming language through some sort of interface between the languages (could be default e.g. for C/C++ or a language bridge mapping more things e.g. JNI). This is not really handled by the SymbolResolver and it would just infer new nodes if the language is not equal.

This PR suggests introducing a very basic model specifying such inter-language bridges.

Related to #2732


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageInterface.kt
```diff
@@ -0,0 +1,46 @@
...
+ *                   \$$$$$   |$$ |      \$$$$$   |
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.frontends
+
+import de.fraunhofer.aisec.cpg.graph.scopes.Symbol
+import de.fraunhofer.aisec.cpg.graph.types.Type
+
+abstract class LanguageInterface<FROM : Language<*>, TO : Language<*>>(val from: FROM, val to: TO) {
```
oxisto: Maybe something like ForeignFunctionInterface? At least that seems to be the regular term here.

oxisto: Or maybe LanguageMapping? Also needs a kdoc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -811,7 +811,43 @@ class ScopeManager(override var ctx: TranslationContext) : ScopeProvider, Contex
                         )
                         ?.toMutableList() ?: mutableListOf()
                 }
-            }
+            }.toMutableList()
+
+        val otherLanguageInterfaces = ctx.languageInterfaces[language]
```
oxisto, line 799: can we extract this into a separate function?

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CAndCxxMapper.kt
```diff
@@ -0,0 +1,50 @@
...
+ *                    \______/ \__|       \______/
+ *
+ */
+package de.fraunhofer.aisec.cpg.frontends.cxx
+
+import de.fraunhofer.aisec.cpg.frontends.LanguageInterface
+import de.fraunhofer.aisec.cpg.graph.scopes.Symbol
+import de.fraunhofer.aisec.cpg.graph.types.Type
+
+class CToCxxMapper : LanguageInterface<CLanguage, CPPLanguage>(CLanguage(), CPPLanguage()) {
```
oxisto: Ok this is complete overkill. The original idea is a great idea but what you are trying to achieve can be solved by a simple adjustment of the language check (see original issue)

PR #2756, opened 2026-05-20
Consider derefed return values

morbitzer:
The PtP did not yet consider the derefed values of return values. Now it does.


cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/PointsToPassTest.kt
```diff
@@ -4581,4 +4581,63 @@ class PointsToPassTest {
...
+
+        // We expect the FS to contain one key. This key has 2 values, the malloc at depth 0 and the
+        // hardcoded UMV freedMemory at
+        // depth 1
+        assertEquals(1, testFuncFS.entries.size)
+        assertEquals(2, testFuncFS.entries.single().value.size)
+        assertTrue(
+            testFuncFS.entries.single().value.any {
+                it.destValueDepth == 1 &&
+                    (it.srcNode as? UnknownMemoryValue)?.name?.localName == "malloc"
```
oxisto: there is an `assertLocalName` that might be a bit easier to handle than a lot of ?

cpg-language-cxx/src/test/resources/pointsToPass/returnfree.cpp
```diff
@@ -0,0 +1,15 @@
+char* returnfreed() {
+  char* ret = malloc(16);
+  
+  free(ret);
+  
+  return ret;
+}
+
+int main() {
+  char* str = NULL;
```
oxisto: C++ technically doesnt have `NULL`, but the `nullptr` keyword. Code probably could also be C instead of C++.

cpg-language-cxx/src/test/resources/pointsToPass/returnfree.cpp
```diff
@@ -0,0 +1,15 @@
+char* returnfreed() {
```
oxisto: maybe in general it would be nice to actually have these snippets compilable, i.e. include proper header (stdio.h and stdlib.h in this case), just so we can also analyze them with other tools (clang/LLVM) and compare, etc.

oxisto (approved):
Minor nitpicks

cpg-language-cxx/src/test/resources/pointsToPass/returnfree.cpp
morbitzer: Is now C

oxisto (approved):

cpg-language-cxx/src/test/resources/pointsToPass/returnfree.cpp
morbitzer: done

PR #2757, opened 2026-05-20
Demo

oxisto:
Might be split into smaller PRs

### DIFF
--- a/codyze-console/src/main/kotlin/de/fraunhofer/aisec/codyze/console/ConsoleService.kt
+++ b/codyze-console/src/main/kotlin/de/fraunhofer/aisec/codyze/console/ConsoleService.kt
@@ -39,7 +39,13 @@ import de.fraunhofer.aisec.cpg.graph.nodes
 import de.fraunhofer.aisec.cpg.passes.concepts.LoadPersistedConcepts
 import de.fraunhofer.aisec.cpg.passes.concepts.LoadPersistedConcepts.PersistedConceptEntry
 import de.fraunhofer.aisec.cpg.passes.concepts.LoadPersistedConcepts.PersistedConcepts
+import de.fraunhofer.aisec.cpg.passes.concepts.config.ProvideConfigPass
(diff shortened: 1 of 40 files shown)


codyze/build.gradle.kts
```diff
@@ -31,7 +31,17 @@ plugins {
...
+application {
+    mainClass.set("de.fraunhofer.aisec.codyze.ApplicationKt")
+    // Suppress two unavoidable JVM warnings from transitive deps:
+    //   * sun.misc.Unsafe deprecation — emitted by kotlin-compiler-embeddable's
+    //     internal IntelliJ-platform copy on every scripting eval.
+    //   * Restricted-method warning — emitted by JNA when loading native libs.
+    // Both are out of our control. We pin the flags here so the generated start
+    // scripts (codyze, codyze.bat) inherit them.
+    applicationDefaultJvmArgs =
+        listOf("--sun-misc-unsafe-memory-access=allow", "--enable-native-access=ALL-UNNAMED")
```
peckto, line 43: The option `--sun-misc-unsafe-memory-access=allow` fails on my java openjdk with "unknown option".
Can we move this to `gradle.properties` if needed?

codyze-console/src/main/kotlin/de/fraunhofer/aisec/codyze/console/ConsoleService.kt
peckto, line 121: It would be nice to also have support for json compilation database and unity-build option.
Not sure how we can do this in a usable/smart way when loading code via the codyze console.
No sure if unity-build mode is a no-op for languages other than CXX.

peckto (commented):
Thanks for the console improvements and the documented Demo!
Two things I came along when playing with it.

codyze/build.gradle.kts
oxisto: Its not needed anymore actually, that sentence can be removed

PR #2766, opened 2026-05-26
ArraySizeEvaluator: branch-merged sizes, value-equal lub

oxisto:
- `ArrayValue.getSize`: replace the prior `TODO()` for unhandled shapes with `LatticeInterval.BOTTOM`, and guard `ArrayConstruction` against empty `dimensions` and non-Number dimension expressions (used to crash with `NoSuchElementException` / `ClassCastException`).
- `ArrayValue.applyEffect`: anchor `Assign`s on the underlying `Variable` rather than the LHS `Reference`, so writes from different branches share one key in general state; skip the trailing `TOP` push for unrelated nodes (lubbing `TOP` into general state would clamp later refinements right back to `TOP`).
- `DeclarationState.lub`: merge keys by `equals` instead of relying on the inherited `IdentityHashMap`. Without this, two value-equal `Integer` keys produced by `Node.objectIdentifier()` in different branches survive a naive lub as separate entries, breaking branch convergence. Routes through `DeclarationStateElement`'s existing value-aware `get`/`put` overrides.
- Integration test (`cpg-language-cxx`) covering fixed buffers, `malloc` with a constant, string-literal initializer, and branch-merged allocations.
- Unit tests (`cpg-analysis`) covering the safer `getSize` paths: empty `dimensions` and non-`Number` dimension expressions.

### DIFF
--- a/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/abstracteval/AbstractIntervalEvaluator.kt
+++ b/cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/abstracteval/AbstractIntervalEvaluator.kt
@@ -95,16 +95,17 @@ class DeclarationState<NodeId>(innerLattice: Lattice<NewIntervalLattice.Element>
         get() = DeclarationStateElement()
 
     /**
-     * Merges two declaration states by [Lattice.lub] on the inner lattice, but matches keys by
-     * `equals` rather than reference identity.
+     * Merges two declaration states by [Lattice.lub] on the inner lattice, going through
+     * [DeclarationStateElement]'s `get`/`put` overrides so equal-value [Int] keys collapse onto a
+     * single entry.
      *
      * Background: [MapLattice.Element] inherits from [IdentityHashMap], so the inherited [lub]
      * treats two value-equal-but-distinct keys as separate entries. Our keys are autoboxed `Int`s
      * from [Node.objectIdentifier]; two branches of an `if/else` produce equal but distinct
      * `Integer` instances, so a naive merge of `if(c) buf=malloc(16) else buf=malloc(64)` ends up
-     * with two `buf` entries instead of a single one holding `[16, 64]`. Iterating entries and
-     * looking up equal keys ourselves restores value-based merge semantics without changing the
-     * underlying lattice infrastructure.
+     * with two `buf` entries instead of a single one holding `[16, 64]`. The element's overridden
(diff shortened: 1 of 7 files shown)


cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/abstracteval/value/ArrayValue.kt
```diff
@@ -85,58 +86,65 @@ class ArrayValue : Value<LatticeInterval> {
...
+        return current
     }
 
-    private fun getSize(node: Node): LatticeInterval {
-        return when (node) {
+    private fun getSize(node: Node): LatticeInterval =
+        when (node) {
             is Literal<*> -> {
-                if (node.value is String) {
-                    // For strings, we return the length of the string
```
KuechA, line 105: I'd prefer to keep the comment because it's easier to understand what was the idea

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/abstracteval/AbstractIntervalEvaluator.kt
```diff
@@ -94,16 +94,35 @@ class DeclarationState<NodeId>(innerLattice: Lattice<NewIntervalLattice.Element>
     override val bottom: DeclarationStateElement<NodeId>
         get() = DeclarationStateElement()
 
+    /**
+     * Merges two declaration states by [Lattice.lub] on the inner lattice, going through
+     * [DeclarationStateElement]'s `get`/`put` overrides so equal-value [Int] keys collapse onto a
+     * single entry.
+     *
+     * Background: [MapLattice.Element] inherits from [IdentityHashMap], so the inherited [lub]
```
KuechA: I'm not sure if this PR really fixes the problem or merely hides some of the consequences. I think that the cleaner version would be to make a second version of the `MapLattice` which inherits from `MutableMap` for such cases. Will we actually get the right values out of the `IdentityHashMap` when using an equal but not identical int as key? I don't think so. Ideally, we could get rid of the `Int` as keys but I guess this has other problems?

oxisto: Added a new HashMapLattice

KuechA (approved):
Assuming that most changes since my last review are copy and paste, I'll approve without further review

PR #2767, opened 2026-05-27
Some speedup of the points to pass

KuechA:


morbitzer (approved):

oxisto (commented):
can we please put the `Analyzing function` log line back to trace or at least debug. we should not print that level of detail on INFO. Overall I have no idea what all these magic atomic values do but it seems to speed up by example from 4s to 3s, so that's good! Even less if one takes the logging from INFO to TRACE

PR #2774, opened 2026-06-01
CXX allocator concept pass + cpg-analysis/cpg-concepts dependency flip

oxisto:
## Summary

Replaces the `malloc` string-match in `ArrayValue.getSize` with a generic read of an `Allocate` memory-concept overlay. The actual recognition of allocator calls moves into a new language-specific pass, `CXXMemoryAllocationPass`, which lives next to the C/C++ frontend.

Along the way:

- **Module layering flipped**: `cpg-analysis` now depends on `cpg-concepts` instead of the other way round. The only coupling was one file (`PolicyQueries.kt`), which moved to its semantic home in cpg-analysis. The new chain is `cpg-language-cxx → cpg-concepts → cpg-analysis → cpg-core`.
- **Existing CXX concept passes relocated**: `CXXDynamicLoadingPass` and `CXXEntryPointsPass` moved from cpg-concepts into cpg-language-cxx. cpg-concepts now contains only language-agnostic concept definitions.
- **Auto-registration**: `CXXLanguageFrontend` carries `@RegisterExtraPass(CXXMemoryAllocationPass::class)` so the pass runs automatically with `defaultPasses()` whenever the C/C++ frontend is loaded — no more manual `it.registerPass<>()` for the common case.

## Commit walkthrough

1. **Flip cpg-analysis ↔ cpg-concepts** — `PolicyQueries.kt` → cpg-analysis; gradle deps swapped.
2. **`Allocate.size: Expression?`** — new field on the existing concept + builder support.
3. **Relocate CXX concept passes** — `CXXDynamicLoadingPass`, `CXXEntryPointsPass` → cpg-language-cxx. `ConceptPass.getConceptOrCreate` widened from `internal` → `protected`.
4. **`CXXMemoryAllocationPass`** — matches malloc/calloc/realloc; for calloc the size is a synthesised `count * elemSize` `BinaryOperator` (ValueEvaluator constant-folds when both operands are literals).
5. **Wire `ArrayValue.getSize`** to read `Allocate` overlays + auto-register the pass. `Memory` concept attaches to the `Component` (one heap per program, shared across TUs).
6. **Tests** — 6 integration tests covering malloc with a constant, malloc via Assign, calloc, realloc, malloc with a non-constant size, and the negative case (no overlay on unrelated calls).

## Why Expression for `size`?

A bare `Int` only handles `malloc(constant)`. Keeping the original `Expression` node means:
- Constant cases: callers do `(size.value.value as Number).toLong()`.
- Variable cases (e.g. `malloc(n)`): callers can run interval analysis, taint, DFG tracing, etc.
- The synthesised `count * elemSize` for calloc folds uniformly across literal-times-literal, variable-times-literal, etc.

## Not covered (one-line extensions, left for follow-ups)

- `new` / `new[]` — different AST node shape in CDT.
- `aligned_alloc`, `posix_memalign` — size lives in different argument positions.
- Python / other language allocator passes — same pattern applies; pass goes in the respective language module.

## Test plan

- [x] `:cpg-core:test` — green
- [x] `:cpg-analysis:test` — green
- [x] `:cpg-language-cxx:test` (including new `CXXMemoryAllocationPassTest` + existing `ArraySizeEvaluatorTest`, `TypedefTest`, etc.) — green
- [x] `:cpg-concepts:integrationTest` (including the relocated `DynamicLoadingTest`, `EntryPointTest`, `PolicyTest`) — green


oxisto:
Anyone care to review? @konradweiss @maximiliankaul

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/memory/cxx/CXXMemoryAllocationPass.kt
```diff
@@ -0,0 +1,146 @@
...
+        } ?: newMemory(component, mode = MemoryManagementMode.UNMANAGED, connect = true)
+    }
+
+    /**
+     * The expression that determines how many bytes the call allocates, or `null` if the call isn't
+     * a recognised allocator. For `calloc(M, N)` the total is the product `M * N`, so we synthesise
+     * a [BinaryOperator] (`*`) over the two arguments — `ValueEvaluator` will constant-fold it when
+     * both operands are literals.
+     */
+    private fun allocationSizeOf(call: Call): Expression? =
```
maximiliankaul, line 120: Maybe log an error over silently failing by returning `null`?

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/memory/cxx/CXXMemoryAllocationPass.kt
```diff
@@ -0,0 +1,146 @@
...
+@Description(
+    "Recognises C/C++ allocator calls (malloc, calloc, realloc) and attaches Allocate concepts."
+)
+class CXXMemoryAllocationPass(ctx: TranslationContext) : ConceptPass(ctx) {
+
+    override fun handleNode(node: Node, tu: TranslationUnit) {
+        if (node is Call) handleCall(node)
+    }
+
+    private fun handleCall(call: Call) {
```
maximiliankaul, line 78: Wouldn't it be easier to understand the logic, if there would be a check for supported calls here over having it hidden in the `allocationSizeOf` function?

cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/concepts/memory/Memory.kt
```diff
@@ -71,13 +72,23 @@ open class Allocate(
     concept: Concept,
     /** A reference to [what] is allocated, e.g., a variable. */
     var what: Node?,
+    /**
+     * The expression that determines the size of the allocation, if known. For example, the `N`
+     * argument of `malloc(N)`, the product `M * N` of `calloc(M, N)`, or the dimension expression
+     * of `new T[N]`. `null` when the size couldn't be derived statically (e.g. for managed
```
maximiliankaul: For my understanding: dimension expression of `new T[N]` -> translates to `sizeof(T)*N`?

cpg-analysis/src/main/kotlin/de/fraunhofer/aisec/cpg/analysis/abstracteval/value/ArrayValue.kt
```diff
@@ -147,10 +148,15 @@ class ArrayValue : Value<LatticeInterval> {
                 }
             }
             is Call -> {
-                if (node.name.localName == "malloc") {
-                    val length = (node.arguments.singleOrNull()?.value?.value as? Number)?.toLong()
-                    length?.let { LatticeInterval.Bounded(length, length) }
-                        ?: LatticeInterval.BOTTOM
+                // Read the allocation size from an attached [Allocate] concept overlay; a
```
maximiliankaul: maybe add "size in bytes" to clarify the unit?

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/memory/cxx/CXXMemoryAllocationPassTest.kt
```diff
@@ -0,0 +1,142 @@
...
+    /** `p = malloc(128)` after a prior declaration → `what` resolves through the LHS Reference. */
+    @Test
+    fun testMallocViaAssign() {
+        val allocate = allocateIn("malloc_via_assign", "malloc")
+        val sizeLit = assertIs<Literal<*>>(allocate.size)
+        assertEquals(128L, (sizeLit.value as Number).toLong())
+        assertIs<Variable>(allocate.what)
+        assertEquals("p", (allocate.what as Variable).name.localName)
+    }
+
```
maximiliankaul, line 95: Tests for `what` are missing starting here.

maximiliankaul (approved):
I like this PR. Only question (which I don't know how to answer): do we want all those `api` dependencies over `implementation`? My gut feeling is to use `api` only where we have a good reason to, but I'm not sure for best-practices here.

PR #2776, opened 2026-06-01
Fix loop in type listener when writing to unary operators

KuechA:
When we write to `*x` in go, this is still handled as `UnaryOperator` and causes a loop in the type listener system because `*` observes `x` and `x` observes `*`. This PR removes `*` as the type listener of `x` to remove the loop.


oxisto:
Would it make sense to already model it as a DereferenceOperaror or however it’s called

KuechA:
> Would it make sense to already model it as a DereferenceOperaror or however it’s called

For sure. But I guess it's an orthogonal problem. And I don't know how is the type listener implementation of the `PointerDereference` and `PointerReference` (let alone what it should look like)

oxisto (approved):

PR #2777, opened 2026-06-02
Add tests for type propagation and hashing issues

KuechA:
Previous PRs fixed some issues with type handling and loops. This PR adds tests for them.


oxisto (approved):
LGTM. If you are improving things on the Go side: It might be worth replacing `&x` with a `PointerReference` now as well.

PR #2778, opened 2026-06-02
Tagging typically depends on DFG and also on EOG

KuechA:


cpg-concepts/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/concepts/TagOverlaysPass.kt
```diff
@@ -43,6 +45,8 @@ import de.fraunhofer.aisec.cpg.passes.configuration.DependsOn
 @DependsOn(SymbolResolver::class)
 @DependsOn(ControlFlowSensitiveDFGPass::class, true)
 @DependsOn(PointsToPass::class, true)
+@DependsOn(DFGPass::class, true)
+@DependsOn(EvaluationOrderGraphPass::class)
```
oxisto, line 49: SymbolResolver should already depend on EOG, so does CF-DFG and/or PointsToPass. Why is this even necessary?

KuechA: For CF-DFG and PointsToPass: These are optional, so if we do not register them, it means that only the SymbolResolver is left. I saw this pass being executed before the DFG Pass which was strange. And I prefer to have such dependencies/assumptions written down explicitly. But I agree, the EOG may not be strictly necessary here.

maximiliankaul (approved):

PR #2780, opened 2026-06-02
Try to improve performance of collection stuff

KuechA:
The implementations of filter(), map(), toList() etc. keep copying all elements of a collection to a new collection. If we do this often enough, this can probably have significant performance impact a) because we allocate a lot of memory which has to be freed afterwards, b) because we potentially iterate the collections multiple times. By checking the current usages of these functions and providing new utility functions, this can be improved.

Typical patterns are:

* `map{ ... }.toList() (or toSet())` -> can be replaced with `mapTo(mutableSetOf()) { ... }`
* `filter{ ... }.toList() (or toSet())` -> can be replaced with `filterTo(mutableSetOf()) { ... }`
* `filter { ... }.map{ ... }` -> can be replaced with new utility function `mapFilteredTo(mutableListOf(), {predicate}), {transformation}`


konradweiss (approved):
Very cool improvements, we should keep the use of these functions, native and those we introduced in mind in future reviews.

PR #2790, opened 2026-06-12
Reduce memory footprint of `Node` classes

KuechA:
We allocate a significant amount of memory in the `Node` classes. Partially because we have a lot of attributes (which may not even be used), partially due to inefficient datastructures. This PR implements a couple of things:

* It extracts `locals` from all `Expression`s and adds it where necessary.
* It changes the `EdgeList` and `EdgeSet` implementations as they often hold a single value but are always a `List`/`Set`.
* Lazy initialization of unused properties in `Node`s and `Edge`s (e.g. `additionalProblems`, `assumptions`).
* Reuse `name` as `code` for `Reference`s which is the case most of the time
* Minor improvements in `Name` class
* Identity-based management in `EdgeList` and `EdgeSet` to speed up the runtime

This seems to have a positive impact - if my tests were correct this may reduce it from ~10 to ~8.5 GB for a 300k LOC project and also helped to reduce the run-time from ~20min to ~16-18min.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Interfaces.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Interfaces.kt
@@ -29,8 +29,11 @@ import de.fraunhofer.aisec.cpg.PopulatedByPass
 import de.fraunhofer.aisec.cpg.frontends.Language
 import de.fraunhofer.aisec.cpg.graph.declarations.Method
 import de.fraunhofer.aisec.cpg.graph.declarations.Operator
+import de.fraunhofer.aisec.cpg.graph.declarations.ValueDeclaration
 import de.fraunhofer.aisec.cpg.graph.declarations.Variable
 import de.fraunhofer.aisec.cpg.graph.edges.MemoryAddressEdges
+import de.fraunhofer.aisec.cpg.graph.edges.ast.AstEdge
+import de.fraunhofer.aisec.cpg.graph.edges.ast.AstEdges
 import de.fraunhofer.aisec.cpg.graph.edges.flows.Dataflows
 import de.fraunhofer.aisec.cpg.graph.edges.flows.FullDataflowGranularity
 import de.fraunhofer.aisec.cpg.graph.expressions.BinaryOperator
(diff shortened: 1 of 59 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/Function.kt
```diff
@@ -85,7 +85,7 @@ open class Function :
      */
     @Relationship(value = "INVOKES", direction = Relationship.Direction.INCOMING)
     val calledByEdges: Invokes<Function> =
-        Invokes<Function>(this, Call::invokeEdges, outgoing = false)
+        Invokes<Function>(this, mirroredCollection = { (it as Call).invokeEdges }, outgoing = false)
```
oxisto, line 88: this seems very ... hacky :D can we improve the function/constructor interface  somehow to make this more readable? the old style was very readable in contrast.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Interfaces.kt
```diff
@@ -45,8 +48,22 @@ import de.fraunhofer.aisec.cpg.passes.DFGPass
 import de.fraunhofer.aisec.cpg.passes.PointsToPass
 import de.fraunhofer.aisec.cpg.passes.SymbolResolver
 import de.fraunhofer.aisec.cpg.persistence.DoNotPersist
+import de.fraunhofer.aisec.cpg.persistence.Relationship
 import de.fraunhofer.aisec.cpg.sarif.PhysicalLocation
 
+interface HasLocals {
```
oxisto, line 54: needs kdoc

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Name.kt
```diff
@@ -69,8 +69,16 @@ class Name(
      * this is basically a cache for [toString]. Otherwise, we would need to call [toString] a lot
      * of times, to implement the necessary functions for [CharSequence].
      */
-    private val fullName: String by lazy {
-        (if (parent != null) parent.toString() + delimiter else "") + localName
+    private var qualifiedFullNameCache: String? = null
+
+    private fun fullName(): String {
```
oxisto, line 74: why is this a function now instead of a property? that changes things on many places in the code unnecessary and might even impact serialization.

Can't you just use `val fullName { get() { lazy { .... } }` which basically does what you want.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NodeBuilder.kt
```diff
@@ -154,6 +155,18 @@ fun Node.applyMetadata(
                 defaultNamespace
             }
         this.name = this.newName(name, doNotPrependNamespace, namespace)
+
+        // For most references, code and name are the same token. If their textual value matches,
+        // reuse the name string instance to avoid duplicate string objects.
+        if (this is Reference) {
```
oxisto, line 161: Isn't the code already set at this stage? so effectible you are adding more memory pressure here until the GC decides to clean up the original code reference string. I am not sure if this is actually of any benefit

oxisto (commented):
So it's good that it seems to lower memory pressure, I suspect it's mainly about the edge collections. Unfortunately they now went from barely manageable code to completely unmanageable code :(

KuechA:
obsoleted by #2829

PR #2794, opened 2026-06-15
More configuration of followX functions

KuechA:


KuechA:
Should be obsolete with #2850

PR #2795, opened 2026-06-18
Fix phantom argument attachment in Fluent DSL member expressions

narangtanmay:
## Summary

`member()` with an explicit `base` lambda (e.g. `call("doSomething") { member("field1") { ref("s1") } }`) attached the base expression to the wrong `ArgumentHolder`. A plain lambda resolves its enclosing holder from where it's written, not where it's invoked, so `ref("s1")` resolved to the outer `Call` and added itself as a phantom second argument.

`base` now declares its own `Holder` context parameter, bound at invocation rather than capture. Calling it under a no-op `DiscardHolder` means any `ref()`/`member()` inside it attaches there instead of leaking into the real outer holder.

Removes the `makeMagic` per-callsite flag previously needed as a workaround at `ref()` call sites.

## Test plan

- [x] `cpg-core:test` (`FluentTest.testMemberWithBaseAttachesExactlyOnce`)
- [x] `cpg-analysis:test` (`QueryTest`)
- [x] `cpg-core:compileKotlin`

### DIFF
--- a/cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/testcases/Passes.kt
+++ b/cpg-analysis/src/test/kotlin/de/fraunhofer/aisec/cpg/testcases/Passes.kt
@@ -60,7 +60,9 @@ class Passes {
                                         thenStmt { ref("y").inc() }
                                         elseStmt { ref("y").dec() }
                                     }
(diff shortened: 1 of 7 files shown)


KuechA:
We removed Fluent DSL, so this is no longer necessary

PR #2796, opened 2026-06-19
Improve FollowXUntilHit

morbitzer:
FollowXUntil hit can take a very long time / infinite when recursion is involved. 
Also it should be more efficient, e.g. by caching already searched paths. 

PR also contains a fix for the PointsToPass that improves handling of PMVs during recursion


KuechA:
Should be obsolete with #2850

PR #2799, opened 2026-06-22
PTP: Improve handling of global structs

morbitzer:
In some occasions, the PtP was not able to correctly draw the prevDFG of PMVs. This should be fixed with this PR.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PointsToPass.kt
```diff
@@ -1539,12 +1530,22 @@ open class PointsToPass(ctx: TranslationContext) : EOGStarterPass(ctx, orderDepe
                 // Since we already have the argVal, AKA the memoryAddress
                 // of the argDerefVal, we simply fetch the last write for the
                 // argVal from the declarationState and add the properties
+                // TODO: what if argVal is global?
```
konradweiss, line 1537: Is this question one that we can address in this PR, is not relevant anymore or is here as a reminder for the future?

konradweiss (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PointsToPass.kt
morbitzer: Handling the TODO would require some more thought and work, so I would consider it a reminder for the future.

PR #2801, opened 2026-06-22
PtP: Add DFG edges to PMVs of recursive functions

morbitzer:
So far, the PtP failed to draw prevDFG edges for the PMVs of functions that are recursively called. This should work now.


konradweiss (approved):
Missed lines in in the codecov report are the logging calls that are not enabled, which is not relevant functionality to test. Are they necessary though? From the content of the message I think we don't need it in production and therefore can remove both and make codecov happy.

PR #2805, opened 2026-06-25
Add cpg_get_last_write MCP tool

narangtanmay:
## Summary

Adds an MCP tool that returns the writes that may reach a given node's value, one-hop, control-flow-sensitive reaching definitions.

- `Node.reachingWrites()` reads the node's existing `prevDFG` edges (already materialized by `PointsToPass`'s kill/union analysis) and returns them as `ReachingWrite(source, granularity, functionSummary)`.
- `cpg_get_last_write` exposes this over MCP: resolves a node by ID, runs `reachingWrites()`, returns the writing nodes with their dataflow granularity and whether the edge came from a function summary.

This is deliberately different from `followPrevDFG`: that API returns a single witness path by design (`findAllPossiblePaths = false`), so on a variable written on multiple branches before a merge it picks one arbitrary write. `reachingWrites()` returns all of them, the full "may" set.

## Test plan

No automated tests included yet, open question whether to add them before merge.

- [ ] Manually exercised against a small fixture with branch writes before a merge point, confirmed multiple `prevDFG` sources come back
- [ ] `./gradlew :cpg-core:compileKotlin :cpg-mcp:compileKotlin` clean


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -408,6 +409,15 @@ fun Node.collectAllPrevDFGPaths(): List<NodePath> {
         .map { it.second }
 }
 
+data class ReachingWrite(
```
maximiliankaul, line 420: Doc please.

maximiliankaul (approved):
LGTM

PR #2811, opened 2026-07-02
Improvements to C frontend

oxisto:
Tighten the C frontend's precision when a translation unit pulls in a full system-header chain (verified against the macOS SDK's `stdio.h`, `stdlib.h`, `string.h`, `math.h`).
                                                                                                                                                                         
  **Frontend / resolver fixes**
  - `ExpressionHandler`: for unary `*` / `&`, stop leaking the operand's name into the resulting `PointerDereference` / `PointerReference` when the operand isn't itself a `Reference` (`*(_p++)` no longer becomes a
  phantom `Variable "++"`), and reuse the already-handled operand instead of running the handler twice.
  - `SymbolResolver`: skip empty-named references (they're `*(complex_expr)` / `&(complex_expr)` cases with no lookup target) so we don't infer phantom global variables; and only synthesize inferred constructors when
  the language actually has classes.
  - `DeclaratorHandler`: stop adding an implicit default constructor to every C struct — plain C has no `struct::struct` method.
  - `TypeResolver`: fix an operator-precedence bug (`A && B || C` → `A && (B || C)`) that let non-`ObjectType`s (e.g. `FunctionPointerType`) reach record inference and log spurious errors.
  - `tryRecordInference` / `Inference.inferRecordDeclaration`: early-return on non-`ObjectType` and empty-named `ObjectType`; demote the previous ERROR log to debug.

  **New opt-in API: compiler-builtins prelude**
  - `TranslationConfiguration.injectCompilerBuiltins(target: String?)` — pass a target name (e.g. `"darwin-arm64"`) to prepend a compiler-intrinsic prelude to every TU (CDT `ExtendedScannerInfo` include-files,
  equivalent to `clang -include`). Unknown targets warn and skip.
  - Resource layout under `cpg-language-cxx/src/main/resources/.../frontends/cxx/builtins/`:
    - `common.h` — cross-target clang builtins (`__builtin_bswap*`, `__builtin_fabs*`, `__builtin_inf*`, `_Float16`, `__int128_t`).
    - `darwin-arm64/cpg_builtins.h` — all 478 clang preprocessor predefines + arm64 `__builtin_va_list` typedef + `#include <common.h>`.
  - Adding another target is a one-line append to `BUILTIN_RESOURCES` plus the new resource file.
  
  **Guardrail test**
  - `CSystemIncludeTest` (`@EnabledOnOs(OS.MAC)` + arm64 SDK check): parses a `hello.c` that exercises `printf`/`malloc`/`free`/`strlen`/`strcpy`/`sqrt` against the real macOS SDK include chain, then asserts each of
  those resolves to a real (non-inferred) declaration, no references are left unresolved, and the only inferred declarations are the expected opaque forward-decls (`__sFILEX`, `_malloc_zone_t`) plus three documented
  CDT parser quirks.  
  - Result: 0 `SymbolResolver` warns, 0 `Inference` errors on the expanded include chain (was 3 errors + 5 phantom decls before).


cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/ExpressionHandler.kt
```diff
@@ -431,20 +431,26 @@ class ExpressionHandler(lang: CXXLanguageFrontend) :
...
+            // PointerReference / PointerDereference are Reference subclasses, so the name we
+            // hand in is fed into symbol resolution. Only meaningful when the operand is itself
+            // a Reference (`*x`, `*p->f`, `&x`); for arbitrary sub-expressions like `*(p++)` we
+            // must pass a null name, otherwise the SymbolResolver looks up nonsense like "++"
+            // and infers a phantom global variable.
+            //
+            // We also reuse the already-computed `input` node — calling handle(ctx.operand) a
+            // second time creates a duplicate, orphaned subtree.
+            val operand = input
+            val refName = (operand as? Reference)?.name
```
KuechA, line 444: What if `operand` is not a reference?

Why do we need the `operand` variable? Couldn't we directly use `input`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -739,6 +756,21 @@ private constructor(
...
+         * Enable prepending a compiler-intrinsic prelude to every C/C++ translation unit for the
+         * given target (e.g. `"darwin-arm64"`). The [target] names a folder under the frontend's
+         * classpath resource tree `builtins/`, whose `cpg_builtins.h` will be included by the
+         * parser before any user code. Pass `null` (the default) to disable.
+         *
+         * Enables references to clang/gcc built-ins (e.g. `__builtin_va_list`) to resolve to real
+         * typedefs / prototypes instead of being inferred as bogus structs. Adds a small number of
+         * implicit declarations per TU.
+         */
+        fun injectCompilerBuiltins(target: String?): Builder {
```
KuechA, line 769: To be honest, I don't like the idea of making the configuration of the CPG translation even more complicated. We MUST make this significantly easier. In my opinion, it's very difficult to correctly add some code/repository to the analysis without mistakes with the current interface.

oxisto: I'm up for ideas, but we basically need to this step in order to properly parse C/C++ files without any errors.

oxisto: Let's merge #2812 in first and re-visit this afterwards.

oxisto:
Will revisit this after #2812

PR #2812, opened 2026-07-04
Add high-level `Project` API for configuring analysis targets

oxisto:
## Motivation

The API for configuring an analysis target has grown too complex: `TranslationConfiguration` mixes ~28 constructor parameters covering what to analyze, language-specific parsing options, pass setup and debug flags, and every consumer (codyze CLI, codyze script DSL, console, MCP, neo4j) duplicated the same boilerplate — including four copies of the ten-line `optionalLanguage(...)` list.

## New `Project` API (cpg-core)

`de.fraunhofer.aisec.cpg.project` is the new high-level entry point. The simplest usages are:

```kotlin
val result = Project.from(Path("main.cpp")).analyze()
val result = Project.from(Path("/path/to/repo")).analyze()
```

with a builder DSL for more control (components, exclusion filters, target environment, and a `translation {}` escape hatch to the underlying `TranslationConfiguration.Builder`).

- **`ComponentDefinition`** replaces the parallel `softwareComponents`/`topLevels` maps.
- **`TargetEnvironment`** describes the OS, architecture (incl. pointer width) and environment variables the project is assumed to run on. It defaults to the host, can be overridden for cross-target analysis, and is available to frontends via `TranslationConfiguration.targetEnvironment`.
- **Auto-detection**, on two independent axes:
  - **Languages**: when no `languages {}` block is given, the project directory is scanned for file extensions, and only default languages with a matching extension are registered (languages without declared extensions, e.g. Go and C/C++, are always registered since they rely on their own detector instead). Falls back to registering every default language if the scan finds nothing or the target is a single file.
  - **Components & settings**: a single `Detector` SPI answers *"which components does this project consist of, and which project-wide settings apply?"* (symbols, include paths, compilation database) in one `detect(root, environment)` call per detector, invoked once on the project root. Languages can implement `Detector` directly; standalone detectors (e.g., the convention-based `DirectoryComponentDetector`) can be added manually. Results are deduplicated by detector name (so `CLanguage`/`CPPLanguage` sharing the same C/C++ detection logic only contribute once). User configuration always wins over detection, and detection results are recorded on the `Project` for inspection.

## Detectors

- **Go**: one component per `go.mod` in the tree, named after the module path; `GOOS`/`GOARCH` are derived from the `TargetEnvironment` instead of implicitly assuming the host. The former `golang.Project.buildProject` moved into the opt-in, tool-backed `GoBuildDetector` (`go list`-based stdlib/vendor resolution, with a bounded timeout so a hanging `go` invocation can't leak a subprocess); go1.x build tags now come from the `go` directive.
- **C/C++**: components and settings from a `compile_commands.json` in the root or `build/` folder, wiring the compilation database into the configuration.

## Consumers

- **cpg-mcp**: `cpg_analyze`/`cpg_translate` accept a `path` to a file or project directory with full auto-detection (incl. compilation databases); results report detected components and detection notes. `SetupConfig.kt` is gone.
- **codyze-core**: `AnalysisProject.temporary` and the `.codyze.kts` DSL build through `Project.from`; both gain auto-detection when no explicit sources/components/modules are configured.
- **codyze-console**: the analyze endpoint follows the same pattern.
- **cpg-neo4j**: keeps its explicit CLI options but reuses the shared `Project.defaultLanguages` list.

## Notes

- `Path.topLevel` now prefers an include path over the component top-level when a file lies outside the component, so dependency translation units (e.g., Go stdlib) get stable, meaningful names (`fmt/print.go`) instead of collapsing to bare file names.
- Nothing reads `targetEnvironment` in the C/C++ frontend yet (built-in macros, pointer widths); that is a planned follow-up, as is rebasing `AnalysisProject` itself onto `Project`.


cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXProjectDetection.kt
```diff
@@ -0,0 +1,86 @@
...
+    val file = directory.resolve(COMPILE_COMMANDS)
+    if (!file.exists()) {
+        return listOf()
+    }
+
+    val db = tryLoadCompilationDatabase(file) ?: return listOf()
+    return db.components.map { (name, files) ->
+        ComponentDefinition(name, root = directory, sources = files.map(File::toPath))
+    }
+}
```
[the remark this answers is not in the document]
oxisto: Good catch. Fixed in 2aa75e5cfc1: `detectCxxComponents` now uses the same lookup as `detectCxxSettings` (database in the visited directory *or* its `build` folder), so the components are rooted in the project directory. The duplicate result produced when the walk later enters `build/` itself is dropped by the name-based de-duplication, since the walk visits parents first. Added a regression test for the `build/compile_commands.json` layout.

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -301,16 +302,9 @@ class Application : Callable<Int> {
...
-                .optionalLanguage("de.fraunhofer.aisec.cpg.frontends.java.JavaLanguage")
-                .optionalLanguage("de.fraunhofer.aisec.cpg.frontends.golang.GoLanguage")
-                .optionalLanguage("de.fraunhofer.aisec.cpg.frontends.llvm.LLVMIRLanguage")
-                .optionalLanguage("de.fraunhofer.aisec.cpg.frontends.python.PythonLanguage")
-                .optionalLanguage("de.fraunhofer.aisec.cpg.frontends.typescript.TypeScriptLanguage")
-                .optionalLanguage("de.fraunhofer.aisec.cpg.frontends.ruby.RubyLanguage")
-                .optionalLanguage("de.fraunhofer.aisec.cpg.frontends.jvm.JVMLanguage")
-                .optionalLanguage("de.fraunhofer.aisec.cpg.frontends.ini.IniFileLanguage")
+                .also { builder ->
+                    Project.defaultLanguages.forEach { builder.optionalLanguage(it) }
```
peckto, line 306: In the neo4j application, currently only the default languages are used from the Project.

peckto:
How we want to handle multi-language projects? Should all source files from all languages be loaded, or just one? Some options currently work only with one language, like the `unity-build` option for CXX Projects. This does currently not work with multi-language projects.

cpg-mcp/src/main/kotlin/de/fraunhofer/aisec/cpg/mcp/mcpserver/tools/CpgAnalyzeTool.kt
```diff
@@ -178,20 +168,32 @@ fun runCpgAnalyze(
...
-            topLevel = file,
-            files = listOf(file.absolutePath),
-            includePaths = emptyList(),
-            runPasses = runPasses,
-        )
-    config.disableCleanup = !cleanup
+    val project =
+        Project.from(path) {
+            if (!runPasses) passes {}
+            translation {
```
peckto, line 180: Do we have/want some default/auto config for the translation config as part of the project auto detect?
Not sure if we want to do this in this PR or another.

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXProjectDetection.kt
```diff
@@ -0,0 +1,78 @@
...
+        .firstOrNull { it.exists() }
+}
+
+/**
+ * Detects C/C++ project structure and settings in one pass. Looks for a [COMPILE_COMMANDS] file in
+ * [root] or its `build` folder. If found, the database supplies per-file include paths and
+ * preprocessor defines, and its targets are translated into [ComponentDefinition]s rooted in
+ * [root]. Returns `null` if no compilation database is found.
+ */
+internal fun detectCxx(root: Path): DetectionResult? {
```
maximiliankaul: maybe `detectCxxProject`? I got confused with the two `return null` before understanding that simply means no cxx project was detected (but handling cxx code/files without a compilation db isn't affected by this).

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXProjectDetectionTest.kt
```diff
@@ -0,0 +1,140 @@
...
+import kotlin.io.path.writeText
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertTrue
+import org.junit.jupiter.api.io.TempDir
+
+class CXXProjectDetectionTest {
+    @Test
+    fun testDetectCompilationDatabaseInBuildFolder(@TempDir tmp: Path) {
```
maximiliankaul: I fail to see the benefit of creating the files at runtime over having them as resources... Is this to have known absolute paths? If yes: can't we re-write the tests to use relative paths? I believe clang supports relative paths for compilation databases.

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
```diff
@@ -301,16 +302,9 @@ class Application : Callable<Int> {
     fun setupTranslationConfiguration(): TranslationConfiguration {
         val translationConfiguration =
```
peckto, line 303: Using still the "old" interface and not the new Project.
Other Apps like mcp or codyze use the Project interface.
Not sure if we want to do this in this PR or another.

peckto (approved):

maximiliankaul (approved):
LGTM

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CXXProjectDetectionTest.kt
oxisto: I extracted this into actual files

cpg-mcp/src/main/kotlin/de/fraunhofer/aisec/cpg/mcp/mcpserver/tools/CpgAnalyzeTool.kt
oxisto: Follow up

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
oxisto: Follow up

cpg-mcp/src/main/kotlin/de/fraunhofer/aisec/cpg/mcp/mcpserver/tools/CpgAnalyzeTool.kt
oxisto: Agreed this would be a nice improvement — auto-deriving sensible translation-config defaults (debug parser, include loading, inference config) as part of `Project` detection. I'd rather keep this PR scoped to the core `Project`/detection API though, so let's track this as a follow-up rather than expanding this one further.

cpg-neo4j/src/main/kotlin/de/fraunhofer/aisec/cpg_vis_neo4j/Application.kt
oxisto: Good catch, and agreed — the neo4j CLI app should move to the `Project` API like MCP/Codyze/console already do. It touches a fair amount of CLI flag handling (`topLevel`, `includePaths`, `softwareComponents`, `useUnityBuild`) though, so I'd like to do that migration as its own follow-up PR rather than growing this one further.

PR #2816, opened 2026-07-06
Fix broken collectAllPrevCDG

morbitzer:


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA, line 1040: @KuechA Does this make sense here? Do we also have to change other parts of the code where the edges are used in the wrong direction?

KuechA: We had to change line 1051. Was tested. Works.

KuechA (approved):

PR #2821, opened 2026-07-13
Fixing implicit modified base types being incorrectly parsed

konradweiss:
`long ** dataPtr` is incorrectly parsed. Instead of generating a PointerType, it runs into an edge case trying to handle nameless types that is IMO incorrectly handled. The old version used the specifiers `sc_unspecified` to decide that we are in the special case. This should be incorrect as sc stands for storage classified and not a type itself. So i changed it to `t_unspecified` that should signal if the CDT cannot parse the type. This however was also flagged and this is because `long` is an implicit `long int`. Howevery, CDT will set the flag `t_unspecified` but `isLong` as true.

To solve this second issue I only run into the edge case if `t_unspecified` is set and none of the modifier flags is true, e.g. `isLong` and `isShort`


cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/enhancements/types/TypeTests.kt
```diff
@@ -29,12 +29,57 @@ import de.fraunhofer.aisec.cpg.*
...
+                loc != null && loc.startLine == 7
+            }
+        // fallback: take any call to foo if line-based lookup fails
+        val actualCall = call ?: calls.firstOrNull()
+        assertNotNull(actualCall)
+        assertInvokes(actualCall, foo)
+        assertIs<Call>(actualCall)
+
+        // The function's first parameter should be a pointer-to-pointer to 'long'
+        val param = foo.parameters.firstOrNull()
```
oxisto: nit: you could also just use `foo.parameters["dataPtr"]` here

oxisto (approved):

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/enhancements/types/TypeTests.kt
konradweiss: done

PR #2829, opened 2026-07-15
 Reduce CPG memory footprint and pass/frontend runtime

KuechA:
## Summary                                                                                                                                                                                      

  Reduces the in-memory footprint and analysis runtime of the CPG without changing
  the graph model or any public pass/edge semantics. The work targets the three
  cost centers that dominate at scale: millions of `Node`s each carrying rarely-used
  properties, millions of `Edge`s each over-allocating, and fixed-point/graph passes
  doing redundant work.

  All changes are isolated, independently reviewable commits. Each was validated
  against the full test suite (excluding the `@Ignore`d `ExternalPerformanceTest`s).

  ## Memory

  - **Per-node edge containers allocated lazily.** CDG, PDG, overlay, memory-value/
    -address (on `Expression` *and* `Declaration`), `annotations`, `locals`, and
    `template*` containers are only populated by optional/specific passes yet were
    eagerly constructed on every node. They now allocate on first use. Measurements
    over the test corpus showed these empty on 95–100% of nodes.
  - **Per-edge overhead cut.** Edges are the most numerous objects: their `labels`
    set is now a shared per-subclass constant instead of a fresh `Set` per edge,
    `Edge.assumptions` is lazily allocated, and `ControlDependence.branches`
    defaults to the shared empty set.
  - **`EdgeList`/`EdgeSet` small-storage.** `EdgeList` no longer eagerly allocates a
    backing array; `EdgeSet` stores up to two elements inline before falling back to
    a `HashSet` (measurements: the vast majority of edge sets hold 0–2 elements).
  - **`Node.assumptions`/`additionalProblems`** use a lightweight `SmallMutableSet`;
    **`IdentitySet`** allocates its backing map lazily (helps every empty
    `typeObservers`, etc.).
  - **`ArtifactLocation` interned per URI** instead of reconstructed (with a
    recomputed `fileName`) per node.

  ## Runtime

  - **`Node.hashCode` cached** (invalidated by the `name`/`location` setters), since                                                                                                              
    passes hash nodes constantly. `Node.id` gets faster for free.
  - **`ScopeManager.lookupSymbolByName` cached**, invalidated on symbol-table
    mutation — cuts repeated scope-chain walks during resolution (frontend-side).
  - **`Worklist.push` O(n) → O(1)** by replacing a per-push linear scan with a
    `LinkedHashMap`.
  - **CDG fixed-point state uses a persistent map** (`kotlinx.collections.immutable`),
    making the per-branch `duplicate()` O(1) structural-sharing instead of a deep
    copy. Measured on `slow-cdg.c`: ~46.6s → ~38.7s median full analysis.
  - **`ProgramDependenceGraphPass` reachability memoized**, and repeated
    `allChildren`/`flattenAST` walks hoisted out of hot loops.

  ## Correctness / safety notes

  - The `hashCode` cache is *behaviorally identical* to the live computation: `Name`
    is immutable and no `PhysicalLocation` is mutated in place, so the `name`/
    `location` setters observe every change that can affect the hash.
  - Lazily-backed containers are never part of `equals`/`hashCode`; unwrapped views
    are `@DoNotPersist`, matching the previous `by unwrapping` delegates, so the
    Neo4j schema is unchanged.
  - The one new dependency is `kotlinx.collections.immutable` (JetBrains, for the
    persistent-map lattice).

  ## Explicitly not included (discussed, deferred as too risky/low-value)
  - Caching `Node.id` (unsound to invalidate — depends on the whole ancestor chain).
  - Pass-level parallelism, type/string interning, scope-collection laziness,
    `PointsToPass` persistent lattice.

  ## Testing
  Full unit-test suite green after each commit. Additional tests on real-world code bases
  to measure impact.

### DIFF
--- a/cpg-core/build.gradle.kts
+++ b/cpg-core/build.gradle.kts
@@ -52,6 +52,7 @@ dependencies {
     implementation(libs.kotlin.reflect)
     implementation(libs.jacksonyml)
     implementation(libs.kotlinx.coroutines.core)
+    implementation(libs.kotlinx.collections.immutable)
 
     testImplementation(libs.junit.params)
     integrationTestImplementation(libs.kotlin.reflect)
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
@@ -77,6 +77,44 @@ class ScopeManager(override var ctx: TranslationContext) : ScopeProvider, Contex
      */
     private val nameScopeMap: MutableMap<Name, NameScope> = mutableMapOf()
 
+    /**
+     * Caches the result of [lookupSymbolByName] (for calls without a custom [predicate], which
+     * cannot be cached safely). [lookupSymbolByName] can be called repeatedly for the same symbol
+     * (e.g. once per reference to the same variable, or once per candidate scope during ambiguous
+     * call/member resolution), and walking the scope chain for every single call is unnecessarily
+     * expensive.
+     *
+     * The cache is invalidated wholesale (see [symbolTableGeneration]) whenever a symbol table is
+     * mutated (see [invalidateSymbolLookupCache]), rather than per-entry, to keep invalidation
+     * trivially correct. This is cheap in practice because virtually all symbol table mutations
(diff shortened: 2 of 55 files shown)


oxisto (approved):
it's not really beautiful code, but if it helps :(

PR #2830, opened 2026-07-16
cpg-ai module

maximiliankaul:
This PR renames `cpg-mcp` to `cpg-ai` and moves LLm functionality from `codyze-console` to the new `cpg-ai` module. Thus, enabling other projects to use ai functionality without depending on `codyze-console`.

Downside of this approach here: there are now JSON wrappers around the `cpg-ai` functionality for `codyze-console`. Those need to be kept up-to-date -> we should come up with a better solution.


codyze-console/src/main/kotlin/de/fraunhofer/aisec/codyze/console/Main.kt
```diff
@@ -63,11 +59,12 @@ fun ConsoleService.startConsole(host: String = "localhost", port: Int = 8080) {
...
-    McpServerHelper.startMcpServer(8081)
+    log.info("Starting MCP server with streamable HTTP on port {}...", 8081)
+    runHttpMcpServerUsingKtorPlugin(port = 8081, server = configureDefaultServer())
 
     val translationResult = getTranslationResult()?.analysisResult?.translationResult
     if (translationResult != null) {
-        McpServerHelper.setGlobalAnalysisResult(translationResult)
+        globalAnalysisResult = translationResult
     }
     chatService.connect()
```
lshala: Might need a try-catch block

maximiliankaul: added

codyze-console/README.md
```diff
@@ -3,6 +3,9 @@
 A web application for Codyze with an optional AI chat, which is enhanced by an MCP client that acts as an agent.
 The agent uses the tools of the CPG MCP server to analyze code and answer questions.
 
+> [!IMPORTANT]
+> codyze-console has a hard build dependency on the `cpg-ai` module, so `enableAIModule=true` must be set in `gradle.properties` (see [AI Chat Features](#ai-chat-features) below) before building or running codyze-console at all - otherwise the build fails.
```
oxisto: That sounds error prone, we need to come up with a better way of handling this. Can we auto-enable the AI module when codyze-console is enabled in the gradle script?

cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/README.md
```diff
@@ -40,7 +40,7 @@ The current implementation uses stdio since Claude Desktop only supports this tr
     {
   "mcpServers": {
     "cpg": {
-      "command": "/path/to/cpg-mcp/build/install/cpg-mcp/bin/cpg-mcp"
+      "command": "/path/to/cpg-ai/build/install/cpg-ai/bin/cpg-ai"
```
oxisto: see above, I think the binary should still be called cpg-mcp or cpg-mcp-server

cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/mcpserver/McpServer.kt
```diff
@@ -60,7 +60,7 @@ fun configureServer(
         this
     }
 ): Server {
-    val info = Implementation(name = "cpg-mcp-server", version = "1.0.0")
+    val info = Implementation(name = "cpg-ai-server", version = "1.0.0")
```
oxisto: mcp?

cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/README.md
maximiliankaul: But mcp is just a protocol. The point of this pr is to collect all/most ai stuff here. Originally, this module only added support for the mcp protocol. But, the point of this pr is to collect also skill handling and every other ai handling here so that other projects can make use of all of our ai features by depending on the cpg-ai module. So, `cpg-mcp` would not describe everything this module is offering (skill handling for example).

oxisto: yes the module can be called `cpg-ai` but the *binary* which only contains the MCP server (the application) should be called `cpg-mcp` because there might be more applications in the module maybe `cpg-new-ai-protocol-server`

codyze-console/README.md
maximiliankaul: `codyze-console` is always enabled. Thus, we could also always enable `cpg-ai`? I don't like this. Probably a better solution would be to make `codyze-console` optional, too (and then try to implement the automatic mechanism for `cpg-ai`)?

maximiliankaul: I could easily make a dependency with gradle: `cpg-ai` configured `on` OR `codyze-console` `on` -> enable `cpg-ai`. However, for this to make sense, we would need something like https://github.com/Fraunhofer-AISEC/cpg/pull/2843 first.

cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/README.md
maximiliankaul: adjusted the path. module is still `cpg-ai` but the binary is called `cpg-mcp` again. good?

cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/Application.kt
```diff
@@ -46,7 +46,7 @@ import kotlinx.io.asSource
 import kotlinx.io.buffered
 import org.slf4j.LoggerFactory
 
-class Application : CliktCommand(name = "cpg-mcp") {
+class Application : CliktCommand(name = "cpg-ai") {
```
oxisto: Still the old path?

cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/OpenWebUI.md
```diff
@@ -9,12 +9,12 @@ pip install mcpo
...
 2. Build the CPG MCP server:
 ```bash
-./gradlew :cpg-mcp:installDist
+./gradlew :cpg-ai:installDist
 ```
 
 3. Start the proxy server:
 ```bash
-uvx mcpo --port 8000 -- /path/to/cpg-mcp/build/install/cpg-mcp/bin/cpg-mcp
+uvx mcpo --port 8000 -- /path/to/cpg-ai/build/install/cpg-ai/bin/cpg-ai
```
oxisto: Still the old path?

oxisto (approved):
Still has the wrong path I think in two cases

codyze-console/README.md
maximiliankaul: dependency implemented: cpg-ai will be enabled if, cpg-ai is enabled OR codyze-console is enabled.

KuechA (approved):

PR #2831, opened 2026-07-16
Removing Fluent DSL

KuechA:
## Summary
  - Removes the Fluent DSL (`graph/builder/Fluent.kt`), a context-receiver-based test helper for building CPG ASTs, along with its implicit `Holder`-as-context-receiver resolution. It relied on a deprecated Kotlin
  feature (`@file:Suppress("CONTEXT_RECEIVERS_DEPRECATED")`) and had several latent bugs: leaf builders self-attached to whichever `ArgumentHolder`/`StatementHolder` was lexically nearest, requiring manual
  `holder -= node.lhs; holder -= node.rhs` undo-hacks for composite expressions, and some operators (e.g. `lt`) never attached at all.
  - Folds the Fluent DSL's actual value — low-boilerplate scope handling and container attachment — directly into the real node builders (`NodeBuilder`, `DeclarationBuilder`, `StatementBuilder`,
  `ExpressionBuilder`). The 16 scope-introducing builders now take optional `enterScope: Boolean` and `init: ((T) -> Unit)?` params, and declaration builders take an optional `holder: DeclarationHolder?` for
  explicit attachment. All new params default to today's behavior (`enterScope = false`, `holder = null`), so this is purely additive for existing frontend call sites — nothing changes for production frontends.
  - Migrates all 23 files that depended on the Fluent DSL (17 in `cpg-core`, 5 in `cpg-analysis`, 1 in `cpg-concepts`) to the extended builder API, and deletes `Fluent.kt` and the now-empty `graph/builder`
  package. `translationResult`/`inferPseudoLocations`, the non-node-construction parts of the old harness, moved to `TestLanguage.kt`/`NodeBuilder.kt`.
  - `Holder`/`DeclarationHolder`/`StatementHolder`/`ArgumentHolder` interfaces are retained as-is — they're genuine domain modeling (e.g. `Function`, `Record`, `Block`, `Call` all implement them) and are still
  used, just via explicit `holder =` arguments instead of implicit context-receiver resolution.
  - Updates 3 KDoc comments (`LanguageFrontend.kt`, `Holder.kt`, `ArgumentHolder.kt`) that referenced "the Node Fluent DSL" to describe the current mechanism.

  Migration is intentionally behavior-preserving, including a few pre-existing Fluent quirks that are now reproduced explicitly with a comment rather than "fixed": `lt`'s missing `ArgumentHolder` context (so the
  built comparison was silently discarded), orphan self-attached statements from `logicAnd`/`logicOr`, and `memberCall(..., construct(X))` leaving a stray `Construction` statement in the enclosing block.

  ## Test plan
  - [x] `./gradlew :cpg-core:test :cpg-analysis:test :cpg-concepts:test` — all green
  - [x] Full multi-module compile (including optional Java/Python frontends) — no regressions
  - [x] `FluentTest`'s scope-placement assertions (`GlobalScope`/`FunctionScope`/`LocalScope`) checked byte-for-byte against pre-migration behavior
  - [x] Repo-wide grep for "Fluent" confirms only explanatory test comments about faithfully-reproduced quirks remain


cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/ExtensionTest.kt
```diff
@@ -99,12 +134,15 @@ class ExtensionTest {
             )
         ) {
             val tr = build {
-                translationResult {
-                    translationUnit {
-                        val record = record("Record") {}
-                        extension(null, record) { method("extFunc") {} }
-                    }
+                val tu = newTranslationUnit(Node.EMPTY_NAME)
```
KuechA: This pattern should ideally be transferred to something like
```
translationResult.components.firstOrNull()?.translationUnits?.add(
    newTranslationUnit(Node.EMPTY_NAME) { tu ->
        newRecord("Record", "class", holder = tu, enterScope = true)
        .... 
    }
)
```

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/declarations/TupleTest.kt
```diff
@@ -136,33 +145,37 @@ class TupleTest {
...
+
+                newFunction("func", holder = tu, enterScope = true) { func ->
+                    func.returnTypes = listOf(objectType("MyClass"), objectType("error"))
+                    func.type = computeType(func)
+                }
+
+                newFunction("main", holder = tu, enterScope = true) { func ->
+                    func.body =
+                        newBlock(enterScope = true) { block ->
+                            // I fear this was too complex for the fluent DSL; so we just use
```
KuechA: This comment can be deleted. Can we also get rid of the intermediate variables and use the `init`-blocks?

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/FluentTest.kt
KuechA, line 1: This test should no longer be required because the Fluent DSL is no longer necessary

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/InterfacesTest.kt
```diff
@@ -28,15 +28,15 @@ package de.fraunhofer.aisec.cpg.graph
...
 import de.fraunhofer.aisec.cpg.graph.expressions.Literal
 import kotlin.test.Test
-import kotlin.test.assertTrue
+import kotlin.test.assertEquals
 
 class InterfacesTest {
 
     @Test
-    fun testRemoveArgument() {
+    fun testSetInitializer() {
```
KuechA: this test should be unnecessary. the file can probably be deleted

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/helpers/ExtensionsTest.kt
```diff
@@ -60,12 +55,25 @@ internal class ExtensionsTest : BaseTest() {
...
-                    function("foo") { body { declare { problemDecl(problemDeclText) } } }
-                        .additionalProblems += ProblemExpression(problemExprText)
+            val tu = newTranslationUnit("foo.bar")
+            scopeManager.resetToGlobal(tu)
+
+            val foo =
+                newFunction("foo", holder = tu, enterScope = true) { func ->
+                    func.body =
+                        newBlock(enterScope = true) { block ->
+                            val declStmt = newDeclarationStatement()
```
KuechA: Should use something like to be consistent. This also happens in other locations.
```
block.statements += newDeclarationStatement() {declStmt ->
    newProblemDeclaration(
                                problem = problemDeclText,
                                problemType = ProblemNode.ProblemType.TRANSLATION,
                                holder = declStmt,
    )
}
```

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/ControlDependenceGraphPassTest.kt
```diff
@@ -175,18 +178,48 @@ class ControlDependenceGraphPassTest {
...
+                    val tu = newTranslationUnit("if.cpp")
+                    scopeManager.resetToGlobal(tu)
+
+                    newFunction("main", holder = tu, enterScope = true) { func ->
+                        func.returnTypes = listOf(objectType("int"))
+                        func.type = computeType(func)
+
+                        func.body =
+                            newBlock(enterScope = true) { block ->
+                                // Fluent's logicAnd/logicOr self-attach their operands (foo()/
```
KuechA: Fluent is no longer relevant here. Why do these comments exist? Remove them and update the code, too to get rid of intermediate variables (this affects the whole file).

konradweiss (approved):
LGTM, we basically merge the fluent DSL with the node building that we already have, and carry over the initialization through a block. Holders have to be defined explicitly but in the frontends the elements can still be added manually. Because these changes are test infrastructure changes I would also ignore IF these code changes have a lower coverage than normally required. We still have to solve some conflicts that have come up since then and if that impacts test fulfillment etc.

PR #2832, opened 2026-07-21
Truncate LLM results

maximiliankaul:
... to stay within token / context limits


maximiliankaul:
No longer needed. Included in #2854.

PR #2842, opened 2026-07-28
Fix duplicate extern variable and failing resolution

KuechA:
## Summary

  Fixes duplicate `Variable` nodes being created for the same global symbol when it has an
  `extern` declaration and a separate definition (or multiple non-`extern` "tentative"
  declarations, C11 §6.9.2) — e.g. `extern int x;` in a header plus `int x = 5;` in the
  including `.c` file. Previously this produced **two** distinct `Variable` nodes in the same
  scope, and `SymbolResolver` would see >1 candidate for any reference to `x` and give up
  (`refersTo = null`) instead of resolving it.

  `Function` already had a declaration/definition concept (`isDefinition`/`.definition`) plus a
  CXX-only pass that links them; `Variable` had no equivalent, and the underlying
  `Scope.addSymbol` mechanism is a plain append-only multimap for every frontend.

  ## Approach

  Rather than patch this in the CXX frontend alone, this generalizes redeclaration handling into
  `Scope`/`ScopeManager` (cpg-core) via a new, opt-in `Language` policy:

  - `Language.supportsDeclarationMerging`, `isRedeclaration(existing, incoming)`,
    `mergeRedeclaration(existing, incoming)` — new `open` members, default to `false`/no-op, so
    every language other than C/C++ is behaviorally unchanged.
  - `Scope.addSymbol` now merges an incoming declaration into an existing, compatible one instead
    of always appending, and returns the canonical declaration (previously returned `Unit`).
    `ScopeManager.addDeclaration` and the sole other `addSymbol` override (`NamespaceScope`)
    propagate this.
  - A new `SymbolMap.collapseRedeclarations()` re-applies the same policy after the existing
    (unchanged) cross-translation-unit scope merge, so a header's `extern` declaration seen by
    multiple `.c` files in one component still resolves correctly regardless of parse order.
  - `CLanguage` implements the actual C/C++ policy: a bare `extern` declaration or a C tentative
    definition (`int x;` repeated without an initializer) is folded into the existing declaration
    for the same symbol at global/namespace scope; two genuine definitions of the same global are
    deliberately left unmerged (ODR violation, surfaced as ambiguous rather than silently
    resolved). `CPPLanguage` disables the tentative-definition case, since a non-class-type
    variable at namespace/global scope is already a full definition in C++.
  - `DeclarationHandler` now tags a declarator with `extern` (read from CDT's
    `IASTDeclSpecifier.storageClass`, distinct from `extern "C" { ... }` linkage specifications)
    and uses the canonical declaration returned by `scopeManager.addDeclaration` when wiring
    nodes into the AST, so a merged-away duplicate is never re-attached.

  **Scope boundary:** fully fixes both symbol resolution and node count for the single-TU case
  (the common case, since CDT inlines `#include`d headers into the same translation unit).
  For the cross-TU case, symbol resolution is fixed, but each TU still keeps its own physical
  `Variable` node for what it declared — mirroring the existing two-node model for `Function`.
  Function-local `extern` referring to a file-scope global is out of scope (would need a
  scope-chain walk-up, not same-scope merging).

  ## Testing

  - `SymbolResolverTest`: restored `testExternVariables` (the original repro) plus new tests for
    order-independence, C tentative definitions, `extern` + initializer, and cross-translation-unit
    resolution.
  - Full `cpg-language-cxx`, `cpg-core`, `cpg-language-python`, and `cpg-language-java` test
    suites pass with zero regressions, confirming the change is a no-op for frontends that don't
    opt into declaration merging and doesn't affect existing C/C++ behavior (`extern "C"` blocks,
    C++ static members, function declaration/definition linking, opaque never-defined externs).


cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CLanguage.kt
```diff
@@ -52,6 +56,53 @@ open class CLanguage :
...
+    open val supportsTentativeDefinitions: Boolean = true
+
+    override fun isRedeclaration(existing: Declaration, incoming: Declaration): Boolean {
+        if (existing::class != Variable::class || incoming::class != Variable::class) {
+            return false
+        }
+        if (existing.scope !is GlobalScope && existing.scope !is NamespaceScope) {
+            return false
+        }
+        existing as Variable
```
KuechA: Can't we use smart casts when checking `existing::class != Variable::class` instead of this feature?

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CLanguage.kt
```diff
@@ -52,6 +56,53 @@ open class CLanguage :
...
+    /**
+     * Whether a bare, non-`extern`, non-initialized redeclaration of a global [Variable] is a valid
+     * "tentative definition" ([ISO/IEC 9899:2011] §6.9.2) rather than an error. This holds for C,
+     * but is overridden to `false` for C++, where a non-class-type variable at namespace/global
+     * scope is already a full definition, even without an explicit initializer, so a second such
+     * declaration would be an ODR violation rather than a redeclaration of the same object.
+     */
+    open val supportsTentativeDefinitions: Boolean = true
+
+    override fun isRedeclaration(existing: Declaration, incoming: Declaration): Boolean {
```
KuechA, line 92: This function should get some documentation/comments

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CLanguage.kt
```diff
@@ -52,6 +56,53 @@ open class CLanguage :
...
+        }
+        if ("extern" in existing.modifiers || "extern" in incoming.modifiers) {
+            return true
+        }
+        // Neither side carries `extern`: this is only a valid tentative-definition redeclaration
+        // in C.
+        return supportsTentativeDefinitions
+    }
+
+    override fun mergeRedeclaration(existing: Declaration, incoming: Declaration) {
```
KuechA, line 121: Needs documentation/comments

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/Language.kt
```diff
@@ -526,6 +526,27 @@ abstract class Language<T : LanguageFrontend<*, *>>() : Node() {
         }
     }
 
+    /**
+     * Whether [Scope.addSymbol] is allowed to merge an incoming declaration into an
+     * already-registered one of the same kind and symbol (see [isRedeclaration]), instead of
+     * appending a duplicate. Used to model languages where multiple textual declarations can refer
+     * to the same underlying object, e.g. C's `extern` declarations and tentative definitions.
+     */
+    open val supportsDeclarationMerging: Boolean = false
```
oxisto: Can we make this a language trait instead? Since the trait is an interface, this interface can then define `isRedeclaration` and `mergeRedeclaration` and the existence of the interface would replace `supportsDeclarationMerging`

KuechA: makes sense. done

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
```diff
@@ -305,6 +306,27 @@ interface HasBuiltins : LanguageTrait {
...
 }
 
+/**
+ * A language trait that specifies that this language allows multiple textual declarations to refer
+ * to the same underlying object, and that [Scope.addSymbol] may therefore merge an incoming
+ * declaration into an already-registered one of the same kind and symbol (see [isRedeclaration]),
+ * instead of appending a duplicate. A common example is C's `extern` declarations and tentative
+ * definitions.
+ */
+interface HasDeclarationMerging : LanguageTrait {
```
oxisto: I wonder if there is a better naming for this, like re-declaration? If we dont come up with a better one, we can leave it at that.

oxisto (approved):
We probably potentially can extend this to form some kind of header de-duplication where we have like 1000 declarations of the same function.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
oxisto: Some choices from Claude:
  1. `HasRedeclarations` — mirrors the existing member name isRedeclaration, so the trait and its methods read consistently. Short and matches the Has<Noun> convention used everywhere else in
  this file (HasGenerics, HasStructs, etc.).
  2. `HasSeparateDeclarationAndDefinition` — names the actual C-language concept (declaration vs. definition are distinct; you can have N declarations but only one definition). More descriptive
  of why the trait exists, at the cost of being longer.
  3. `HasNonUniqueDeclarations` (or `HasRepeatableDeclarations`) — emphasizes the property that a symbol's declaration doesn't have to be singular, without committing to the "definition"
  terminology which may not translate as cleanly to non-C languages that might adopt this trait later.

PR #2848, opened 2026-08-01
Model the C/C++ `static` keyword and access control with context-dependent semantic

KuechA:
# Motivation

In C/C++ the static keyword means three different things depending on where it appears, and public/protected/private control member accessibility — but the CPG previously stored these keywords only as raw
modifier strings, with no canonical, language-agnostic representation. This made it impossible for passes to reason about internal linkage, static members, or access control in a portable way. This PR introduces
a canonical model for visibility and "static-ness" and wires the C/C++ (and, where relevant, Java) frontends into it, so that symbol resolution respects these semantics.

# What changed

The work is split into focused, individually reviewable commits:

1. Canonical Visibility model + consistent isStatic — adds a language-agnostic Visibility enum (PUBLIC/PROTECTED/PRIVATE/INTERNAL/UNKNOWN) with derived helpers (hasInternalLinkage, hasRestrictedVisibility), and
a uniform isStatic for fields.
2. HasKeywordSemantics trait — a language trait that maps a raw keyword + DeclarationContext (GLOBAL/LOCAL/RECORD) to canonical KeywordSemantics(visibility, isStatic), so keyword meaning lives in the language, not the handler.
3. Eager static interpretation in the CXX frontend — the storage class is projected onto the canonical visibility/isStatic at parse time, in the correct context.
4. Isolate C internal-linkage symbols across translation units (3A) — file-static (internal-linkage) symbols no longer leak across translation units during resolution.
5. Filter inaccessible members during resolution (3C) — SymbolResolver drops private/protected candidates that are inaccessible from the point of access, gated behind a new HasAccessControl trait. Conservative by design: it never removes the last candidate, so it can't make a resolvable reference unresolvable. friend/nested-class access is intentionally not modeled (documented in-code).
6. Test coverage top-up — covers the inferred single-static-import branch in JavaImportResolver.

# Testing

New test suites: VisibilityTest, CLanguageKeywordSemanticsTest, CXXStaticSemanticsTest, CXXInternalLinkageTest, CXXAccessControlTest, plus a new case in StaticImportsTest, with C++/Java fixtures. Patch coverage: 98.0% (97/99 added executable lines); the 2 remaining lines are defensive guards that are unreachable for currently-implemented languages (a non-HasAccessControl language producing restricted-visibility members, and a smart-cast guard that cannot be removed).

# Notes for reviewers

- The access-control filter is trait-gated (HasAccessControl), so languages that don't opt in are completely unaffected.
- 3A was implemented as a SymbolResolver candidate filter rather than via FileScope (an earlier FileScope approach caused type-resolution regressions).


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Visibility.kt
```diff
@@ -0,0 +1,94 @@
...
+ * same declaration at once, so that a single value is always sufficient:
+ * - **Access control** ([PUBLIC], [PROTECTED], [PRIVATE]) restricts access relative to the
+ *   *declaring [Record]* and only applies to record members.
+ * - **Linkage** ([INTERNAL]) restricts access relative to the *translation unit* and only applies
+ *   to non-member (file- or namespace-scope) declarations.
+ *
+ * Static-vs-instance "membership" is an orthogonal axis and is intentionally *not* modeled here; it
+ * is captured by [de.fraunhofer.aisec.cpg.graph.declarations.ValueDeclaration.isStatic].
+ */
+enum class Visibility {
```
KuechA, line 70: Check if this list works for all the supported languages: C, C++, Java, Rust, Go, Ruby, TypeScript, JavaScript

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
```diff
@@ -327,6 +328,82 @@ interface HasRedeclarations : LanguageTrait {
     fun mergeRedeclaration(existing: Declaration, incoming: Declaration)
 }
 
+/**
+ * The syntactic position in which a declaration appears. The *meaning* of a storage-class or
+ * visibility keyword such as C/C++'s `static` depends entirely on this context, so a frontend must
+ * classify a declaration before asking a [HasKeywordSemantics] language to interpret its keywords.
+ */
+enum class DeclarationContext {
```
KuechA: I'm wondering how this is aligned with the `Scope` which should somehow already model visibility/availability.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -327,6 +332,28 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
+     * with [Visibility.INTERNAL] (in C/C++ a file-scope `static`, see the frontend's
+     * `HasKeywordSemantics` mapping) is confined to its own translation unit, so it must not be
+     * resolved from a reference in a different one. This is what makes cross-translation-unit
+     * lookups of `static` globals and functions fail, as the language semantics require.
+     *
+     * Candidates without internal linkage are always kept, so languages that never assign
+     * [Visibility.INTERNAL] are completely unaffected. As internal linkage is comparatively rare,
+     * we avoid resolving [ref]'s translation unit unless at least one candidate actually has it.
+     */
+    private fun Set<Declaration>.withoutForeignInternalLinkage(ref: Reference): Set<Declaration> {
```
KuechA: Can we find a better name for this method?

I'm also wondering if "internal" may not cover all declarations which aren't visible for all the programming languages...

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -527,9 +554,87 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
+    }
+
+    /**
+     * Whether this member declaration is accessible from the record [from] in which the access
+     * occurs. A [Visibility.PRIVATE] member is only accessible from within its own declaring
+     * record, a [Visibility.PROTECTED] member additionally from records that (transitively) inherit
+     * from the declaring one. Any other visibility (including [Visibility.UNKNOWN]), and any
+     * language without the [HasAccessControl] trait, imposes no restriction.
+     *
+     * This deliberately models only the two most common access relationships. Access that is
```
KuechA: I'm not really sure what this means. Can you give examples in practice where this is relevant? Why this limit to the two most common access relationships and not any number?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -527,9 +554,87 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
+            else -> true
+        }
     }
 
+    /**
+     * The [Record] that declares this member. For a [Method] this is its [Method.recordDeclaration]
+     * (which is also set for out-of-line definitions), for any other member it is the closest
+     * enclosing [Record] in the AST. Returns `null` for non-members.
+     */
+    private val Declaration.declaringRecord: Record?
```
KuechA, line 647: Can't we get this from the scope? or astParent?

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/CLanguage.kt
```diff
@@ -132,6 +143,35 @@ open class CLanguage :
...
+     * - on a record member ([DeclarationContext.RECORD]) it makes the member *static*, i.e. bound
+     *   to the record itself rather than to an instance;
+     * - inside a function body ([DeclarationContext.LOCAL]) it only affects storage duration, which
+     *   is irrelevant to symbol resolution, so it carries no canonical semantics.
+     *
+     * The access specifiers `public`/`protected`/`private` map onto the corresponding [Visibility]
+     * regardless of context (they only ever occur on record members). Any other keyword yields
+     * empty [KeywordSemantics].
+     */
+    override fun interpretKeyword(keyword: String, context: DeclarationContext): KeywordSemantics {
```
KuechA: Shouldn't we have this in almost all languages? Why don't we have a different implementation for `CLanguage` and `CXXLanguage`?

cpg-language-cxx/src/test/resources/c/static_semantics.c
```diff
@@ -0,0 +1,14 @@
+// In C, a file-scope `static` gives the variable/function internal linkage (TU-local). The
+// non-static counterparts have external linkage.
+static int internalGlobal = 1;
+int exportedGlobal = 2;
+
+static int internalFunction() { return internalGlobal; }
```
KuechA, line 6: it would probably be useful to have a second file with a function `internalFunction` to see if it resolves properly.

KuechA (commented):
I think that it should be generalized further to the other languages.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
oxisto: This sounds exactly like a Scope

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
```diff
@@ -327,6 +328,90 @@ interface HasRedeclarations : LanguageTrait {
...
+
+/**
+ * The canonical semantics that a single declaration keyword implies, as returned by
+ * [HasKeywordSemantics.interpretKeyword].
+ *
+ * Every property is nullable and defaults to `null`, meaning "this keyword says nothing about this
+ * axis". This lets a frontend fold the results of several keywords together (e.g. with [merge])
+ * without one keyword's silence overwriting another keyword's opinion.
+ */
+data class KeywordSemantics(
```
oxisto: I don't really get the idea behind that class

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
```diff
@@ -327,6 +328,90 @@ interface HasRedeclarations : LanguageTrait {
...
+ * static storage duration inside a function, and a class-level member inside a record.
+ *
+ * Frontends keep the raw keyword in [HasModifiers.modifiers] and additionally call
+ * [interpretKeyword] to obtain the canonical [KeywordSemantics], which they then project onto
+ * [HasVisibility.visibility] and
+ * [de.fraunhofer.aisec.cpg.graph.declarations.ValueDeclaration.isStatic]. Centralizing the mapping
+ * behind this trait keeps the per-language knowledge of "what does this keyword mean in this
+ * position" in a single place, instead of scattering it across the frontend's handlers.
+ */
+interface HasKeywordSemantics : LanguageTrait {
```
oxisto: I mean doesn't basically every language have that? I do like the concept of somehow "parsing" the modifiers into something with semantics. But is it just visibility?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
```diff
@@ -327,6 +328,90 @@ interface HasRedeclarations : LanguageTrait {
...
+ * restrict from where a member may be accessed based on an access specifier such as C++'s `public`
+ * / `protected` / `private`.
+ *
+ * When a language declares this trait, the [SymbolResolver] takes the canonical
+ * [de.fraunhofer.aisec.cpg.graph.HasVisibility.visibility] of a member into account while resolving
+ * a member by name, preferring members that are actually accessible from the point of access over
+ * inaccessible ones (see [SymbolResolver.resolveMemberByName]). Languages without this trait are
+ * left completely unaffected and keep resolving members regardless of visibility.
+ */
+interface HasAccessControl : LanguageTrait
```
oxisto: How is acccess control different from HasVisbility?

oxisto (commented):
You may want to prompt Claude to take another pass which concepts really should be designed from scratch and which existing ones should be (re)-used

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
KuechA: Agreed — `DeclarationContext` was a redundant re-modeling of `Scope`. Removed the enum entirely. `applyModifiers` now receives the actual `Scope?`, and the frontends match on `RecordScope` / `GlobalScope` / `NamespaceScope` directly.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
KuechA: Removed. It was a transient carrier for a computed result that we then copied onto the declaration anyway. The frontend now sets `Declaration.visibility` / `ValueDeclaration.isStatic` in place, so the intermediate class is gone.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
KuechA: Two good points. (1) Since nearly every language needs it, it's no longer a niche trait — it's now a defaulted no-op method `applyModifiers` on the base `Language`, which each frontend overrides. (2) It's not just visibility: it also derives `isStatic` (and is the natural home for any further modifier-derived semantics), so keeping it as a single "project the raw modifiers onto canonical properties" hook rather than a visibility-only mapping.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/LanguageTraits.kt
KuechA: This was the core confusion, now resolved by collapsing the two:
- Recording visibility is universal → `visibility` is now a plain property on every `Declaration` (no interface).
- Enforcing record-level access control is the language-specific part → kept as a trait, renamed `HasAccessControl` → `HasVisibilityModifiers`, whose KDoc explicitly draws the distinction: every declaration records a visibility, but only a language with this trait has the `SymbolResolver` act on `private`/`protected` during resolution. (e.g. a language that records `__x` as PRIVATE purely for documentation records visibility but must not declare the trait.)

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
KuechA: For lexically-nested members, yes — and that's what the default path does (`firstParentOrNull<Record>()`, which walks `astParent`; the `RecordScope` would give the same answer). The reason it isn't only that: an out-of-line C++ definition (`void C::foo() {}`) has its AST parent/scope in the enclosing namespace, not the record, so we prefer `Method.recordDeclaration` which points at the record even then. Documented this in the `declaringRecord` KDoc.

cpg-language-cxx/src/test/resources/c/static_semantics.c
KuechA: Added `c/internal_linkage/a.c` + `b.c` and `CXXInternalLinkageTest`, which asserts each TU's static symbol resolves only within its own TU while the externally-linked symbol still resolves across TUs.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -527,9 +565,107 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
+        get() = (this as? Method)?.recordDeclaration ?: firstParentOrNull<Record>()
+
+    /**
+     * All (transitively) inherited super-records of this [Record] — the transitive closure of
+     * [Record.superTypeDeclarations] (which only holds the *direct* super-records), excluding the
+     * record itself.
+     */
+    private val Record.allSuperclasses: Set<Record>
```
oxisto: I think we have this already somewhere @copilot can you have a look where/if?

oxisto (approved):
looks good overall, some small nit

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
oxisto: You should have only told me, not modify code but ok ...

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/JavaImportResolver.kt
```diff
@@ -144,7 +142,13 @@ open class JavaImportResolver(ctx: TranslationContext) : ComponentPass(ctx) {
         if (result.isEmpty()) {
             // the target might be a field or a method, we don't know. Thus, we need to create both
             val targetField =
-                newField(name, UnknownType.getUnknownType(base.language), setOf(), null, false)
+                newField(
+                    name,
+                    UnknownType.getUnknownType(base.language),
+                    setOf(),
+                    initializer = null,
+                    implicitInitializerAllowed = false,
```
KuechA, line 150:
```suggestion
                    implicitInitializerAllowed = false,
+                   isStatic = true,
```

cpg-language-java/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/JavaImportResolver.kt
```diff
@@ -144,7 +142,13 @@ open class JavaImportResolver(ctx: TranslationContext) : ComponentPass(ctx) {
         if (result.isEmpty()) {
             // the target might be a field or a method, we don't know. Thus, we need to create both
             val targetField =
-                newField(name, UnknownType.getUnknownType(base.language), setOf(), null, false)
+                newField(
+                    name,
+                    UnknownType.getUnknownType(base.language),
+                    setOf(),
```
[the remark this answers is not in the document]
KuechA: done

PR #2850, opened 2026-08-02
Rewrite followXUntilHit for correct interprocedural loop detection

### LINKED ISSUE #2375: Loop detection of `followXUntilHit` does not work
The look detection of `followXUntilHit` does not work properly. This problem should occur if there are alternating ways to loop on a call expression.

@morbitzer can you maybe provide an example?

KuechA:
## Problem

`followXUntilHit` — the worklist engine behind all the `followXEdgesUntilHit` traversals (EOG/DFG/CDG/PDG, context- and field-sensitive, MAY/MUST, forward/backward) — failed to terminate on **interprocedural data-flow recursion**. PR #2794 tried to address this but didn't fully fix it.

The root cause: the old loop guard only detected a repeated `(node, callStack)` state on the *current path*. That never fires when:

- recursion goes through **fresh call sites** (each level pushes a distinct `Call`, so no `(node, callStack)` state ever repeats), or
- the traversal crosses a call boundary via a **DFG interprocedural edge**, where the advancing node is a parameter/argument/return — not the `Call` sitting on the stack.

Both cases produced unbounded re-traversal / hangs.

## Approach

The engine is split into two regimes that share a single loop-cut decision:

- **MAY** (`findAllPossiblePaths = false`) → visit-once **BFS** over the precise `(node, callStack, indexStack)` state with parent pointers. Each state is expanded once and reached via its shortest path first, yielding the shortest witness the MAY consumers actually use (`fulfilled.any` / `minByOrNull { size }`). This also removes the old **O(W²)** per-step witness scan and keeps only parent pointers in memory (full paths are rebuilt lazily for the few recorded hits).
- **MUST** (`findAllPossiblePaths = true`) → depth-first **faithful enumeration** of every maximal path, so `failed` holds exactly the maximal paths that miss the predicate (inevitability holds iff `failed` is empty). Acyclic fulfilled/failed counts are preserved.

Loop/recursion cutting is unified in `contextExplosion`, which additionally detects a repeated `Call` **by object identity** on the call stack (catching fresh-call-site recursion regardless of which node type advances) and applies hard call/index-stack backstops against unbounded growth.

Also in this PR:
- Fix **shared-`Context` mutation** across sibling edges in the PDG/CDG manual callbacks by cloning the `Context` per edge.
- Remove the **dead `continueAfterHit`** parameter (declared but never read by any code path, wrapper, or caller).
- Add a **benchmark suite** (`FollowXTraversalBenchmark`) covering linear chains, diamonds, self-loops, and repeating/fresh interprocedural recursion.

## Results

| scenario | before | after |
|---|---|---|
| `linearChain(4000)` MAY | 453 ms | **10 ms** |
| `recursion-fresh` MAY | ~7.3 s (blowup) | **~280 ms**, terminates |
| `recursion-fresh` MUST | ~6.0 s (hang) | **~150 ms**, terminates |
| `recursion-repeating` | — | 0–3 ms (identity cut fires immediately) |

`diamondChain(14)` MUST remains bounded by `maxSteps` — that's inherent to faithful MUST path enumeration (2¹⁴ paths), unchanged from baseline.

`cpg-core` suite: **269 tests, 0 failures**. `ShortcutsTest` exact-count gate and the benchmark sanity assertions pass.


KuechA:
Follow-up commit `cf45dd5` adds two things on top of the traversal rewrite:

**1. `continueAfterHit` for MAY reachability early-exit**
Re-introduced with new semantics. For a MAY analysis (`findAllPossiblePaths = false`), `continueAfterHit = false` turns the traversal into a pure reachability query: it stops at the *first* reached target and returns a single shortest witness. `failed` is left empty in this mode, and the flag is ignored for a MUST analysis. Defaults to `true` (unchanged behaviour) and is threaded through all `follow*UntilHit` wrappers.

Micro-benchmark impact (`multiTargetFan(2000)`): two reachable targets, one immediate and one behind a 2000-node chain.

| config | steps | witnesses |
|---|---|---|
| `continueAfterHit = true` (MAY-all) | 2002 | 2 |
| `continueAfterHit = false` (MAY-1st) | 1 | 1 |

**2. prev-PDG traversal end/start fix**
`followPrevPDGUntilHit` mapped every incoming prev-PDG edge to `edge.end` (the node itself), so a backward intraprocedural traversal never made progress. It now uses `edge.start` for regular PDG edges and only `edge.end` for the interprocedural `Usage` edges — mirroring the CDG fix in #2816. Added a regression test in `ProgramDependenceTest`.

Full `cpg-core` suite: **270 passed, 0 failed** (1 skipped).

oxisto (commented):
This PR does a LOT OF testing, but it does not actually seem to test the right things, if the patch coverage is still about 50 % only?

cpg-core/src/performanceTest/kotlin/de/fraunhofer/aisec/cpg/graph/FollowXTraversalBenchmark.kt
```diff
@@ -0,0 +1,517 @@
...
+ *    true` the current engine does not terminate.
+ * 2. **Exponential path enumeration** – a chain of diamonds (or a loop containing a branch) has an
+ *    exponential number of distinct paths. `findAllPossiblePaths = true` materializes all of them.
+ *
+ * Every scenario is guarded by a [Budget] that aborts the traversal after a fixed number of
+ * next-step evaluations, so a pathological run reports "BUDGET EXCEEDED" in bounded time/memory
+ * instead of hanging the JVM. The number of next-step evaluations ("steps") is the primary
+ * efficiency metric; wall-clock time is reported as a secondary, machine-dependent signal.
+ */
+class FollowXTraversalBenchmark {
```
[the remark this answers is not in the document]
KuechA: Tagged and moved

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSourcesTest.kt
```diff
@@ -0,0 +1,329 @@
...
+            val ifds =
+                q.ifdsReachingSourcesDFG(
+                    direction = Backward(GraphToFollow.DFG),
+                    predicate = predicate,
+                )
+            // IFDS with summary edges is complete on recursion.
+            assertEquals(setOf<Node>(s), ifds, "IFDS(k=inf) must find the complete set {s}")
+
+            // Sanity: IFDS is a strict superset of the (empty) engine result here.
+            assert(ifds.containsAll(engine) && ifds.size > engine.size)
```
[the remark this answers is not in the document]
KuechA: replaced, also checked for other occurrences

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -1044,195 +1011,340 @@ fun Node.followPrevPDGUntilHit(
...
  */
-fun Node.followXUntilHit(
-    x:
-        (
-            Node, Context, List<Triple<Node, Edge<Node>?, Context>>, MutableSet<NodePath>,
-        ) -> Collection<Triple<Node, Edge<Node>, Context>>,
-    collectFailedPaths: Boolean = true,
-    findAllPossiblePaths: Boolean = true,
-    continueAfterHit: Boolean = false,
+fun Node.followDFGEdgesUntilHitNodes(
```
KuechA, line 1038: As this is a public-facing API, I'd prefer a description of the parameters, and what the function is used for (not so much all the technical details of the implementation). I'd move this to comments in the function body/code. Same for the other functions

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -1044,195 +1011,340 @@ fun Node.followPrevPDGUntilHit(
...
+                sensitivities = sensitivities,
             )
+        },
+        earlyTermination = earlyTermination,
+        predicate = predicate,
+    )
+}
+
+/**
+ * Path-free MAY variant of [followPrevFullDFGEdgesUntilHit]: returns the set of frontier nodes
```
KuechA: What's a "frontier" node? I'd just stay with "node".

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/GraphTraversal.kt
```diff
@@ -0,0 +1,504 @@
...
+    val loopingPaths: MutableSet<NodePath> = ConcurrentHashMap.newKeySet()
+
+    // First check if the current node satisfies the predicate.
+    // If it does, we consider this path fulfilled and skip further traversal.
+    if (predicate(this)) {
+        fulfilledPaths.add(NodePath(mutableListOf(this), emptyList()).addAssumptionDependence(this))
+        return FulfilledAndFailedPaths(fulfilledPaths, failedPaths)
+    }
+
+    if (!findAllPossiblePaths) {
```
KuechA, line 121: Since the two branches (BFS vs DFS) are quite large, why don't we make two helper/utility functions out of this?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
```diff
@@ -1044,195 +1011,340 @@ fun Node.followPrevPDGUntilHit(
...
+    // regime that [ifdsReachingSources] covers. Anything else (FieldSensitive, Intraprocedural, a
+    // custom earlyTermination, or a pre-seeded call stack) falls through to the legacy engine.
+    if (
+        (direction is Forward || direction is Backward) &&
+            sensitivities.toSet() == setOf<AnalysisSensitivity>(ContextSensitive) &&
+            scope is Interprocedural &&
+            earlyTermination === noEarlyTermination &&
+            ctx.callStack.isEmpty()
+    ) {
+        return this.ifdsReachingSources(direction, k = Int.MAX_VALUE, predicate = predicate)
```
KuechA, line 1064: I'm wondering why this is the only location were we do this. Can't we also do the same for EOG? Why only when we collect nodes only and not for the paths, too?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
```diff
@@ -0,0 +1,554 @@
...
+    val start: Node,
+    val direction: AnalysisDirection,
+    val k: Int,
+    val predicate: (Node) -> Boolean,
+    /**
+     * Optional cross-query cache of predicate-independent pure summaries; see [IfdsSummaryCache].
+     */
+    val cache: IfdsSummaryCache? = null,
+) {
+    private val backward = direction is Backward
```
KuechA, line 199: This doesn't work if we want to analyze forwards and backwards at the same time. Not sure if this is a problem though because it's not fully implemented.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
```diff
@@ -0,0 +1,554 @@
...
+            }
+        } else {
+            IntraStep(target)
+        }
+    }
+}
+
+private fun eogSteps(node: Node, backward: Boolean): List<IfdsStep> {
+    if (backward) {
+        if (node is Call && node.invokes.isNotEmpty()) {
```
KuechA: The backward and forward steps should probably be moved into two separate helpers to avoid such large functions which effectively do two different things

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
```diff
@@ -0,0 +1,554 @@
...
+ *    summaries ("enter this callee, pop this token, resume at this node") are stable graph facts
+ *    that can be tabulated once and reused by every query.
+ * 2. The only thing a query's [predicate] changes is where the frontier stops. A callee whose
+ *    transitive callee-subgraph contains **no sink** can therefore never trigger frontier pruning,
+ *    so its pure summary is exactly what any query would compute — we replay it and skip the
+ *    callee.
+ *
+ * A callee is **dirty** if its function (transitively, up the call graph) contains a sink. Sinks
+ * are registered by the owning pass via [markSink] as overlays / state entries are added to nodes
+ * (that is what a [predicate] like the CBOM `getNodesBasedOnDfg` tests). Dirtiness is monotonic
```
KuechA:
```suggestion
* . Dirtiness is monotonic
```

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
```diff
@@ -0,0 +1,554 @@
...
+        // summary reuse, exactly as in the query solver).
+        pureComplete.addAll(seededThisDrain)
+    }
+
+    private fun seedPure(ctx: Node?, node: Node) {
+        // No predicate: the pure tabulation treats no node as a sink.
+        if (pureVisited.add(PathEdge(ctx, node))) pureWorklist.addLast(PathEdge(ctx, node))
+    }
+
+    private fun expandPure(ctx: Node?, node: Node) {
```
KuechA: this function needs some comment. Keep it short.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
```diff
@@ -0,0 +1,554 @@
...
+ * one-time scan of pre-existing overlays). Any query that finds no cache runs the exact uncached
+ * tabulation.
+ *
+ * Instances are used from a single pass that executes its queries sequentially; no internal locking
+ * is required.
+ *
+ * @param direction The traversal this cache's pure summaries were tabulated for. A query with a
+ *   different [AnalysisDirection] shape must not reuse them (enforced in [ifdsReachingSources]).
+ */
+class IfdsSummaryCache(val direction: AnalysisDirection) {
```
KuechA: I'd suggest to move this into an own file

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: Simplified the KDoc on `followDFGEdgesUntilHitNodes`/`followEOGEdgesUntilHitNodes` to just describe params/purpose, moved the delegation rationale into a code comment above the guard.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: Removed "frontier" from the public docs on the `...UntilHitNodes` wrappers (8 occurrences) and just say "node"/"PDG node"/"CDG node" now.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/GraphTraversal.kt
KuechA: Done, split into `eogStepsBackward`/`eogStepsForward` helpers, dispatched from `eogSteps`. GraphTraversal.kt's BFS/DFS branches are a separate, orthogonal function (the legacy engine's worklist strategy), not touched by this reply -- let me know if you meant that one too and I'll split it as well.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
KuechA: Confirmed: a single `IfdsReachingSourcesSolver`/`IfdsSummaryCache` instance analyzes exactly one `AnalysisDirection`; mixed forward+backward in one query isn't supported and would need two separate solves. Added a code comment calling this out explicitly. No current caller needs it (each query picks one direction), so I left it as a documented limitation rather than implementing it.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
KuechA: Done, split `eogSteps` into `eogStepsBackward`/`eogStepsForward`.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
KuechA: Applied, thanks.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
KuechA: Added a one-line comment on `expandPure`.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/IfdsReachingSources.kt
KuechA: Moved `IfdsSummaryCache` (and only it) into its own `IfdsSummaryCache.kt`. It shares some private tabulation types/helpers (`PathEdge`, `SummaryRecord`, `CallerRecord`, `TokenKey`, `IfdsStep`+subtypes, `ifdsStepsOf`, `ifdsTokensMatch`) with the solver in `IfdsReachingSources.kt`, so those became `internal` instead of file-private.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Extensions.kt
KuechA: EOG delegation is already implemented too, in `followEOGEdgesUntilHitNodes` right below this (same guard shape, minus the `ctx`/start-stack check since EOG traversal always starts from an empty stack).

On paths: I deliberately scoped this to the node-set (MAY) variants only. The IFDS solver reuses a callee's summary edge across every caller that pushes into it -- that's exactly what makes it recursion-complete and fast -- but a summary only records "popping this token resumes at this node", not the sub-path taken to get there. Reconstructing one concrete witness path per hit would mean every tabulated path edge and summary would additionally need to carry provenance (a predecessor pointer / partial-path fragment), and summaries would need to splice the caller-specific prefix back in on each reuse (the classical IFDS "path edge with witness" extension, e.g. as done in Bodden's IDE-based path reconstruction). That's a real chunk of extra bookkeeping with its own correctness argument, so I left `followDFGEdgesUntilHit`/`followEOGEdgesUntilHit` (the path-collecting variants) on the legacy engine for now and added a code comment at the guard explaining this. Happy to pick it up as a follow-up if we need it.

konradweiss (approved):
LGTM

PR #2854, opened 2026-08-05
Migrate to koog

maximiliankaul:
# Summary                                                                         
                                                                                   
Migrates `cpg-ai`'s LLM/agent orchestration from a hand-rolled client stack (GeminiClient/OpenAiClient, custom skill loader, manual tool-calling loop) onto [Koog](https://github.com/JetBrains/koog), and generalizes the resulting API so `cpg-ai` is a clean, reusable module for any consumer. All changes are contained to `cpg-ai` plus two small `cpg-core` concept files (`GenericLLMConcept`/`GenericLLMOperation`); no frontend/language changes are part of this PR.                                     
                                                                                   
## Koog migration                                                                     
                                                                                   
- Replace the custom GeminiClient/OpenAiClient HTTP clients and hand-rolled tool-calling loop with Koog's AIAgent/ToolRegistry/MCP integration (ChatService.kt rewrite).
- Adopt Koog ChatMemory (keyed by sessionId) so tool-call/tool-result history survives across calls instead of being lossily re-serialized as plain text each turn.
- Streaming responses, RetryingLLMClient for transient provider failures, structured completion (TaskStatus) instead of re-parsing persisted output, concurrent execution of read-only tool calls in a turn.
- Bump Koog 1.1.1 → 1.2.0.                                                         
- Migrate skill activation from a hand-rolled SkillLoader/activate_skill tool to Koog's native Agent Skills module (discoverSkills), using Koog's generic file tools jailed to .agents/skills/ (SkillFileTools.kt,
  tested against ../-traversal escapes).                                           
                                                                                   
## Context-window robustness                                                          
                                                                                   
- Fact-retrieval history compression (FactRetrievalHistoryCompressionStrategy) plus a token-based compression trigger, not just a message-count threshold.
- A single-oversized-tool-result truncation safety valve (truncateForLlm) — compression can only shrink older turns, so this is the only mechanism that can catch one new result alone blowing the context budget.
- Pagination for the bulk cpg_list_* MCP tools (default cap 20, limit/offset).  
- cpg_list_functions now returns a minimal nodeId + signature index by default instead of the full FunctionInfo (parameters/callees/file-line detail per function), which alone could exceed a model's context window on a large codebase. cpg_get_node remains the way to fetch full details for a specific function.
                                                                                   
## Public API cleanup                                                                 
                                                                                   
- ChatService's history-compression concepts were hardcoded around one consumer's specific skill vocabulary (baked-in references to two named skills). They're now an injectable constructor parameter (List<HistoryCompressionConcept>, a plain DTO — Koog's own Concept/FactType stay implementation-scoped, consistent with koog-agents itself) with a small domain-agnostic default; a caller with its own task vocabulary passes its own list.                                                  
- Generic concepts/operations tool consolidation: NodeReference and fixed-value support, DFG-connection + mutation-locking (GraphMutationLock) ported in from being externally duplicated, LLMProperty.description made required with schema-driven backfill.                                       
- Lenient defaults for optional LLM-supplied schema fields (LLMPropertyDescription.type, LLMOperationDescription.properties) instead of hard-crashing the whole tool call on omission.
- Authorization header redacted from OpenAI-compatible client debug logging.       
- Model-id fallback (chatLlm.model.id) when Koog leaves ResponseMetaInfo.modelId null.
                                                                                   
## Testing                                                                            
                                                                                   
- cpg-ai:test (unit) and cpg-ai:integrationTest (real CPG analysis end-to-end) both green, aside from one pre-existing, unrelated native-lib (UnsatisfiedLinkError) failure in CpgAnalyzeToolTest.testReanalyze, confirmed present on the unmodified base too.                                    
- New tests: SkillFileToolsTest (jail traversal), ChatMemorySpikeTest, LenientSchemaDefaultsTest, PaginatedTextContentTest, BulkListingCodeOmissionTest, CpgGenericConceptsToolUnitTest, GetFunctionsByNameToolTest.
- Verified against a real downstream consumer (DUST) via a full publish-to-mavenLocal + compile + package cycle, confirming no module-boundary breaks for either the cpg_list_functions shape change or the 
  compression-concepts API change.


cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/skills/SkillFileTools.kt
```diff
@@ -0,0 +1,115 @@
...
+ * Two things tuned here based on `log_skills_module`'s real-LLM test run (2026-09-08):
+ * - Says "read" only, not "list its directory and read" - the model was doing both per activation,
+ *   costing an extra tool round trip it doesn't need (the catalog already gives the exact file
+ *   path). Cut per-activation cost roughly in half.
+ * - States the access restriction explicitly, since [ReadFileTool]/[ListDirectoryTool]'s own
+ *   descriptions can't be changed (they're Koog's own non-`open` classes) and the rejection message
+ *   a jailed path actually gets back - "File not found: ... (ensure the path is absolute)" - reads
+ *   like the file doesn't exist rather than that it's out of bounds. Telling the model the boundary
+ *   upfront is cheaper than letting it find out by trial and error: the same real run also had the
+ *   model wander into listing `/home/cpg/dust`/`/home/cpg` and reading an unrelated `/tmp/...` path
```
maximiliankaul, line 98: remove `dust` reference

PR #2857, opened 2026-08-05
Model Python naming-convention visibility

KuechA:
## Summary

Wires the `:cpg-language-python` frontend into the canonical `Visibility` model. Python has **no enforced access control**; member visibility is purely a PEP 8 naming convention, so this is a **documented heuristic** driven by the member's (local) name.

### Mapping (`PythonLanguage.visibilityForName`)
- Leading **double** underscore that is **not** a trailing dunder (e.g. `__x`, but not `__init__`) — Python name-mangling, conventionally "private" -> `Visibility.PRIVATE`.
- **Single** leading underscore (e.g. `_x`) — conventionally "non-public" -> `Visibility.PROTECTED`.
- Everything else, including dunders such as `__init__` / `__str__` -> `Visibility.PUBLIC`.

The mapping is projected onto:
- **methods / constructors / operators** in `DeclarationHandler.handleFunctionDef` (record members only), and
- **fields** (both instance attributes `self.__x` and class attributes) in `PythonAddDeclarationsPass`.

Raw modifiers are left untouched. Free (non-member) functions keep the default `Visibility.UNKNOWN` — Python has no module-level access control that we model here.

### Judgement calls
- **`_name` -> PROTECTED (not PACKAGE):** since this mapping is applied primarily to record members, the record-relative axis (protected) is a closer fit than module-relative `PACKAGE`, and it matches how subclasses conventionally still use such members. Documented in a code comment.
- **No `HasAccessControl`:** Python does not *enforce* access, so the frontend deliberately does not declare `HasAccessControl`. It only records the conventional visibility so passes may read it, without the `SymbolResolver` ever dropping a resolution candidate.
- **`isStatic` left unchanged:** kept out of scope to avoid touching orthogonal behaviour; only visibility is modeled here.

## Tests
Added `VisibilityTest` (a pure unit test of `visibilityForName` plus a frontend test over a new `visibility.py` fixture) covering PUBLIC, PROTECTED, PRIVATE, dunder-is-public, and the UNKNOWN default for free functions. `./gradlew :cpg-language-python:test -PenablePythonFrontend=true` is green (116 tests, 0 failures, 1 pre-existing skip).

## Notes
This PR is agent-generated and stacked on `ak/static-c` (base branch of this PR).


KuechA:
## Fixer pass: applied review feedback

All four reviews landed at **approve_with_nits** (no blockers/majors). Summary of what each raised and how it was handled:

**oxisto (minor + nits)**
- *Applied* — Flattened the doubly-nested `?.let` with shadowed `it` in `PythonAddDeclarationsPass` field site to a single named block: `(language as? PythonLanguage)?.let { pythonLanguage -> (decl as? Field)?.visibility = ... }`. This also removes the mixed `it`/`field` naming.
- *Skipped* — `when` branch reordering (dunder-first). Explicitly optional/judgment call; kept the existing form which matches the CXX `return when` style and minimizes churn.
- *Note* — the two call sites now read more consistently; a full `is`-smart-cast at the pass site isn't possible since `language` is a custom-getter property (no smart cast), so the `as?` flatten is the tidiest option there.

**maximiliankaul (nits)**
- *Applied* — Same nested-`let` readability concern addressed by the flatten above (named lambda param).
- *Applied* — Extended the name-mangling comment in `visibilityForName` to state Python's actual rule (>=2 leading underscores, at most one trailing underscore), making `endsWith("__")` self-justifying (`__foo_` is mangled; `__init__` is not).
- *Skipped* — "s.name is already a localName" clarifying comment at the DeclarationHandler site. Low value; the KDoc on `visibilityForName` already documents the localName contract.

**adversarial-correctness (nits)**
- *Skipped* — Degenerate all-underscore names (`__`, `___`) mapping to PUBLIC. No realistic member construct misbehaves; left as-is.
- *Skipped* — SymbolResolver fast-path micro-cost. Confirmed behaviorally identical (Python is not `HasAccessControl`); not a correctness issue.

**language-behavior (nits)**
- *Skipped* — Both nits were affirmations (single-underscore→PROTECTED tradeoff and mangling-restricted-to-class-bodies are faithful). No change requested.

Formatting via `spotlessApply`; `:cpg-language-python:test` is green.

KuechA:
TODO: rename `HasVisibilityModifiers` to `HasVisibilityEnforcement`

oxisto (approved):
I think its fine since python does NOT enforce access controls, but the whole `HasVisibilityModifers` trait is a bit confusing now tbh

PR #2858, opened 2026-08-05
Model Java member and package-level visibility

KuechA:
## Summary

Wires the `:cpg-language-java` frontend into the canonical, language-agnostic `Visibility` model introduced on `ak/static-c`.

`JavaLanguage` now declares two traits:
- **`HasKeywordSemantics`** — `interpretKeyword` maps the access modifiers `public`/`protected`/`private` onto `Visibility.PUBLIC`/`PROTECTED`/`PRIVATE`, and `static` (on a record member) onto `KeywordSemantics(isStatic = true)`.
- **`HasAccessControl`** — Java genuinely enforces member access control, so the `SymbolResolver` can take member visibility into account.

`DeclarationHandler` derives the syntactic `DeclarationContext` from the active scope and, for fields, methods, constructors, and (nested) types, folds the present keywords into `KeywordSemantics` and projects the result onto `Declaration.visibility` and `ValueDeclaration.isStatic`. The raw, lossless keywords stay in `modifiers`.

### Visibility mapping

| Java surface | Canonical `Visibility` |
| --- | --- |
| `public` | `PUBLIC` |
| `protected` | `PROTECTED` |
| `private` | `PRIVATE` |
| *(no access modifier)* | `PACKAGE` |

`static` is orthogonal to visibility and is projected onto `isStatic` (not the visibility axis).

### Judgement calls

- **Package-private default (the high-value case).** `interpretKeyword` only ever sees keywords that are *present*, so it cannot express the "no modifier" default. The handler therefore maps a declaration whose folded visibility is still `UNKNOWN` to `Visibility.PACKAGE`. This is applied to fields, methods, constructors, nested types, and top-level types — a bare `class Foo` is `PACKAGE`, not `PUBLIC`.
- **Interface members.** Java implicitly makes interface methods/fields `public` regardless of the absence of a modifier. This is *not* special-cased here; such members currently fall through to the `PACKAGE` default. Interfaces are modeled as `"class"` records by the existing frontend, so distinguishing them would need extra plumbing; left as a follow-up and called out for reviewers.
- `handleConstructor` and `handleEnumeration` previously did not populate `modifiers` at all; this PR now captures their raw keywords as well so the projection has something to work from.

## Test status

Added `JavaVisibilityTest` plus a `src/test/resources/java/visibility/Visibility.java` fixture covering every produced value: `PUBLIC`/`PROTECTED`/`PRIVATE`/`PACKAGE` for fields, methods, and (nested) types, the `PACKAGE` default for a no-modifier constructor and a no-modifier top-level class, and `isStatic` for static members.

`./gradlew :cpg-language-java:test -PenableJavaFrontend=true` is green: 70 tests, 0 failures, 2 pre-existing skips.

---

This PR is agent-generated and stacked on `ak/static-c`. It touches only `:cpg-language-java`; `cpg-core` is untouched.


KuechA:
## Fixer: applied review feedback

Thanks for the four reviews. Summary of what was raised and what I did.

### Applied

**Interface members mislabeled PACKAGE instead of PUBLIC** (major — adversarial-correctness, language-behavior, maximiliankaul)
Interfaces are now modeled as records with `kind = "interface"` (the previously TODO'd case). `applyVisibility` detects an enclosing interface and defaults its members (fields, methods, nested types) to `PUBLIC` instead of `PACKAGE`, and marks interface fields `static` (JLS §9.3–§9.5). Added `Visibility.java` fixture + `testVisibility` asserting PUBLIC/static for interface members and PACKAGE for ordinary-class members.

**Local/anonymous types get PACKAGE though they have none** (minor — language-behavior)
`currentDeclarationContext` now emits `DeclarationContext.LOCAL` for function/block scopes, and `applyVisibility` skips the PACKAGE fallback for LOCAL declarations (leaves UNKNOWN). This also resolves maximilian's doc concern about LOCAL never being produced.

**Enum constants left UNKNOWN** (nit — adversarial-correctness, language-behavior)
`handleEnumConstant` now sets `PUBLIC` + `isStatic` (JLS §8.9.1).

**Cleanups** (oxisto, maximilian)
- Dropped the redundant `modifiers` parameter from `applyVisibility` (all call sites passed `declaration.modifiers`).
- Extracted a `NodeList<Modifier>.keywords` helper, removing the four duplicated `modifiers.map { it.keyword.asString() }.toSet()` copies.
- Switched the trait check to the `val language = language as? HasKeywordSemantics ?: return` idiom (removes the opaque self-assign + `!is` guard).
- Clarified KDoc on `interpretKeyword` and `currentDeclarationContext`; added a one-line note that `applyVisibility` is authoritative for member `isStatic`.

**HasAccessControl regression check** (minor — adversarial-correctness)
Ran the full `:cpg-language-java` test suite — all green, no member-resolution regressions.

### Skipped

- **Extend `isAccessibleFrom` for Java inner-class→enclosing-private access** (minor — adversarial-correctness): out of scope for this PR and guarded by the conservative "never drop the last candidate" rule; noted as a possible follow-up.
- **Rename keyword string constants to avoid sharing names with `Visibility`** (nit — maximilian): low value; the KDoc already labels them as source keywords.
- The "add a TODO documenting the interface limitation" and "comment the `val language = language` smart-cast" nits are moot — the limitation is fixed and the idiom was replaced.

oxisto (approved):

oxisto (approved):

PR #2859, opened 2026-08-05
Model Go exported/unexported (package) visibility

KuechA:
## Summary

Wires the Go frontend (`:cpg-language-go`) into the canonical, language-agnostic `Visibility` model introduced on `ak/static-c`.

Go has no access-control keywords: visibility is determined purely by **identifier casing** (https://go.dev/ref/spec#Exported_identifiers). The mapping we chose:

| Go concept | Canonical `Visibility` |
| --- | --- |
| Exported (first rune is an upper-case Unicode letter) | `PUBLIC` |
| Unexported (any other first rune, incl. lower-case / leading `_`) | `PACKAGE` |
| Blank identifier `_` and unnamed declarations | `UNKNOWN` (no restriction) |

The mapping lives in a small helper `GoLanguage.exportVisibility(identifier)`, which inspects the first full **code point** (so Unicode identifiers such as `Über` vs `über` are handled correctly) and special-cases the blank identifier. The declaration/specification handlers project the result onto `Declaration.visibility` for:

- top-level funcs and methods (`DeclarationHandler`)
- top-level vars and consts, incl. tuple deconstruction (`SpecificationHandler`)
- named struct/interface/other types (`SpecificationHandler`)
- struct fields (incl. embedded fields, whose promotion follows the type-name casing)
- interface methods

The raw name is left untouched; only the canonical `visibility` is additionally set.

## Judgement calls

- **`HasAccessControl` is intentionally NOT declared.** Go has no record-level `private`/`protected` access control, so unexported members map to module-scoped `PACKAGE`, not `PRIVATE`. Declaring `HasAccessControl` would make the `SymbolResolver` treat unexported members as access-controlled, which is wrong for Go.
- **`HasKeywordSemantics` is not used** either, because Go has no visibility keywords at all — a casing helper is the natural hook.
- **Local (block-scoped) variables** are left `UNKNOWN`: export semantics only apply to package-level declarations and struct/interface members.
- **Blank identifier `_`** and empty/unnamed declarations are left `UNKNOWN` since they carry no meaningful, referenceable visibility.

## Tests

Added `VisibilityTest` (2 tests, both green) plus fixture `src/test/resources/golang/visibility.go`:
- a unit test of the casing helper covering exported/unexported, leading `_`, blank identifier, empty string and Unicode (`Über`/`über`);
- an integration test asserting the produced `visibility` on funcs, vars, consts, unicode vars, struct types + fields, unexported types, methods, interface types + methods, and a local variable.

`./gradlew :cpg-language-go:spotlessApply` and `./gradlew :cpg-language-go:test` (with `-PenableGoFrontend=true`) both pass.

## Notes

This PR is agent-generated and stacked on `ak/static-c`. Only `:cpg-language-go` (plus its tests/resources) is touched; `cpg-core` is unchanged.


KuechA:
## Review follow-up applied

Four independent reviews ran on the Go visibility mapping. Summary of what was applied vs skipped.

**Applied (correctness/behavior):**
- **Type aliases now get a visibility** (adversarial + language-behavior): `handleTypeDef`'s `spec.assign != 0` branch (`type A = B`) previously created a `Typedef` without visibility. It now sets `typedef.visibility = GoLanguage.visibilityOf(spec.name.name)`, matching the other named-type branches. Fixture (`type ExportedAlias = int` / `type unexportedAlias = int`) and assertions added.
- **Top-level guard is now tested** (adversarial): the only local-scope case (`notExported := 1`) is an `*ast.AssignStmt` and never reached `handleValueSpec`, so the `if (topLevel)` guard was uncovered. Added a block-scoped classic `var Exported = 1` / `var unexported = 2` inside `local()` and assert both are `UNKNOWN` — this drives `DeclStmt -> handleValueSpec` with `topLevel = false` and protects the guard.

**Applied (style/understandability):**
- **`exportVisibility` -> `visibilityOf`** (oxisto): idiomatic single-expression `when`, `String` parameter instead of `CharSequence` (all call sites pass `String`), noun-style name. Companion object moved to the bottom of `GoLanguage`.
- **Comments** (maximiliankaul): documented why `Character.isUpperCase` is the right predicate (returns false for uncased scripts, matching Go's Lu-only rule); added cross-reference notes on the two mirrored var/const visibility blocks and a note on the embedded-interface (super-type) branch.

**Skipped (with reasons):**
- **`Character.isUpperCase` vs `Character.getType == UPPERCASE_LETTER`** (language-behavior nit): the reviewer notes current behavior is already correct for all valid Go identifiers (the two predicates agree on every letter that can start an identifier). Addressed with a clarifying comment instead of a functional change.
- **Trim per-site casing comments** (oxisto nit): conflicts with the maximiliankaul review, which explicitly values those comments for readability. Kept them.

Tests: `:cpg-language-go:test` is green.

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
```diff
@@ -236,6 +237,42 @@ open class GoLanguage :
...
+         *   [Visibility.PACKAGE] (module/package visibility, not tied to a record — Go has no
+         *   record-level `private`/`protected` access control).
+         *
+         * The blank identifier `_` as well as empty/unnamed declarations carry no meaningful
+         * visibility and are left as [Visibility.UNKNOWN] ("no restriction").
+         *
+         * This applies uniformly to top-level declarations (funcs, vars, consts, types) as well as
+         * to struct fields and methods.
+         */
+        fun exportVisibility(identifier: CharSequence): Visibility {
```
oxisto: why is this not using the `applyModifiers` of the language (trait)?

KuechA: Good catch — it should be, and it now is (b8746e1).

`GoLanguage` overrides `applyModifiers` and the handlers call it instead of assigning `visibility` themselves, so the hook is the single entry point here too.

Two things that came out of it beyond consistency:

* The guard I had open-coded was `frontend.scopeManager.currentScope is NameScope`, duplicated across both value-spec paths. Since both `NamespaceScope` and `RecordScope` are `NameScope`s, a single scope check in the override reproduces all eight previous call sites exactly (package-level funcs/vars/consts/types plus struct fields and interface methods), and the duplicated `topLevel` guard is gone.
* That guard turned out to be untested: the only local case in the fixture (`notExported := 1`) is an `*ast.AssignStmt` and never reaches `handleValueSpec`. I added a block-scoped classic `var` so the non-package path is actually exercised.

One caveat worth flagging, since it is why this was not using the hook in the first place: `applyModifiers` was documented as projecting *the raw `Declaration.modifiers`*, and Go has no visibility keywords at all — the Go frontend even uses `modifiers` for an unrelated purpose (marking embedded struct fields). So I broadened the contract in `Language.applyModifiers` to say implementations may derive the canonical visibility from anything about the declaration, the identifier included, and that frontends should not set `visibility` themselves. Happy to rename the hook instead if you would rather it stayed strictly about modifiers, though that would now touch the already-merged Ruby override too.

The same critique applied to Python, which derived visibility from the PEP 8 naming convention the same way — that is fixed in #2857.

oxisto (approved):

cpg-language-go/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/golang/GoLanguage.kt
```diff
@@ -236,6 +240,24 @@ open class GoLanguage :
         return CastNotPossible
     }
 
+    /**
```
oxisto, line 244: Oh wait do we need to extend HasVisbilityWhatever now?

KuechA: Yes... I added it back

PR #2860, opened 2026-08-05
Model TypeScript/JavaScript member visibility

KuechA:
## Summary

Wires the `:cpg-language-typescript` frontend (which hosts **both** `TypeScriptLanguage` and `JavaScriptLanguage`) into the canonical, language-agnostic `Visibility` model, following the CXX reference pattern (`CLanguage`/`CPPLanguage` + `DeclarationHandler`).

## Mapping

| Surface syntax | Language | `Visibility` | notes |
|---|---|---|---|
| `public member` | TypeScript | `PUBLIC` | |
| `protected member` | TypeScript | `PROTECTED` | |
| `private member` (compile-time) | TypeScript | `PRIVATE` | |
| `#name` member (hard/runtime private) | TS + JS | `PRIVATE` | identified by a `PrivateIdentifier` name node, not a keyword |
| `static member` | TS + JS | (visibility unchanged) | sets `ValueDeclaration.isStatic = true` |
| any other record member | TS + JS | `PUBLIC` | class members are public by default in both languages |

The raw keyword spellings (`public`/`protected`/`private`/`static`, and the synthetic `#private`) are additionally kept losslessly in `HasModifiers.modifiers`.

## Trait wiring / judgement calls

- `JavaScriptLanguage` (base) implements **`HasKeywordSemantics`** (for `static`) and **`HasAccessControl`** — the latter because `#private` is genuine, runtime-enforced access control, even though JavaScript has no access-modifier *keywords*.
- `TypeScriptLanguage` (subclass) overrides `interpretKeyword` to add the compile-time `public`/`protected`/`private` specifiers, delegating everything else to the JavaScript interpretation (mirroring `CPPLanguage` extending `CLanguage`).
- `#private` members currently surface with an empty local name (the frontend's `getIdentifierName` only reads `Identifier`, not `PrivateIdentifier`); this pre-existing limitation is out of scope here, so the tests locate them via their `#private` modifier. Visibility itself is set correctly.
- The `DeclarationHandler` derives the `DeclarationContext` from the active scope and projects `KeywordSemantics` onto `Declaration.visibility` / `ValueDeclaration.isStatic` for fields, methods and constructors, before descending into the member's own scope. `cpg-core` is untouched.

## Tests

Adds `VisibilityTest` with `visibility.ts` / `visibility.js` fixtures covering every produced value (`PUBLIC`, `PROTECTED`, `PRIVATE`, `static`, hard-`#private`, and the public-by-default case) for both fields and methods, in both languages.

Status: `./gradlew :cpg-language-typescript:test` is **green** (3 new tests, existing tests unaffected).

---

This PR is agent-generated and is stacked on `ak/static-c`.


KuechA:
## Fixer: applied review feedback

Four reviews ran (adversarial-correctness, language-behavior, oxisto, maximiliankaul), all `approve_with_nits`. Summary of what each raised and what I did:

**Applied**
- **oxisto (minor): `handleModifiers` imperatively shaped.** Refactored to collect all applicable keyword spellings (explicit `*Keyword` children + synthetic `#private`) into one list, append in a single statement, and `fold` the semantics — no `var`, no duplicated append/merge, and the early-continue disappears. Also adopted the idiomatic `frontend.language as? HasKeywordSemantics ?: return` (oxisto nit).
- **adversarial + language-behavior (minor): TS parameter properties not mapped.** This is a pre-existing limitation (the frontend never modeled `constructor(private x)` as a field). Documented it explicitly as out of scope in `handleParameter` rather than partially implementing, since full support requires synthesizing a `Field` on the enclosing record.
- **adversarial + language-behavior (nits): missing test coverage.** Added `private static` field/method (fold precedence → PRIVATE + isStatic), `static #x` (hard-private + static), and a `private constructor()` to `visibility.ts`/`visibility.js` with assertions.
- **maximiliankaul (nits): comments.** Clarified why the `is Method` gate also covers constructors (`Constructor extends Method`), why the PUBLIC default guards on `UNKNOWN`, the JS `static` context-gating rationale, and noted that the keyword consts are raw *spellings* (not `Visibility` values).
- **oxisto (nit): duplicated test helpers.** Unified the per-Field/per-Method `byName`/`hardPrivate` helpers into generic `Iterable<T : Declaration>` extensions.

**Skipped**
- **language-behavior (nit): ES module export/non-export visibility.** Out of scope — the task deliberately covers only record-member visibility; module-level export visibility is intentionally unmodeled.
- **maximiliankaul (nit): rename generic `PUBLIC`/`PRIVATE`/… consts.** Renaming is broader than this PR's scope; addressed the confusion concern via a KDoc note clarifying these are raw modifier spellings.
- **adversarial (nit): dedicated resolution test proving `HasAccessControl` narrows candidates.** The behavior lives in cpg-core `SymbolResolver.onlyAccessibleFrom`; a frontend-level resolution test would be brittle and duplicative. Ran the full `:cpg-language-typescript` suite — no existing resolution assertions shifted.

`:cpg-language-typescript:test` is green.

cpg-language-typescript/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/typescript/JavaScriptLanguage.kt
```diff
@@ -26,15 +26,78 @@
...
+
+/** The keyword marking a class-level (rather than per-instance) member. */
+const val STATIC = "static"
+
+/**
+ * A synthetic keyword standing in for a JavaScript/TypeScript *hard private* member, i.e. one whose
+ * name starts with `#` (a `PrivateIdentifier`). Unlike TypeScript's compile-time `private`, this is
+ * enforced at runtime and exists in plain JavaScript as well.
+ */
+const val HARD_PRIVATE = "#private"
```
oxisto: maybe only `#` as the modifier? the modifier I think should mostly be very close to the raw "code"

oxisto (approved):
small nit

cpg-language-typescript/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/typescript/JavaScriptLanguage.kt
KuechA: makes sense

PR #2862, opened 2026-08-05
Model Ruby method visibility

KuechA:
## Summary

Wires the `:cpg-language-ruby` frontend into the canonical, language-agnostic `Visibility` model (introduced on the base branch `ak/static-c`).

- `RubyLanguage` now declares the `HasAccessControl` marker trait (Ruby has genuine record-member access control) and `HasKeywordSemantics`, mapping the `public`/`protected`/`private` visibility modifiers onto `Visibility.PUBLIC` / `PROTECTED` / `PRIVATE`.
- `DeclarationHandler` now builds a `Record` for `class` definitions and, while walking the class body, tracks the *current visibility*:
  - a bare `public`/`protected`/`private` statement flips the default for all subsequently-defined methods (this is how Ruby models it — it is a separate statement, not a lexical modifier on the `def`);
  - the `private def ... end` form applies to that single method only, without flipping the default;
  - the `private :symbol` form retroactively re-tags an already-built method.
- The raw keyword is kept losslessly in `Declaration.modifiers`, and additionally projected onto `Declaration.visibility` (following the CXX reference pattern). The default (no modifier) is `PUBLIC`.

## Mapping / judgement calls

- Ruby's runtime semantics do not map perfectly onto the canonical access-control axis, so the mapping is deliberately lossy but faithful and documented in `RubyLanguage.interpretKeyword`:
  - `private` in Ruby means "callable only *without* an explicit receiver" (implicitly on `self`) — stricter than "declaring record only", closest canonical value `PRIVATE`.
  - `protected` means "callable *with* an explicit receiver of the same class or a subclass" — `PROTECTED`.
- Top-level `def`s (outside any class) are left `UNKNOWN`: Ruby's visibility keywords are meaningful for record members, and the canonical `PROTECTED`/`PRIVATE` are record-member concepts.
- `Integer`/`Float` etc. are unaffected. No changes were made to `cpg-core` or any other module.

## Tests

- Added `testMethodVisibility` plus a `visibility.rb` fixture covering every value produced: the `PUBLIC` default, a `protected`-flipped method, a `private`-flipped method, a `public`-flipped-back method, a single-method `private def`, and a trailing method proving the `private def` did not flip the default.
- `./gradlew :cpg-language-ruby:test -PenableRubyFrontend=true` is green (4/4). The Ruby frontend is gated behind `-PenableRubyFrontend=true` in `settings.gradle.kts`.

This PR is agent-generated and stacked on `ak/static-c`.


KuechA:
## Fixer pass: applied review feedback

All four reviews landed at **approve_with_nits** (no blockers/majors). Summary of what each raised and what I did.

### Applied
- **`initialize` tagged PUBLIC but is always private in Ruby** (adversarial-correctness, language-behavior): now forced to `Visibility.PRIVATE` (modifier `private`) regardless of the ambient class default, with an explanatory comment. Added a test asserting it.
- **Retroactive `private :symbol` path was untested** (adversarial-correctness): added `withdraw`/`transfer` defined public then `private :withdraw, :transfer`, exercising the single- and multi-symbol form and asserting the prior `public` modifier is replaced (not accumulated).
- **`visibilityKeyword` parameter shadows the `visibilityKeyword()` helper** (oxisto): renamed the value parameter to `keyword` across `addMethod`/`retagVisibility`/`applyVisibility`.
- **`applyVisibilityToExisting` clunky name** (oxisto): renamed to `retagVisibility`.
- **`applyVisibility` lacked KDoc / opaque safe-cast** (maximiliankaul): added KDoc explaining the `HasKeywordSemantics` indirection, the `RECORD` context, the defensive `as?`, and the `?.let` "no opinion" contract.
- **Awkwardly wrapped comment + ambiguous `[visibilityKeyword]` KDoc link** (oxisto): reflowed and reworded.
- **Empty `else -> {}` / retroactive-vs-fresh distinction under-explained** (oxisto, maximiliankaul): added explanatory comments; documented that the `SymbolNode` branch relies on the `def` already existing and `retagVisibility` no-ops otherwise.
- **Class/singleton methods (`def self.foo`), `*_class_method`, module member visibility, top-level-def simplification** (language-behavior, adversarial-correctness): documented as deliberate, out-of-scope omissions in the relevant KDoc/comments (handleClassNode, handleRootNode, handleDefnNode).
- **Private-in-subclass mismatch** (adversarial-correctness): added a TODO in `RubyLanguage.interpretKeyword` noting that once superclasses are modeled, PRIVATE will need handling closer to PROTECTED.

### Skipped
- **"public" modifier injected for unmarked methods** (adversarial-correctness, nit): the design intentionally records the *effective* visibility keyword losslessly, and the existing test asserts `"public" in deposit.modifiers`. Changing this would contradict both the test and the stated intent — left as-is.
- **Extract shared body-flattening helper** (oxisto, nit): the two copies live in different classes (DeclarationHandler vs. frontend); extracting a shared extension is low-value churn for a nit. Skipped to avoid gold-plating.

Formatted with spotless; `:cpg-language-ruby:test` is green.

oxisto (approved):

PR #2863, opened 2026-08-05
Model LLVM IR linkage visibility

KuechA:
## Summary

Wires the `:cpg-language-llvm` frontend into the canonical, language-agnostic `Visibility` model (introduced on `ak/static-c`). LLVM IR expresses visibility through the [linkage type](https://llvm.org/docs/LangRef.html#linkage-types) of global values and functions rather than through record access control, so linkage is the axis we map here.

## Mapping

- `private` / `internal` linkage (and the historical `linker_private` / `linker_private_weak` spellings) confine a symbol to its own module (translation unit) -> `Visibility.INTERNAL`. This is the meaningful case: the `SymbolResolver` drops such declarations as candidates for references originating in another translation unit.
- The default `external` linkage participates in normal cross-module linking and is visible everywhere -> `Visibility.PUBLIC`.
- Every other linkage type (`weak`, `linkonce`, `common`, `appending`, `extern_weak`, ...) carries no canonical visibility meaning, so it leaves the visibility `UNKNOWN` ("no restriction").

Implementation:
- `LLVMIRLanguage` now implements `HasKeywordSemantics` and interprets linkage keywords into `KeywordSemantics`. LLVM has no record access control, so it deliberately does **not** implement `HasAccessControl`.
- `DeclarationHandler` reads `LLVMGetLinkage` for both global variables and functions, records the raw linkage keyword losslessly in `Declaration.modifiers`, and projects the interpreted `Visibility` onto `Declaration.visibility`.

## Judgement calls

- `external` is LLVM's default linkage and is **not** written out in the IR, so it is not added to `modifiers` (only non-default linkage keywords are recorded there), but its visibility is still set to `PUBLIC`. There is no way to distinguish an explicitly-spelled `external` from the implicit default via the LLVM API, and both are treated identically.
- LLVM also has *visibility styles* (`default`/`hidden`/`protected`), but linkage is the primary and more meaningful axis for symbol resolution, so these are left out of scope for now.
- No `ValueDeclaration.isStatic` mapping: LLVM has no notion of static record members.

## Tests

Added `LLVMLinkageTest` plus a `linkage.ll` fixture covering `external` (default), `internal`, `private`, and `weak` linkage on both globals and functions, and a direct unit test of `interpretKeyword`. `./gradlew :cpg-language-llvm:test -PenableLLVMFrontend=true` is green (the LLVM frontend is gated behind that property in `settings.gradle.kts`).

---

This PR is agent-generated and stacked on `ak/static-c`.


KuechA:
Applied the linkage-visibility review feedback (four reviews: adversarial-correctness, language-behavior, oxisto, maximiliankaul). Summary of what each raised and what I did:

**Central issue (raised by all four):** the `external -> PUBLIC` mapping plus the `else -> null` fallback in `linkageKeyword`. The KDoc claimed lossless recording in `modifiers`, but `else -> null` dropped the keyword and `handleLinkage` reconstructed `external` from the null, silently classifying any unrecognized/obsolete linkage (dllimport/dllexport/ghost, etc.) as `PUBLIC` (maximiliankaul flagged this as a **major**; adversarial-correctness and language-behavior as minors).

**Applied:**
- Map the default `external` linkage onto `Visibility.UNKNOWN` instead of `PUBLIC`, matching the sibling linkage-based C frontend and the canonical convention documented in `Visibility.kt` (external linkage carries no access-control restriction). This resolves the language-behavior consistency finding and neutralizes the misclassification risk entirely.
- Reworked `handleLinkage` so it no longer reconstructs `external` from a null keyword (oxisto: circular round-trip). A null keyword now records nothing and leaves visibility untouched; recognized linkages are still recorded losslessly in `modifiers`.
- Rewrote the `linkageKeyword`/`handleLinkage`/`interpretKeyword` KDoc and comments to describe the actual fallback, added an inline comment on `else -> null`, documented the `val language = language` smart-cast idiom (maximiliankaul minor), reflowed the orphaned-word comments (oxisto/maximiliankaul nits), and shortened the fully-qualified KDoc links (oxisto nit).
- Extended tests/resource: added a `common` global and a `weak` function (keyword recorded in `modifiers`, visibility `UNKNOWN`) to cover the "recorded-but-UNKNOWN" branch, and assert `external` stays `UNKNOWN` with no modifier (adversarial-correctness thin-coverage nit).

**Skipped:**
- Dropping/enumerating `linker_private` / `linker_private_weak` and the obsolete DLL/ghost linkages (adversarial-correctness/language-behavior nits): kept the historical spellings as defensive handling (they are documented and cheap), and did not add spellings for the obsolete values since they have no reachable/meaningful keyword and now correctly fall through to UNKNOWN with the honest fallback comment.
- Referencing shared string constants between `LLVM_INTERNAL_LINKAGES` and the `when` (oxisto nit): the two serve different mappings (int->spelling vs spelling->semantics) and sharing would couple unrelated concerns for little gain.

Formatted with spotless; `:cpg-language-llvm:test` is green.

PR #2864, opened 2026-08-06
cpg-ai: add free-text notes and parameter-correspondence to LLMConcept/LLMOperation

maximiliankaul:
Adds a `notes` field to `LLMConcept`/`LLMOperation` (and their `GenericLLMConcept`/`GenericLLMOperation` graph overlays) for free-text context such as required setup/prerequisite calls (e.g. `init` → `setKey` before `encrypt`).

Also fixes a pre-existing bug where a property's `description` was silently dropped when applying a concept/operation to the graph (`GenericProperties` only stored the bare value). `description` is now preserved, and its docs mention
using it to record which of the tagged function's actual parameters a property corresponds to.

- `LLMProperty`/`LLMOperation`/`LLMConcept`: `description`/`notes` wording updated
- `GenericProperties`/`GenericPropertyValue`: new value type carrying `description` alongside `value` (property `type` propagation left as a follow-up TODO)
- `GenericLLMConcept`/`GenericLLMOperation`: new `notes` field
- `CpgGenericConceptsTool`: threads `description`/`notes` through when applying concepts/operations


maximiliankaul:
No longer needed. Included in #2854.

PR #2867, opened 2026-08-06
PtP: Stable handleCall

morbitzer:
Inconsistencies can arise in handleCall due to varying order in sets. This PR should fix this problem and ensure correct, consistent results over runs by
+ Deduplicating entries to be processed
+ Adding generalState entries instead of overwriting old ones


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PointsToPass.kt
```diff
@@ -1618,23 +1618,23 @@ open class PointsToPass(ctx: TranslationContext) : EOGStarterPass(ctx, orderDepe
         val param: Node?, // From which parameter entry in the functionSummary did we gather the
         // information
         val srcNode: Node?,
-        val lastWrites: MutableSet<NodeWithPropertiesKey>,
+        val lastWrites: ConcurrentIdentitySet<NodeWithPropertiesKey>,
```
KuechA: Is the usage of `IdentitySet` a good idea? This would not de-duplicate equal but non-identical entries and aren't these always new objects?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PointsToPass.kt
```diff
@@ -1618,23 +1618,23 @@ open class PointsToPass(ctx: TranslationContext) : EOGStarterPass(ctx, orderDepe
...
     ) {
         override fun equals(other: Any?): Boolean {
             return other is MapDstToSrcEntry &&
+                param == other.param &&
                 srcNode === other.srcNode &&
-                lastWrites.size == other.lastWrites.size &&
-                lastWrites.all { lw -> other.lastWrites.any { it === lw } } &&
-                propertySet.size == other.propertySet.size &&
-                propertySet.all { p -> other.propertySet.any { it == p } } &&
+                // Skip checking for lastWrites b/c we change that
```
konradweiss, line 1629: Here i would propose to either write why we skip lastWrites checks here or document what is relevant to be checked here. It is tempting to document the diff to the previous code, and certainly can be useful for the reviewer, but the comment will not increase understanding once the PR is merged.

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PointsToPass.kt
```diff
@@ -2231,24 +2330,39 @@ open class PointsToPass(ctx: TranslationContext) : EOGStarterPass(ctx, orderDepe
             }
         }
 
+        fun insertOrMerge(
```
konradweiss, line 2339: Some doc may be good here, I am not certain if this function is truly trivial and does not need the documentation, so I will let you decide if you can find some relevant parts to explain or point out here.

konradweiss (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/PointsToPass.kt
morbitzer: I don't see why it wouldn't help understanding the code after merge. When you look at it, it's useful to know why we don't compare the lastwrite, now and after the PR

PR #2874, opened 2026-08-11
Improved positions for JVM bytecode

KuechA:
SootUp uses the jimple IR to emit the code of JVM bytecode. However, when reporting the code location, it refers to some dummy lines in the dex or jar file or so. This PR aligns the two things by doing the following: It creates intermediate jimple-files containing the IR and identifies the locations there. This is beneficial to align the code representation with the reported lines which, in turn, can help debugging, identifying nodes based on locations, and also helps to create more unique IDs compared to the other approach. I'm well aware that this PR is not really pretty but I'm not sure what would be better alternatives.


konradweiss (approved):
LGTM although it is not pretty as you mentioned. I lack the understanding of any negative impacts this may have but it is isolated in the JVM language frontend and therefore should be a strict improvement when working with the JVMFrontend and not negatively impact the remaining project.

PR #2878, opened 2026-08-17
Making the `DFGPass` `open` and adding a fallback if `ControlFlowSensitiveDFGPass` and `PointsToPass` are not registered

KuechA:


KuechA:
Moved to PR #2921 and extended the logic there

PR #2879, opened 2026-08-18
C language frontend: `wchar_t` support

KuechA:


cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/ExpressionHandler.kt
```diff
@@ -766,21 +777,33 @@ class ExpressionHandler(lang: CXXLanguageFrontend) :
...
+            // instead of running it through the byte-recombination logic below.
+            return newLiteral(single, primitiveType("int"), rawNode = ctx)
         } else {
-            // Somehow make an int out of, this is "implementation" specific. We follow the way
-            // clang does it
+            // Somehow make an int out of, this is "implementation" specific.
             var intValue = 0
-            for ((n, c) in chars.reversed().withIndex()) {
-                intValue += (c.code * 256.0f.pow(n)).toInt()
+            if (isWideChar) {
```
lshala: Can we test this branch somehow? It doesn't seem to be covered by any test

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/ExpressionHandler.kt
```diff
@@ -766,21 +777,33 @@ class ExpressionHandler(lang: CXXLanguageFrontend) :
...
                 }
             }
             i++
         }
 
         val single = chars.singleOrNull()
-        if (single != null) {
-            return newLiteral(single, primitiveType("char"), rawNode = ctx)
+        if (single != null && single <= Char.MAX_VALUE.code) {
+            return newLiteral(single.toChar(), primitiveType("char"), rawNode = ctx)
```
lshala: I was wondering, why this always returns type char or int? Do we need to register `wchar_t` in the builtin types (as CPPLanguage already does)to be able to return it here? Or is this irrelevant?

lshala (approved):

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/ExpressionHandler.kt
KuechA: Fair point. I think it's irrelevant, but I added it nevertheless

cpg-language-cxx/src/main/kotlin/de/fraunhofer/aisec/cpg/frontends/cxx/ExpressionHandler.kt
KuechA: I added a test. This was somehow my first attempt, but it turned out to work differently than expected. The code is still useful, though.

PR #2897, opened 2026-08-19
Ak/eog based symbol resolver

KuechA:


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -457,8 +478,12 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
     /**
      * The central entry-point for all symbol-resolving. It dispatches the handling of the node to
      * the appropriate function based on the node type.
+     *
+     * Visibility is `internal` rather than `protected` so that [acceptWithIterateEOG] (an extension
+     * function living in a separate file) can invoke it as well; both traversal strategies funnel
+     * through this single dispatcher.
      */
-    protected open fun handle(node: Node?) {
+    internal open fun handle(node: Node?) {
```
KuechA: Should be kept `protected` to allow overwriting this

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -119,6 +120,19 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
+     * declared symbols come from during a [handleReference] lookup, forwarded to
+     * [ScopeManager.lookupSymbolByNodeName] as `localSymbols`. `null` (the default) means every
+     * lookup uses [de.fraunhofer.aisec.cpg.graph.scopes.Scope.symbols] as usual, which is what the
+     * default, non-flow-sensitive traversal in [accept] relies on.
+     * [de.fraunhofer.aisec.cpg.passes.acceptWithIterateEOG] sets this to answer "which declarations
+     * have been reached by this point in the EOG" for
+     * [de.fraunhofer.aisec.cpg.graph.scopes.LocalScope]s specifically, making local (block-scoped)
+     * references flow-sensitive while leaving every other kind of scope untouched.
+     */
+    internal var localSymbolsOverride: ((Scope, Symbol) -> List<Declaration>?)? = null
```
KuechA: protected or public

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
```diff
@@ -828,6 +830,7 @@ class ScopeManager(override var ctx: TranslationContext) : ScopeProvider, Contex
         location: PhysicalLocation? = null,
         startScope: Scope? = currentScope,
         replaceImports: Boolean = true,
+        localSymbols: ((Scope, Symbol) -> List<Declaration>?)? = null,
         predicate: ((Declaration) -> Boolean)? = null,
     ): List<Declaration> {
         val extractedScope = extractScope(name, language, location, startScope)
```
lshala: The flow-sensitive symbol source is not used while resolving a qualified name’s scope. For `A::member`, `extractScope` delegates to `lookupScopeByName`, which looks up `A` in the complete static `scope.symbols`; therefore a block-local `A` declared later can already be selected here. Please thread `localSymbols` through scope extraction as well, so the qualifier and terminal symbol use the same program-point state.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolverEOGIteration.kt
```diff
@@ -25,411 +25,271 @@
...
+    // tuple/multiple declaration, or a locally-declared function prototype (itself its own,
+    // separate EOG starter, possibly with a nonempty nextEOGEdges of its own for evaluating a
+    // default argument value - but never reached *from* this starter's own traversal, since
+    // nothing has an edge pointing into it) - exactly like a Function's Parameters are never
+    // reached via EOG either. Since these are declared "atomically" as part of a larger construct
+    // rather than as an ordinary, sequential statement, they must be visible from the very start
+    // rather than waiting to be "reached" by a transfer() call that will never come. The relevant
+    // criterion is having no *incoming* EOG edge; an unrelated, nonempty outgoing chain of its own
+    // doesn't change that nothing in *this* traversal will ever visit it.
+    val preseeded =
```
lshala: Pre-seeding every local declaration without an incoming EOG edge makes it visible from the start of the entire EOG starter. In particular, the locally declared function prototype mentioned above is only in scope from its declaration point; a call earlier in the same block will now resolve to that later prototype. Declarations that are atomic parts of constructs (such as catch parameters) need to be introduced when entering that construct, while prototypes need to enter the state at their lexical declaration point.

lshala (commented):
Copilot: I found two flow-sensitivity gaps that can resolve declarations before they are visible.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
KuechA: Agreed, legitimate. Restored `handle` to `protected`. The problem was that `acceptWithIterateEOG` used to be an extension function in a separate file, and an extension function outside a class cannot call a `protected` member even in the same module. Fixed by moving `acceptWithIterateEOG` and everything it uses into `SymbolResolver` as class members (and deleting the now-empty `SymbolResolverEOGIteration.kt`), so `handle` can stay `protected` as before. See c81840b32a09.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
KuechA: Agreed, legitimate (same root cause as the `handle` visibility comment). Went with `protected` for consistency with `handle` and the rest of the class's overridable members - fixed alongside that comment (c81840b32a09).

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/ScopeManager.kt
KuechA: Confirmed and fixed. Threaded `localSymbols` through both `extractScope` overloads and `lookupScopeByName`, and forwarded it from `lookupSymbolByName`'s call site into `extractScope` (previously it was accepted as a parameter but never actually passed along). Qualified lookups like `Local::bar()` now use the same flow-sensitive state for the qualifier as for the terminal symbol; added a regression test (`testLocalClassQualifiedAccessOnlyVisibleFromItsDeclarationPoint`) covering exactly this case. See c81840b32a09.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolverEOGIteration.kt
KuechA: Confirmed and fixed - thanks for the precise repro scenario. Replaced the blanket pre-seed-everything-at-start approach with a per-declaration `SeedPlan`: declarations that are structurally atomic parts of their enclosing construct (e.g. a catch clause's exception parameter) are still seeded at the start, but declarations that are themselves a direct statement-list element (a local function prototype, or a tuple/multiple-declaration's wrapper) are anchored to the nearest *preceding* EOG-reachable sibling statement and only become visible once that anchor has *finished* being handled - not before.

Note the anchor points backward (to the preceding reachable sibling), not forward to the next one: a compound statement's own sub-expressions (e.g. a call's callee reference) are reached via EOG before the statement node itself is, so anchoring forward and seeding beforehand was still one step too late for anything referencing the declaration within the very statement it's anchored to. This was caught via several regression failures across cpg-core/cxx/java/python during a full-diagnostic run before landing on the backward-anchor design.

Added a regression test (`testLocalPrototypeOnlyVisibleFromItsDeclarationPoint`) reproducing exactly the scenario you described - the early call now correctly fails to resolve to the local prototype and falls back to inference, and the late call resolves correctly.

One transparent note: under a forced-diagnostic run (temporarily routing every pipeline through the new engine, not just opt-in ones), `testScopedFunctionResolutionWithDefaults` shows an expected, unrelated cascading effect - a genuinely invalid call in that fixture (`f(8)`, calling a 2-arg function with 1 arg, no default) now correctly fails to resolve instead of incorrectly matching, and its resulting inferred global function becomes an extra viable candidate for an unrelated call in the same fixture. This test does not opt into `experimentalEOGWorklist` in the normal suite, so it is unaffected in CI; flagging it here for visibility since it surfaced directly from this fix. See c81840b32a09.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -171,6 +206,324 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
+                    newState[scope]?.filter { it.name.localName == symbol } ?: emptyList()
+                } else {
+                    // Not a LocalScope we're tracking flow-sensitively (either not a LocalScope
+                    // at all, or one belonging to an enclosing, already-resolved starter): fall
+                    // back to the default, static Scope.symbols[symbol] lookup.
+                    null
+                }
+            }
+
+            if (node is BinaryOperator || node is UnaryOperator) {
```
lshala: **Deferring operator handling until after the traversal can break downstream resolution that depends on the operator's result type.**

In the `ScopedWalker` path, `handleOverloadedOperator` replaces the operator with an `OperatorCall` *in EOG order*, so the node that consumes the operator already sees its return type. Here the operator is pushed to `deferredOperatorNodes` and only handled after `iterateEOG` finishes, while `propagateOperatorType` leaves it with the raw arithmetic/deref type.

Concrete C++ to check: `(*it).field` with an overloaded `operator*` on an iterator, or `(a + b).foo()` with an overloaded `operator+` returning a class. When `handleMemberAccess`/`handleCall` runs for `.field`/`.foo`, the base is still the un-replaced `UnaryOperator`/`BinaryOperator` whose type is `iteratorType.dereference()` / `arithmeticOpTypePropagation(...)` — i.e. unknown — so `handleCall` bails out with *"Cannot resolve call … because the base type is unknown"*. The default path resolves this.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -171,6 +206,324 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
+        if (declarationScope is LocalScope) {
+            newState =
+                lattice.lub(
+                    newState,
+                    LocalDeclarationElement(declarationScope to PowersetLattice.Element(node)),
+                    true,
+                )
+        }
+
+        if (handledNodes.add(node)) {
```
lshala: **`handle` only ever runs with the *first* state that reaches a node, so at a join that is not a plain merge the resolution is decided by whichever predecessor the engine schedules first.**

`handledNodes` correctly prevents double-handling, but as a consequence the `iterateEOG` fixpoint has no effect on resolution: the node is never re-handled with the joined state.

Concrete, legal C++:

```cpp
switch (c) {
  case 1: int y = 1; break;
  case 2: y = 2; break;   // y is in the switch block's LocalScope
}
```

`handleSwitch` wires `selector -> Case2` directly, so the reference `y` in case 2 is first reached on a path that never passed `int y = 1`. `localSymbolsOverride` returns `emptyList()` for that `LocalScope`, the candidate set is empty, and `tryVariableInference` invents a bogus variable. The `ScopedWalker` path resolved it to the local `y`.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -141,21 +176,21 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
-                    if (passConfig?.skipUnreachableEOG == true) {
-                        Strategy::REACHABLE_EOG_FORWARD
-                    } else {
-                        Strategy::EOG_FORWARD
-                    },
-                )
+        walker =
+            ScopedWalker(
+                scopeManager,
+                if (passConfig?.skipUnreachableEOG == true) {
```
lshala, line 185: **`Configuration.skipUnreachableEOG` is silently ignored on the worklist path.**

`walker` is now built with `Strategy::REACHABLE_EOG_FORWARD` when the flag is set, but on the worklist path the walker is never iterated; `iterateEOG` follows every `nextEOGEdges` regardless of `EvaluationOrder.unreachable`. Enabling `skipUnreachableEOG` together with `experimentalEOGWorklist` therefore has no effect — unreachable code is resolved (and inferred into) anyway, with no warning to the user.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -171,6 +206,324 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
+     * [handle] does not compute the type of a [BinaryOperator] or [UnaryOperator] itself; normally,
+     * it relies on [HasType]'s reactive [HasType.TypeObserver] mechanism to propagate the type of
+     * [BinaryOperator.lhs]/[BinaryOperator.rhs] (or [UnaryOperator.input]) once they are resolved.
+     * That mechanism can be switched off entirely via
+     * [TranslationConfiguration.Builder.disableTypeObserver], in which case nothing else computes
+     * these types. Since the EOG guarantees [node]'s operands were already handled (and thus have
+     * their final type) by the time [node] itself is reached, we can compute the type here
+     * directly, exactly mirroring what the reactive path would have done.
+     */
+    private fun propagateOperatorType(node: Node) {
```
lshala, line 508: **`propagateOperatorType` does not mirror `BinaryOperator.typeChanged`, and runs even when the type observer is enabled.**

`BinaryOperator.typeChanged` special-cases `.*`/`->*` (`this.type = newType`, the function-pointer type — see `BinaryOperator.kt:92`); `propagateOperatorType` does not, and `Language.propagateTypeOfBinaryOperation` falls into `else -> unknownType()` for those operator codes.

Since the operands are reached before the operator in the EOG, the observer has already set the correct function-pointer type by the time `propagateOperatorType` runs and overwrites it with `unknownType()` — and that write re-fires the observers, propagating "unknown" downstream.

Also, the KDoc's claim that this runs only as a stand-in for `disableTypeObserver` is not what the code does: the call is unconditional.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
```diff
@@ -171,6 +206,324 @@ open class SymbolResolver(ctx: TranslationContext) : EOGStarterPass(ctx) {
...
+                        starterScope,
+                        anchoredSeeds,
+                    )
+                },
+            )
+        if (timeout) {
+            log.warn("Could not compute final state for EOG starter {} (due to timeout)", t.name)
+        }
+
+        deferredOperatorNodes.forEach {
```
lshala: **`localSymbolsOverride` is still installed, pointing at an arbitrary program point's state, during the deferred-operator phase, and is not reset on an exception path.**

The lambda installed further up captures the `var newState` of the *last* `transfer` call and survives until it is cleared after this loop. Every `handle(it)` in the deferred loop therefore runs with that stale flow state, so any `LocalScope` lookup reached from `handleOverloadedOperator` (today none, but nothing enforces that) would be answered for the wrong program point.

Separately, `localSymbolsOverride = null` is not in a `finally`, so a throw inside `iterateEOG` leaves the override installed on the pass instance.

lshala (commented):
Automated review by **Claude Code** (posted on behalf of @lshala).

Five findings on the new EOG-worklist path in `SymbolResolver`, most-severe first. All are reasoned from the code rather than reproduced against a failing test, so please treat the concrete examples as hypotheses to check rather than confirmed repros.

Also checked and found *not* to be problems: the anchor chosen by `seedPlanFor` does dominate the declaration point (because `handleIfElse`/`handleWhile`/`handleSwitch` all `attachToEOG(node)` last); `FunctionScope` is not a `LocalScope`, so parameters are unaffected; the symbol-lookup cache is correctly bypassed when `localSymbols != null`; `Declaration.symbol == name.localName`, so the override's filter matches the symbol-table key; passes are instantiated per EOG-starter and driven sequentially, so the new mutable `localSymbolsOverride` field is not a race; and constructing `walker` without iterating it is enough for `walker.replace`.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
KuechA: Confirmed and fixed. Restructured `acceptWithIterateEOG` into two phases instead of calling `handle` directly from the `iterateEOG` transformation: phase 1 (`collectDeclarationState`) is a pure, side-effect-free pass that only computes each node's final LocalScope flow state; phase 2 reuses the exact same `ScopedWalker`/`Strategy` the default path already builds in `accept()` to do the actual resolution (calling `handle` on every node, consulting the precomputed state for `localSymbolsOverride`). Since `ScopedWalker` already tolerates `handleOverloadedOperator`'s mid-traversal AST/EOG mutation (it re-reads `nextEOGEdges` lazily, exactly like the default non-experimental path relies on), operators no longer need to be deferred at all - `(a + b).foo()`/`(*it).field` now resolve exactly like the default path. See c60ec9c6a43.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
KuechA: Confirmed and fixed - good catch, and the switch/case repro made the mechanism very clear. This was the same root architectural issue as the operator-deferral comment: resolving eagerly inside the iterateEOG transformation meant `handle` only ever saw whichever edge's state arrived first. Fixed alongside it by splitting the function into two phases (see c60ec9c6a43): phase 1 purely accumulates - via the lattice's own union `lub` - every state that ever reaches a node from every incoming edge into a per-node map, with no resolution happening yet; phase 2 (the reused `ScopedWalker`) then resolves each node using that node's fully-converged, final entry from the map. For your switch example, the reference `y` in case 2 now correctly sees the union of both incoming states (whether or not `int y = 1` executed), instead of whatever the engine happened to schedule first.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
KuechA: Confirmed and fixed as part of the same restructuring (c60ec9c6a43). Phase 2 of `acceptWithIterateEOG` now reuses the very same `walker` instance (and therefore the same `REACHABLE_EOG_FORWARD`-vs-`EOG_FORWARD` `Strategy` choice based on `skipUnreachableEOG`) that `accept()` already builds for the default path, instead of driving resolution from `iterateEOG`'s own unconditional edge traversal. `iterateEOG` (phase 1) still walks every edge including unreachable ones, but only to compute flow state - no resolution or inference happens there, so unreachable code is never resolved when `skipUnreachableEOG` is set, matching the default path's behavior.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
KuechA: Confirmed and fixed. Rather than reimplementing the propagation logic (and inevitably drifting from it), `propagateOperatorType` now invokes `BinaryOperator`/`UnaryOperator`'s own `typeChanged` callback directly (they implement `HasType.TypeObserver` themselves) with the operand's type/node as the observer would - so the `.*`/`->*` special case is reused verbatim rather than reimplemented, and can't drift from it again. It's also now only invoked when `ctx.config.disableTypeObserver` is actually true (moved the check to the phase-2 call site in c60ec9c6a43), matching the KDoc's original claim.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/SymbolResolver.kt
KuechA: Confirmed and fixed as part of the same restructuring (c60ec9c6a43). There is no more separate deferred-operator phase at all - phase 2 now handles every node (including operators) inline, in true EOG order, via the reused `ScopedWalker`, so `localSymbolsOverride` is always set immediately before the specific node's `handle` call using that node's own precomputed flow state, never a stale one left over from a prior node. Phase 2 (the resolving phase) is also now wrapped in `try { ... } finally { localSymbolsOverride = null }`, so it's reset even if something inside throws.

lshala (approved):

PR #2907, opened 2026-08-20
Add Rust support to README

konradweiss:
Updated README to include Rust support in the library description and added Rust to the supported languages table.


konradweiss:
Apparently this documentation got somehow lost before we merged the frontend. The most likely reason is that I forgot to push before merging

oxisto (approved):

PR #2916, opened 2026-08-24
Further reduce `Node` memory size by interning `code` and `name`

KuechA:
## Summary

Follow-up to #2829: reduces per-node memory further by de-duplicating data that
today gets copied redundantly across the graph, and by shrinking the object
layout of `Node`'s location. No public API changes; `Node.code`, `Node.name`
and `PhysicalLocation.region` all keep their exact existing get/set contracts.

## `Node.code` interning

`Node.code` stores a copy of the source snippet per node, even though the AST
is hierarchical — a `Method`'s code contains all its statements' code, which
contains all its expressions' code, and so on, so the same characters get
duplicated once per ancestor level.

- **`FileContentCache`**: caches a file's full text once (bounded, like
  `ArtifactLocation`'s cache), with a precomputed line-start index for O(1)
  line/column → char-offset conversion.
- **`CodeSpan`**: a lazily-populated `(content, start, end)` triple. `Node.code`
  is backed by either a `CodeSpan` or a literal `String`, chosen automatically:
  a node is only switched to the shared representation *after verifying
  byte-for-byte* that the cached file substring reproduces the frontend's code
  exactly. This makes it safe regardless of how a given frontend derives
  `code` — a substring of the real file (CXX, Python), a token-stream
  reconstruction (Java), a decompiled-bytecode re-serialization (JVM), or
  absent entirely (LLVM, `locationOf` always returns `null`): a mismatch just
  falls back to the literal string, exactly as before this change existed.
- Gated by `TranslationConfiguration.codeInterning` (default `true`) for
  A/B comparison and rollback.
- `Node.copyCodeFrom(other)` preserves the interned representation across
  `duplicate()`/cast copy sites that used to do a plain `code = other.code`.

**Measured intern hit rates** (regression-tested for byte-identical `code`
with the flag on/off): CXX ~100%, Java ~94%, Python ~93%, JVM/LLVM 0% (as
expected — no real source text ever exists for those two).

## `Name` interning

`Name` (`localName`, `parent`, `delimiter`) is *eagerly* allocated on every
node (unlike `code`) and is already fully immutable and structurally
equal/hashable, so it interns cleanly with no fallback needed:

- `NameCache`: bounded, global cache, keyed by structural equality.
- Wired into `parseName` (interning each part as an FQN is built, so shared
  prefixes like a common namespace/class end up sharing one parent instance),
  `NodeBuilder.newName`, `Name.fqn`, `append`, `replace`.
- While in there: cached `Name.hashCode()` (immutable, never needs
  invalidating), replaced `fullName`'s `by lazy` with a manual cache (avoids
  the delegate + captured-lambda object `lazy {}` allocates per instance), and
  stopped `length` from forcing that computation eagerly.

## `PhysicalLocation`/`Region` flattening

`PhysicalLocation` held a dedicated `Region` object per located node. Storage
is now four `Int` fields directly on `PhysicalLocation`, with `region` as a
get/set property that materializes/decomposes a `Region` on demand — `Region`
itself is unchanged and remains the public type used everywhere else.
`hashCode()` deliberately still delegates through the `region` getter (rather
than hashing the fields directly) so the computed value stays byte-for-byte
identical to the previous `Objects.hash(artifactLocation, region)` formula,
since `Node.id` (used for Neo4j persistence) derives from it transitively.

This required first fixing a latent bug: `Fluent.kt`'s `inferPseudoLocations`
mutated an existing `Region`'s fields in place, bypassing `Node.location`'s
setter — the only thing that invalidates the cached `hashCode`. That already
violated the invariant `Node.hashCode`'s own doc comment claims. Fixed by
reconstructing and reassigning `PhysicalLocation` wholesale instead.

## Explored and rejected

- **Disk-backed slicing instead of a shared in-memory copy**: would only
  additionally save the ~0.4–0.9% of retained heap the shared file text
  itself accounts for, while making every `code` read (including inside
  `Node.equals()`, called constantly by worklists/hashmaps) ~5.5x slower even
  under a warm OS page cache — a bad trade for a marginal gain.
- **`-XX:+UseStringDeduplication`**: only 0.04% additional reduction on top of
  the explicit interning above — confirms the two dominant duplicate-string
  sources were already found and fixed explicitly.
- **Flattening `CodeSpan` directly onto `Node`** (like `Region`): rejected.
  Unlike `Region` (present on essentially every node), `code` interning has a
  0% hit rate on JVM/LLVM; flattening would make those frontends pay two
  unconditional `Int` fields for zero benefit. The current lazy-wrapper
  design (`codeSpan: CodeSpan?`) costs nothing extra on nodes that never
  intern.

## Measured impact (JOL `GraphLayout`, vs. the pre-branch baseline)

| File | Before | After | Reduction |
|---|---|---|---|
| `double_scalarmult_vartime.cpp` (5.5k lines, CXX) | 49.07 MB | 43.57 MB | **-11.2%** |
| `morbi.c` (93.6k lines, C) | 962.9 MB | 835.9 MB | **-13.2%** |

The percentage holds at scale (a 17x bigger file gave essentially the same
percentage, just proportionally more absolute savings) — it's a structural
property of how much of a `Node`'s footprint this data accounts for, not a
small-file artifact.

## Testing

Full unit-test suite green after each commit. New tests: `FileContentCacheTest`,
`NameCacheTest`, `PhysicalLocationTest`, `NodeTest` additions, CXX/Python
before/after code-interning equality tests, plus `@Ignore`d ad-hoc benchmarks
(`CodeInterningBenchmarkTest`, `NameInterningBenchmarkTest` per language module)
documenting the measurements above for anyone who wants to reproduce them.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Name.kt
```diff
@@ -65,15 +65,30 @@ class Name(
...
+                cachedFullName = f
+            }
+            return f
+        }
+
+    /** Caches [hashCode]. [Name] is fully immutable, so this never needs to be invalidated. */
+    private var cachedHashCode: Int = 0
 
-    public override fun clone(): Name = Name(localName, parent?.clone(), delimiter)
+    public override fun clone(): Name =
```
KuechA: Won't this return the same object again? Why do we need `clone`?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NameCache.kt
```diff
@@ -0,0 +1,66 @@
...
+internal object NameCache {
+    /**
+     * Upper bound on the number of cached names. Names are far more numerous than the files
+     * [de.fraunhofer.aisec.cpg.sarif.PhysicalLocation.ArtifactLocation] interns (every node has
+     * one), hence the larger bound; dropping entries once it's reached only forgoes sharing for
+     * subsequent lookups, never correctness.
+     */
+    private const val MAX_CACHE_SIZE = 1_000_000
+
+    private val cache = ConcurrentHashMap<Name, Name>()
```
KuechA: It could make sense to initialize this map with a larger size to avoid copies of tiny maps

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/sarif/PhysicalLocation.kt
```diff
@@ -87,7 +87,23 @@ class PhysicalLocation(uri: URI?, region: Region) {
...
+    // The region is stored as four flat Int fields rather than a dedicated Region object, saving
+    // one object header per location; Region (still the public type everywhere else) is
+    // materialized on demand by the region getter/setter below.
+    private var startLine = 0
+    private var startColumn = 0
+    private var endLine = 0
+    private var endColumn = 0
+
     var region: Region
+        get() = Region(startLine, startColumn, endLine, endColumn)
```
KuechA, line 135: setter missing? Or could be `val`

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
```diff
@@ -708,6 +721,18 @@ private constructor(
...
         }
 
+        /**
+         * Enables or disables interning [Node.code][de.fraunhofer.aisec.cpg.graph.Node.code] as
+         * offset ranges into a shared per-file text (default `true`). Disable this only for A/B
+         * comparisons or as a rollback; it never changes the value nodes report for `code`.
+         *
+         * @param b the new value
+         */
+        fun codeInterning(b: Boolean): Builder {
```
KuechA: I'd remove this option. I don't see why the user should configure this as it's purely internal

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/CodeInterningBenchmarkTest.kt
```diff
@@ -0,0 +1,233 @@
...
+import kotlin.test.Test
+import org.openjdk.jol.info.GraphLayout
+
+/**
+ * Ad-hoc benchmark, not part of CI: reports how many characters of `code` payload are retained with
+ * [de.fraunhofer.aisec.cpg.TranslationConfiguration.codeInterning] on vs. off, on a real
+ * multi-thousand-line file.
+ */
+@Ignore
+class CodeInterningBenchmarkTest {
```
KuechA: remove before merging this PR

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/NameInterningBenchmarkTest.kt
```diff
@@ -0,0 +1,103 @@
...
+ * Ad-hoc benchmark, not part of CI: unlike `code`, `Name` interning has no on/off config flag (it
+ * is purely additive and behaviorally invisible, like `Node`'s cached hashCode), so there's no
+ * "before" state to toggle back to in the same run. Instead, this measures sharing directly: for
+ * every node's name chain (the name plus every ancestor via [Name.parent]), it counts how many
+ * *chain positions* exist in total (what would have been allocated with zero sharing, since every
+ * node used to build its own fresh chain) vs. how many *distinct* [Name] instances those positions
+ * actually resolve to (by identity) now that they're interned.
+ */
+@Ignore
+class NameInterningBenchmarkTest {
```
KuechA: Should be removed before merging this PR

cpg-language-java/src/test/kotlin/de/fraunhofer/aisec/cpg/CodeInterningBenchmarkTest.kt
```diff
@@ -0,0 +1,82 @@
...
+ * [de.fraunhofer.aisec.cpg.TranslationConfiguration.codeInterning] on vs. off.
+ *
+ * Java's `codeOf` (`JavaLanguageFrontend.codeOf`) returns `astNode.tokenRange.toString()` (falling
+ * back to the pretty-printer), i.e. a reconstruction from JavaParser's token stream, not a plain
+ * substring of the file. FileContentCache.rangeOf only interns when its own line/column-derived
+ * substring matches that reconstruction character for character, so this measures how often that
+ * happens to be the case in practice.
+ */
+@Ignore
+class CodeInterningBenchmarkTest {
```
KuechA: Should be removed before merging this PR

cpg-language-jvm/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/CodeInterningBenchmarkTest.kt
```diff
@@ -0,0 +1,71 @@
...
+ * Ad-hoc benchmark, not part of CI: reports how many characters of `code` payload are retained with
+ * [de.fraunhofer.aisec.cpg.TranslationConfiguration.codeInterning] on vs. off, for a `.jar`.
+ *
+ * This frontend has no original source text at all -- `code` is SootUp's re-serialization of the
+ * decompiled bytecode (`JVMLanguageFrontend.codeOf`: `SootMethod.body.toString()` /
+ * `astNode.toString()`), never a substring of any file on disk. FileContentCache.rangeOf can't
+ * verify a match against nothing, so interning is expected to never engage.
+ */
+@Ignore
+class CodeInterningBenchmarkTest {
```
KuechA: Should be removed before merging this PR

cpg-language-cxx/build.gradle.kts
```diff
@@ -51,4 +51,5 @@ dependencies {
 
     testImplementation(libs.junit.params)
     testImplementation(project(":cpg-analysis"))
+    testImplementation(libs.jol.core)
```
KuechA: Is this really necessary?

cpg-language-llvm/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/llvm/CodeInterningBenchmarkTest.kt
```diff
@@ -0,0 +1,71 @@
...
+ * Ad-hoc benchmark, not part of CI: reports how many characters of `code` payload are retained with
+ * [de.fraunhofer.aisec.cpg.TranslationConfiguration.codeInterning] on vs. off, on a real ~900-line
+ * `.ll` file.
+ *
+ * `LLVMIRLanguageFrontend.locationOf` always returns `null` (no location support at all), so
+ * FileContentCache.rangeOf never even has a region to try -- interning is expected to never engage
+ * for this frontend.
+ */
+@Ignore
+class CodeInterningBenchmarkTest {
```
KuechA: Should be removed before merging this PR

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/CodeInterningBenchmarkTest.kt
```diff
@@ -0,0 +1,123 @@
...
+import kotlin.test.Test
+import org.openjdk.jol.info.GraphLayout
+
+/**
+ * Ad-hoc benchmark, not part of CI: reports how many characters of `code` payload are retained with
+ * [de.fraunhofer.aisec.cpg.TranslationConfiguration.codeInterning] on vs. off, on a real
+ * multi-thousand-line file.
+ */
+@Ignore
+class CodeInterningBenchmarkTest {
```
KuechA: Should be removed before merging this PR

cpg-language-python/build.gradle.kts
```diff
@@ -42,4 +42,6 @@ dependencies {
     // to evaluate some test cases
     testImplementation(projects.cpgAnalysis)
     integrationTestImplementation(projects.cpgAnalysis)
+
+    testImplementation(libs.jol.core)
```
KuechA: Is this really necessary?

gradle/libs.versions.toml
```diff
@@ -31,6 +31,7 @@ kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-c
 kotlinx-collections-immutable = { module = "org.jetbrains.kotlinx:kotlinx-collections-immutable", version = "0.4.0"}
 
 reflections = { module = "org.reflections:reflections", version = "0.10.2"}
+jol-core = { module = "org.openjdk.jol:jol-core", version = "0.17"}
```
KuechA: Necessary?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/sarif/FileContentCache.kt
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Upper bound on the number of cached files, mirroring
+     * [de.fraunhofer.aisec.cpg.sarif.PhysicalLocation.ArtifactLocation]'s bound: this only needs to
+     * help within the working set of files currently being analyzed, so dropping cache entries only
+     * forgoes sharing, never correctness.
+     */
+    private const val MAX_CACHE_SIZE = 50_000
+
+    /** `Optional.empty()` is cached too, so an unreadable/virtual [URI] is not retried per node. */
+    private val cache = ConcurrentHashMap<URI, Optional<IndexedContent>>()
```
KuechA: Instead of having this here: Why don't we use the code of a `TranslationUnit`. That should be the top declaration of most nodes (unless inferred and implicit ones). In the `Node`class, we could then lookup the code of the parentTU (may require a new field to access the translation unit quickly) and get its code from there. This would probably move the cache to the translationUnit but we don't have to make a lookup for the URI, only get the substring for the index. What are Pros and Cons of this?

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/Name.kt
KuechA: Good catch — fixed in e0d1d47a43d. Name is fully immutable, so clone() has nothing to actually copy; it now just returns `this`.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/graph/NameCache.kt
KuechA: Done in e0d1d47a43d — initialized with capacity 4096.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/sarif/PhysicalLocation.kt
KuechA: Fixed — it's a `val` now (only ever set once, in the init block).

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/TranslationConfiguration.kt
KuechA: Removed the option entirely (and the underlying field/config plumbing) in 6f4889bdf9b. Interning is opportunistic and self-verifying by construction (a mismatch always falls back to the literal string), so there's no real scenario where a user would need to disable it — it only existed for the A/B-comparison tests in this PR, which are removed separately.

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/sarif/FileContentCache.kt
KuechA: Great suggestion, implemented in 6f4889bdf9b. Replaced FileContentCache with an `indexedContent` field on `PhysicalLocation.ArtifactLocation` (already interned per URI, so no separate lookup) populated from the TranslationUnit's own `code` once its location is finalized. Pros: no separate disk read (the TU already parsed the whole file), no separate cache/map or bounded-eviction policy, one less lookup since the ArtifactLocation is already reachable via `location.artifactLocation`. One wrinkle: I couldn't register the content inline at the point code/location are first assigned in `setCodeAndLocation`, because some frontends (Python) assign an approximate/null location first and finalize it afterward — verified this empirically, it silently dropped Python's interning to 0%. Fixed by overriding `TranslationUnit.location`'s setter instead, which fires correctly regardless of a frontend's construction order.

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/CodeInterningBenchmarkTest.kt
KuechA: Removed in 9cc93be8bee.

cpg-language-cxx/src/test/kotlin/de/fraunhofer/aisec/cpg/NameInterningBenchmarkTest.kt
KuechA: Removed in 9cc93be8bee.

cpg-language-java/src/test/kotlin/de/fraunhofer/aisec/cpg/CodeInterningBenchmarkTest.kt
KuechA: Removed in 9cc93be8bee.

cpg-language-jvm/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/jvm/CodeInterningBenchmarkTest.kt
KuechA: Removed in 9cc93be8bee.

cpg-language-python/src/test/kotlin/de/fraunhofer/aisec/cpg/frontends/python/CodeInterningBenchmarkTest.kt
KuechA: Removed in 9cc93be8bee.

cpg-language-cxx/build.gradle.kts
KuechA: No longer -- it was only used by the benchmark tests, which are removed in 9cc93be8bee, so jol-core is dropped too.

cpg-language-python/build.gradle.kts
KuechA: No longer -- same as above, removed in 9cc93be8bee.

gradle/libs.versions.toml
KuechA: Removed in 9cc93be8bee, it was only needed by the benchmark tests.

konradweiss (approved):

PR #2922, opened 2026-08-25
 cpg-ai: omit source code from bulk-listing tools

maximiliankaul:
`cpg_list_functions`, `cpg_list_calls`, and `cpg_list_calls_to` embed every matching function/call's full source (`FunctionInfo.code`/`CallInfo.code`) — measured at 35–45% of a typical result's bytes. These tools are for finding candidates by  name/signature; agents only need the body of a handful of what's returned, so embedding all of it is mostly wasted context.

Adds an `includeCode` parameter (default `true`) to `FunctionInfo`/`CallInfo` and has these three tools pass `false`. `cpg_get_node` already covers fetching full details for a specific node once picked.

In a real agent session analyzing several C libraries, per-call result sizes dropped ~33–46% depending on the library, with zero sampled calls leaking code post-fix — confirming the estimate holds in practice.

Covered by `BulkListingCodeOmissionTest`.


KuechA (approved):
looks good

PR #2930, opened 2026-08-31
Fix ValueEvaluator for globals/statics with other writes

lshala:
This PR fixes the `ValueEvaluator` by not only taking the initializer for global/static variables, but also checking whether there is a write elsewhere.

### DIFF
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluator.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluator.kt
@@ -179,18 +179,9 @@ open class ValueEvaluator(
      * their value. If not, we can try to use [handlePrevDFG].
      */
     protected fun handleHasInitializer(node: HasInitializer, depth: Int): Any? {
-        // A global or static variable can have writes somewhere else too, not just its
-        // initializer. If the DFG shows more than one, use handlePrevDFG to decide instead.
-        if (
-            node is Variable &&
-                (node.isGlobal || node.modifiers.contains("static")) &&
-                node.prevFullDFG.size > 1
-        ) {
-            return handlePrevDFG(node as Node, depth)
-        }
-
-        // If we have an initializer, we can use it. Otherwise, we can fall back to the prevDFG
-        return if (node.initializer != null) {
+        // If we have an initializer and nothing else writes to it, we can use the initializer.
+        // Otherwise, we let handlePrevDFG decide.
+        return if (node is Node && node.initializer != null && node.prevFullDFG.size <= 1) {
             evaluateInternal(node.initializer, depth + 1)
         } else {
             handlePrevDFG(node as Node, depth)
--- a/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluator.kt
+++ b/cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluator.kt
@@ -179,6 +179,16 @@ open class ValueEvaluator(
      * their value. If not, we can try to use [handlePrevDFG].
      */
     protected fun handleHasInitializer(node: HasInitializer, depth: Int): Any? {
+        // A global or static variable can have writes somewhere else too, not just its
+        // initializer. If the DFG shows more than one, use handlePrevDFG to decide instead.
+        if (
+            node is Variable &&
+                (node.isGlobal || node.modifiers.contains("static")) &&
+                node.prevFullDFG.size > 1
+        ) {
+            return handlePrevDFG(node as Node, depth)
+        }
+
         // If we have an initializer, we can use it. Otherwise, we can fall back to the prevDFG
         return if (node.initializer != null) {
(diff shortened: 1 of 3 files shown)


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluator.kt
```diff
@@ -179,6 +179,16 @@ open class ValueEvaluator(
...
      */
     protected fun handleHasInitializer(node: HasInitializer, depth: Int): Any? {
+        // A global or static variable can have writes somewhere else too, not just its
+        // initializer. If the DFG shows more than one, use handlePrevDFG to decide instead.
+        if (
+            node is Variable &&
+                (node.isGlobal || node.modifiers.contains("static")) &&
+                node.prevFullDFG.size > 1
+        ) {
+            return handlePrevDFG(node as Node, depth)
```
KuechA: I'm not quite sure about this code. Why should we do this only for globals or static variables? And why only for `prevFullDFG.size > 1`? And: Do we need this dedicated handling for `node.initializer != null`? Usually, the prevDFG should end up there anyways.

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluatorTest.kt
```diff
@@ -796,4 +796,60 @@ class ValueEvaluatorTest {
...
+    fun testHandleHasInitializerWithWrite() {
+        with(TestLanguageFrontend()) {
+            // int counter = 0;
+            val counter = newVariable("counter")
+            counter.initializer = newLiteral(0)
+
+            // counter = 5; somewhere else, i.e. second prevDFG edge to counter.
+            val write = newReference("counter")
+            write.refersTo = counter
+            write.prevDFG = mutableSetOf(newLiteral(5))
```
KuechA: I do not really like the idea of re-implementing the DFG Passes in this test file but I'm also wondering why the tests above don't use fluent DSL

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluatorTest.kt
```diff
@@ -796,4 +796,60 @@ class ValueEvaluatorTest {
...
+            counter.prevDFG = mutableSetOf(counter.initializer as Node, write)
+
+            // A read of counter, e.g. the condition of `if(counter)`.
+            val read = newReference("counter")
+            read.refersTo = counter
+            read.prevDFG = mutableSetOf(counter)
+
+            // There is a write to counter, so we cannot evaluate the value of counter.
+            assertEquals("{counter}", ValueEvaluator().evaluate(read))
```
KuechA: Can we check if this works for the `MultiValueEvaluator` in addition to the failing evaluation of the `ValueEvaluator`? that one should usually return a result like `setOf(0, 1)`.

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluatorTest.kt
```diff
@@ -796,4 +796,60 @@ class ValueEvaluatorTest {
...
+
+            val read = newReference("counter")
+            read.refersTo = counter
+            read.prevDFG = mutableSetOf(counter)
+
+            assertEquals(0, ValueEvaluator().evaluate(read))
+        }
+
+        with(TestLanguageFrontend()) {
+            // A non-Variable HasInitializer, e.g. `int a[] = {1, 2, 3};`
```
KuechA: The comment doesn't match the code. The initializer is wrong

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluator.kt
lshala: Thx. Changed the condition

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluatorTest.kt
lshala: Adapted the tests as `ValueEvaluationTests` does by using fluent DSL

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluatorTest.kt
lshala: Done

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/evaluation/ValueEvaluatorTest.kt
lshala: Done

KuechA (approved):

PR #2933, opened 2026-09-03
Go EOG pass scope resolution replaced by ast lookup

konradweiss:
In the GoEvaluationOrderGraphPass we perform a `scopeManager.currentFunction` lookup but the EOGPass does not perform an enter scope. I changed it to a lookup in the parent Chain such that the scope is not needed.


cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
```diff
@@ -1322,12 +1325,18 @@ open class EvaluationOrderGraphPass(ctx: TranslationContext) : TranslationUnitPa
 
     /** We use the scope where the current [node] is in, to find a statement labeled with [label] */
     fun getLabeledASTNode(node: Node, label: String): Node? {
-        scopeManager.jumpTo(node.scope)
-        val labelStatement = scopeManager.getLabel(label)
-        labelStatement?.subStatement?.let {
-            return it
+        // Temporarily jump to the node's scope to perform label lookup and restore the old scope
+        val oldScope = scopeManager.jumpTo(node.scope)
```
KuechA, line 1326: Why don't you use `withScope` instead?

KuechA (approved):

cpg-core/src/main/kotlin/de/fraunhofer/aisec/cpg/passes/EvaluationOrderGraphPass.kt
konradweiss: Done that

PR #2938, opened 2026-09-04
Iterative SCC Pass

maximiliankaul:
The current implementation of Tarjan (used in the SccPass) is recursive. This doesn't work for real-world code as the algorithm crashes due to reaching the stack limit. 
This PR transfers the algorithm to an iterative implementation, which fixes this issue.

### DIFF
--- a/cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/SccPassTest.kt
+++ b/cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/SccPassTest.kt
@@ -32,6 +32,52 @@ import kotlin.test.assertTrue
 
 class SccPassTest {
     /**
+     * Regression test for the `StackOverflowError` that the original recursive [SccPass.tarjan] hit
+     * on deep EOGs (recursion depth = longest simple path in the EOG). Hand-builds a chain of [n]
+     * plain nodes linked via `nextEOG` (same approach as
+     * [testBlacklistedNodeDoesNotAbortSuccessorScan] below, just chained deep) and runs `tarjan` on
+     * a separate [Thread] with a deliberately small stack, so the test is deterministic regardless
+     * of how many bytes the JVM uses per recursive frame on a given platform.
+     *
+     * An earlier version of this test built the chain through the real frontend/DSL (many
+     * sequential `if`s, to force basic-block boundaries via `BasicBlockCollectorPass`). That turned
+     * out unreliable in practice: `defaultPasses()` pulls in unrelated passes
+     * (`ControlFlowSensitiveDFGPass`/`SymbolResolver`) with their own complexity blowups on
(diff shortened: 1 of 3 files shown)


cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/SccPassTest.kt
```diff
@@ -0,0 +1,123 @@
...
+     * out unreliable in practice: `defaultPasses()` pulls in unrelated passes
+     * (`ControlFlowSensitiveDFGPass`/`SymbolResolver`) with their own complexity blowups on
+     * functions with many branches (`OutOfMemoryError` at just 3,000 `if`s), and even after
+     * trimming to the minimal passes `SccPass` needs, the resulting EOG chain was consistently far
+     * shallower than the `if` count, and default-JVM-stack-sized runs never overflowed even at very
+     * high `n`. Hand-building the chain directly avoids both problems and gives an exact, known
+     * depth.
+     */
+    @Test
+    fun testDeepEogDoesNotStackOverflow() {
```
maximiliankaul: Do we want a test like this? I'm thinking about removing it.

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/SccPassTest.kt
```diff
@@ -0,0 +1,123 @@
...
+ *
+ */
+package de.fraunhofer.aisec.cpg.passes
+
+import de.fraunhofer.aisec.cpg.TranslationContext
+import de.fraunhofer.aisec.cpg.graph.AnnotationMember
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class SccPassTest {
```
maximiliankaul, line 51: do we want tests that the algorithm works? so some graphs with loops and nested loops to check the algorithm here?

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/SccPassTest.kt
konradweiss: I would remove this

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/SccPassTest.kt
konradweiss: I am thinking about whether we want to have few tests, just to see if the algorithm works on a base level and to have some guardrails against future changes breaking this feature.

konradweiss (approved):

cpg-core/src/test/kotlin/de/fraunhofer/aisec/cpg/passes/SccPassTest.kt
maximiliankaul: done

PR #2939, opened 2026-09-04
cpg-ai: Add `CpgSession` to support multiple projects and reusable MCP tools

lshala:
This PR aims to make the MCP server more flexible, so that consumers from other projects can use our existing MCP tools without needing to implement custom ones. For this, the previous `runOnCpg` function is extended and renamed to `runOnSession`, which now accepts a `CpgSession` object that holds, besides the `TranslationResult` also the `TranslationContext`, `nodeToPass`, and `CpgSessionStatus`.


cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/mcpserver/tools/utils/Util.kt
```diff
@@ -251,21 +257,92 @@ fun getAvailableOperations(): List<Class<out Operation>> {
...
+ */
+interface HasProjectNamePayload {
+    val projectName: String?
+}
+
+/**
+ * Adapts a payload of arbitrary type [T], one that does not implement [HasProjectNamePayload]
+ * itself
+ */
+class ToolCall<T>(override val projectName: String?, val payload: T) : HasProjectNamePayload
```
KuechA, line 310: The wrapping makes manual testing (e.g. via the mcp inspector tool) more difficult. The other option would be to extend the Payload classes. It also has the problem that not all tools really need a session/project, so for these, it adds overhead which is unnecessary (e.g. tools simply listing concept and operation classes/yaml).

cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/mcpserver/tools/utils/Serializable.kt
```diff
@@ -222,6 +222,8 @@ data class CpgAnalysisResult(
     val components: List<String> = listOf(),
     /** Notes about what the project auto-detection recognized, e.g., a compilation database. */
     val detectionNotes: List<String> = listOf(),
+    /** The names of all currently analyzed projects that can be targeted via `projectName`. */
+    val projectNames: List<String> = listOf(),
```
KuechA, line 226: why is this a list?

KuechA (dismissed):
Still have to test it, first review

cpg-ai/src/main/kotlin/de/fraunhofer/aisec/cpg/ai/mcp/mcpserver/tools/utils/Util.kt
lshala: Ok, I hadn't thought about the mcp inspector, fair point. I intentionally wanted a solution where a user doesn't need to think about whether a payload class needs to implement that or not, it's handled automatically instead.

overhead point: I'm not sure if this applies to these tools, as they actually don't go through our `addTool` wrapper, but their own mcp sdk function instead. So no projectName gets added and no session lookup happens.

KuechA: ok

